PR #2, opened 2021-07-09
feat: add Slider component

NadiaPosch:


src/components/Slider/Slider.tsx
```diff
@@ -0,0 +1,44 @@
...
+import { Dispatch, ReactElement, SetStateAction } from "react";
+import css from "./Slider.module.css";
+
+export interface IconItem {
+    icon: ReactElement<IconProps>;
+    name: string;
+}
+
+export interface SliderProps {
+    items: string[] | IconItem[];
```
dwirz, line 13: IMO it would be better to use an object with an `id` prop instead of plain string. atm if the string is translated the dev who uses the `Slider` somehow needs to map the selected string to a unique item outside of the `Slider`. do you know what i mean? 🙂

src/components/Slider/Slider.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+const isIconItem = (item: string | IconItem): item is IconItem => (item as IconItem).icon !== undefined;
+
+export default function Slider({ items, activeItem: active, onChange }: SliderProps): ReactElement<SliderProps> {
+    return (
+        <div data-test-id="Slider" className={css.wrapper}>
+            {items.map((item, i) =>
+                isIconItem(item) ? (
+                    <div
+                        key={i}
```
dwirz, line 26: if the item has an `id`-prop use this instead of `i`. also using the array index is discouraged to use if there is another unique value you can use. (also see comment on SO https://stackoverflow.com/a/46735689 why using index should be avoided)

src/components/Slider/Slider.spec.tsx
```diff
@@ -0,0 +1,32 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+import { mount } from "@cypress/react";
+import { useState } from "react";
+import Slider from "./Slider";
+import css from "./Slider.module.css";
+
+const Slider_Items = ["abc", "def", "ghi"];
```
dwirz, line 8: not sure if this is a convention of frontify (@SamuelAlev) but @NadiaPosch IMO either write everything in capital letters eg `SLIDER_ITEMS` or classical snake_case. this is more of a gusto thingy but personally i never saw this kind of variable "convention" 😉

src/components/Slider/Slider.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+export interface SliderProps {
+    items: string[] | IconItem[];
+    activeItem: string;
+    onChange: Dispatch<SetStateAction<string>>;
+}
+
+const isIconItem = (item: string | IconItem): item is IconItem => (item as IconItem).icon !== undefined;
+
+export default function Slider({ items, activeItem: active, onChange }: SliderProps): ReactElement<SliderProps> {
```
dwirz, line 20: @SamuelAlev not sure if naming this component `Slider` is somehow misleading, wdyt? (if frontify always names these kind of component slider i'm fine with it 🙂 )

src/components/Slider/Slider.tsx
NadiaPosch: yes, good idea!

.husky/pre-push
```diff
@@ -1,4 +0,0 @@
-#!/bin/sh
-. "$(dirname "$0")/_/husky.sh"
-
-npm test
```
dwirz, line 4: why did you remove the pre-push hook?

src/components/Slider/Slider.tsx
```diff
@@ -5,37 +5,36 @@ import { Dispatch, ReactElement, SetStateAction } from "react";
...
 
-const isIconItem = (item: string | IconItem): item is IconItem => (item as IconItem).icon !== undefined;
+const isIconItem = (item: TextItem | IconItem): item is IconItem => (item as IconItem).icon !== undefined;
+
+export default function Slider({ items, activeItemId, onChange }: SliderProps): ReactElement<SliderProps> {
+    const getItemClasses = (id: string) => [css.item, id === activeItemId ? css.active : ""].join(" ");
 
-export default function Slider({ items, activeItem: active, onChange }: SliderProps): ReactElement<SliderProps> {
     return (
         <div data-test-id="Slider" className={css.wrapper}>
```
dwirz, line 29: actually since it is a "list" of items you could use the "semantically correct"  html element `<ul>` and items as `<li>` wdyt?

dwirz (changes requested):

src/components/Slider/Slider.tsx
```diff
@@ -0,0 +1,43 @@
...
+export interface SliderProps {
+    items: TextItem[] | IconItem[];
+    activeItemId: string;
+    onChange: Dispatch<SetStateAction<string>>;
+}
+
+const isIconItem = (item: TextItem | IconItem): item is IconItem => (item as IconItem).icon !== undefined;
+
+export default function Slider({ items, activeItemId, onChange }: SliderProps): ReactElement<SliderProps> {
+    const getItemClasses = (id: string) => [css.item, id === activeItemId ? css.active : ""].join(" ");
```
SamuelAlev, line 26: It can be moved out of the function

src/components/Slider/Slider.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+    return (
+        <ul data-test-id="Slider" className={css.wrapper}>
```
SamuelAlev, line 29:
```suggestion
        <ul data-test-id="slider" className={css.wrapper}>
```

We use kebab-case for the id

src/components/Slider/Slider.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+export interface TextItem {
+    id: string;
+    name: string;
+}
+
+export interface SliderProps {
+    items: TextItem[] | IconItem[];
+    activeItemId: string;
+    onChange: Dispatch<SetStateAction<string>>;
```
SamuelAlev, line 20:
```suggestion
    onChange: (id: string) => void;
```

We don't use Redux within the product

src/components/Slider/Slider.tsx
```diff
@@ -0,0 +1,43 @@
...
+const isIconItem = (item: TextItem | IconItem): item is IconItem => (item as IconItem).icon !== undefined;
+
+export default function Slider({ items, activeItemId, onChange }: SliderProps): ReactElement<SliderProps> {
+    const getItemClasses = (id: string) => [css.item, id === activeItemId ? css.active : ""].join(" ");
+
+    return (
+        <ul data-test-id="Slider" className={css.wrapper}>
+            {items.map((item) =>
+                isIconItem(item) ? (
+                    <li key={item.id} onClick={() => onChange(item.id)} className={getItemClasses(item.id)}>
```
SamuelAlev, line 32: It is missing the actual selection change (clicking the second item doesn't put the active state on it)

src/components/Slider/Slider.stories.tsx
```diff
@@ -0,0 +1,22 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
```
SamuelAlev, line 1: It misses

```ts
Default.argTypes = {
    onChange: { action: "onClear" },
};
```

src/components/Slider/Slider.stories.tsx
```diff
@@ -0,0 +1,22 @@
...
+export default {
+    title: "Slider",
+    component: Slider,
+} as Meta;
+
+const SliderTemplate: Story<SliderProps> = (args) => <Slider {...args} />;
+
+export const Default = SliderTemplate.bind({});
+
+Default.args = {
```
SamuelAlev: We could probably add another story with icons only

src/components/Slider/Slider.spec.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+describe("Slider Component", () => {
+    it("should render foo text correctly", () => {
+        const Component = () => {
+            const [active, setActive] = useState(SLIDER_ITEMS[0].id);
+
+            return <Slider items={SLIDER_ITEMS} activeItemId={active} onChange={setActive} />;
+        };
+
+        mount(<Component />);
```
SamuelAlev, line 22:
```suggestion
        const onChangeStub = cy.stub().as("onChangeStub");

        mount(<Slider items={SLIDER_ITEMS} activeItemId={SLIDER_ITEMS[0].id} onChange={onChangeStub} />);
```

We can use cypress stub here

src/components/Slider/Slider.spec.tsx
```diff
@@ -0,0 +1,36 @@
...
+import css from "./Slider.module.css";
+
+const SLIDER_ITEMS = [
+    { id: "a", name: "abc" },
+    { id: "b", name: "def" },
+    { id: "c", name: "ghi" },
+];
+
+describe("Slider Component", () => {
+    it("should render foo text correctly", () => {
```
SamuelAlev, line 15:
```suggestion
    it("renders", () => {
```
I would move out the click to its own test `it('handles click and change the active item')`

SamuelAlev (changes requested):
Nice one! Just got a few points to fix otherwise looks great

src/components/Slider/Slider.tsx
NadiaPosch: The state change will be handled by the parent component so that the active item is also accessible from outside the Slider. I have updated the story to reflect a working example.

src/components/Slider/Slider.spec.tsx
NadiaPosch: I'm not sure if a stub would work for the state change. I have updated the tests, would be great if you could have another look at it :)

dwirz (approved):

src/components/Slider/Slider.tsx
```diff
@@ -0,0 +1,51 @@
...
+    [css.item, id === activeItemId ? css.active : ""].join(" ");
+
+export default function Slider({ items, activeItemId, onChange }: SliderProps): ReactElement<SliderProps> {
+    return (
+        <ul data-test-id="slider" className={css.wrapper}>
+            {items.map((item) =>
+                isIconItem(item) ? (
+                    <li
+                        key={item.id}
+                        onClick={() => onChange(item.id)}
```
SamuelAlev, line 34: > The state change will be handled by the parent component so that the active item is also accessible from outside the Slider. I have updated the story to reflect a working example.

What are the pros of doing of moving out the active item id?
My way of doing it would be like:
```ts
const onItemClicked = (id: string) => {
	setActiveItemId(id) // Local useState
	onChange && onChange(id)
}

//[...]
<li
    key={item.id}
    onClick={() => onItemClicked(item.id)}
    className={getItemClasses(item.id, activeItemId)}
>
```

src/components/Slider/Slider.spec.tsx
```diff
@@ -0,0 +1,38 @@
...
+};
+
+beforeEach(() => {
+    mount(<Component />);
+    cy.get("[data-test-id=slider]").as("slider");
+    cy.get("@slider").children().first().as("firstItem");
+    cy.get("@slider").children().last().as("lastItem");
+});
+
+describe("Slider Component", () => {
```
SamuelAlev, line 26: We could also add an icons slider in the tests as the result may change depending on texts vs icons.

SamuelAlev (commented):
Let's talk about my comments this Monday :)

.husky/pre-push
NadiaPosch, line 4: I'm not sure how to undo the deletion of the pre-push file. Could you help me with that @SamuelAlev ?

SamuelAlev: `git checkout main .husky/pre-push` and then commit it back

PR #3, opened 2021-07-13
feat: add dropdown component

NadiaPosch:


src/components/Dropdown/DropdownMenuItem/DropdownMenuItem.tsx
```diff
@@ -0,0 +1,76 @@
...
+import { ReactComponent as Check } from "@components/Icon/Svg/Check.svg";
+import { ReactElement } from "react";
+import css from "./DropdownMenuItem.module.css";
+
+export enum MenuItemVariant {
+    "Small",
+    "Large",
+}
+
+export type MenuItem = {
```
dwirz, line 13: Try to add/use a union of two types for small and large

dwirz:
align styling of trigger for hover and disable

dwirz (changes requested):

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,68 @@
...
+    activeItemId = "",
+    placeholder = "",
+    size = Size.Small,
+}: DropdownProps): ReactElement<DropdownProps> {
+    const [open, setOpen] = useState(false);
+    const activeItem = getActiveItem(menuItems, activeItemId);
+
+    return (
+        <div className={css.wrapper}>
+            <div
```
dwirz, line 35: from a semantic perspective IMO this should be a `<button>` (you then potentially need to wrap your content within a `<span>` because grid/flex does not work as expected for a button 🙈 )

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,67 @@
...
+}: DropdownProps): ReactElement<DropdownProps> {
+    const [open, setOpen] = useState(false);
+    const activeItem = getActiveItem(menuItems, activeItemId);
+
+    return (
+        <div className={css.wrapper}>
+            <button
+                data-test-id="dropdown-trigger"
+                onClick={() => setOpen(!open)}
+                className={`${css.trigger} ${!activeItem ? css.inactive : ""} ${size === Size.Large ? css.large : ""}`}
```
dwirz: There is also another way you could do that but not sure it is more readable
`[css.trigger, !activeItem && css.inactive, size === Size.Large && css.large].filter(Boolean).join(" " )`

src/components/Dropdown/DropdownMenu/DropdownMenu.tsx
```diff
@@ -0,0 +1,44 @@
...
+}
+
+export default function DropdownMenu({
+    items,
+    onChange,
+    size = Size.Small,
+    activeItemId = "",
+}: DropdownMenuProps): ReactElement<DropdownMenuProps> {
+    return (
+        <ul data-test-id="dropdown-menu" className={`${css.menu} ${size === Size.Large ? css.large : ""}`}>
```
dwirz, line 22: same as above

src/components/Dropdown/Dropdown.spec.tsx
```diff
@@ -0,0 +1,124 @@
...
+        cy.get(MENU_ITEM_TEXT_ID).eq(2).click();
+        cy.get(MENU_ITEM_TEXT_ID).first().contains("Large warning");
+    });
+    it("renders divider", () => {
+        mount(<Component items={LARGE_ITEMS} size={Size.Large} initialActiveId={LARGE_ITEMS[0][0].id} />);
+        cy.get(DROPDOWN_TRIGGER_ID).click();
+        cy.get(DROPDOWN_MENU_ID).children().as("itemBlocks");
+        cy.get("@itemBlocks").should("have.length", 2);
+        cy.get("@itemBlocks").first().should("have.css", "border-bottom", BORDER_STYLE);
+        cy.get("@itemBlocks").last().should("not.have.css", "border-bottom", BORDER_STYLE);
```
dwirz, line 122: can't we just for the prop and ignore the style itself?

src/components/Dropdown/DropdownMenuItem/DropdownMenuItem.stories.tsx
```diff
@@ -0,0 +1,44 @@
...
+    component: DropdownMenuItem,
+    argTypes: {
+        warning: {
+            options: [true, false],
+            control: { type: "radio" },
+        },
+        disabled: {
+            options: [true, false],
+            control: { type: "radio" },
+        },
```
dwirz, line 20: for boolean cases you could add
```
args: {
  warning: false,
  disabled: false,
}
```
this way storybook creates a bool toggle

src/components/Dropdown/DropdownMenuItem/DropdownMenuItem.stories.tsx
```diff
@@ -0,0 +1,44 @@
...
+const DropdownMenuItemTemplate: Story<DropdownMenuItemProps> = (args) => <DropdownMenuItem {...args} />;
+
+export const Large = DropdownMenuItemTemplate.bind({});
+Large.args = {
+    title: "Large Menu Item",
+    subtitle: "Subtitle",
+    icon: <Audio size={IconSize.Size16} />,
+    size: Size.Large,
+    warning: false,
+    disabled: false,
```
dwirz, line 33: if you add `warning` and `disabled` to `default.args` above you can omit it here

src/components/Dropdown/DropdownMenuItem/DropdownMenuItem.stories.tsx
```diff
@@ -0,0 +1,44 @@
...
+    onClick: { action: "onClear" },
+};
+
+export const Small = DropdownMenuItemTemplate.bind({});
+Small.args = {
+    title: "Small Menu Item",
+    icon: <Audio size={IconSize.Size16} />,
+    warning: false,
+    disabled: false,
```
dwirz, line 42: same as above

src/components/Dropdown/DropdownMenuItem/DropdownMenuItem.tsx
```diff
@@ -0,0 +1,84 @@
...
+    disabled = false,
+    active = false,
+}: DropdownMenuItemProps): ReactElement<DropdownMenuItemProps> {
+    return (
+        <li
+            data-test-id={icon !== undefined ? "dropdown-menu-item-icon" : "dropdown-menu-item-text"}
+            className={getClassNames(size, warning, disabled, subtitle)}
+            tabIndex={1}
+            onClick={() => {
+                if (onClick) {
```
dwirz, line 67: for completeness i would add `&& !disabled`

src/components/Dropdown/Dropdown.tsx
NadiaPosch: I have extracted it into a function as in the `MenuItem` component. It's more readable that way in my opinion.

src/components/Dropdown/DropdownMenu/DropdownMenu.tsx
NadiaPosch: Since there are only two different classes here I don't think the readablility would be improved by joining it together in an array. Therefore I'd leave it as it is.

src/components/Dropdown/Dropdown.spec.tsx
NadiaPosch: I tried but somehow it only works when specifying a style 🤷🏽‍♀️

src/components/Dropdown/DropdownMenuItem/DropdownMenuItem.stories.tsx
```diff
@@ -0,0 +1,33 @@
...
+import DropdownMenuItem, { DropdownMenuItemProps } from "./DropdownMenuItem";
+
+export default {
+    title: "DropdownMenuItem",
+    component: DropdownMenuItem,
+    args: {
+        warning: false,
+        disabled: false,
+        onClick: { action: "onClear" },
+    },
```
dwirz, line 16: The `onClick`
```
    args: {
        //...
        onClick: { action: "onClear" },
    },
```
should be an `argTypes`
```
    argTypes: { 
      onClick: { action: "Item Clicked" },
    },
```
otherwise you get the following error `Uncaught TypeError: onClick is not a function`

src/components/Dropdown/DropdownMenuItem/DropdownMenuItem.module.css
```diff
@@ -0,0 +1,101 @@
...
+}
+
+/**
+ * Disabled
+ */
+.disabled {
+    pointer-events: none;
+    color: var(--f-black-40);
+    background-color: var(--f-black-00);
+    pointer-events: none;
```
dwirz, line 76: sonar told me that this prop is duplicated 😏

src/components/Dropdown/DropdownMenu/DropdownMenu.tsx
```diff
@@ -0,0 +1,44 @@
...
+                                onClick={() => onChange(id)}
+                                title={title}
+                                icon={icon}
+                                subtitle={subtitle}
+                                size={size}
+                                style={style}
+                                disabled={disabled}
+                                active={id === activeItemId}
+                            />
+                        ))}
```
oliverschwendener, line 38: Is there a reason behind using object destructuring here for the `menuItem`? I think if we would just use 

```tsx
{menuItems.map((menuItem) => (
    <DropdownMenuItem
        key={menuItem.id}
        onClick={() => onChange(menuItem.id)}
        title={menuItem.title}
        icon={menuItem.icon}
        subtitle={menuItem.subtitle}
        size={menuItem.size}
        style={menuItem.style}
        disabled={menuItem.disabled}
        active={menuItem.id === activeItemId}
    />
))}
```

it would be easier to read and understand where each variable is coming from. Additionally it would help avoiding naming collisions with other variables that are defined in an upper scope. For example when using object destructuring the variable `size` on line 26 is colliding with the variable `size` from line 18.

@NadiaPosch What do you think? :)

dwirz (approved):

src/components/Dropdown/DropdownMenu/DropdownMenu.tsx
NadiaPosch: I try to destructure whenever it's possible to reduce redundancy but if you prefer the version above that's fine for me :)

oliverschwendener: In this case I'd prefer not to destructure the object because we are otherwise overwriting variables from the upper scope.

oliverschwendener:
@NadiaPosch Something that I just noticed: should the dropdown menu hide when the dropdown loses focus, for example when the user clicks outside of the dropdown?

src/components/Dropdown/DropdownMenu/DropdownMenu.tsx
```diff
@@ -0,0 +1,44 @@
...
+export default function DropdownMenu({
+    items,
+    onChange,
+    size = Size.Small,
+    activeItemId = "",
+}: DropdownMenuProps): ReactElement<DropdownMenuProps> {
+    return (
+        <ul className={`${css.menu} ${size === Size.Large ? css.large : ""}`}>
+            {items.map((menuItems, i) => (
+                <li key={i} className={css.divider} data-test-id="dropdown-divider">
```
SamuelAlev, line 24: We may want to have an `id` instead of using the index to cache the rendered elements better.

```
Array of MenuItem:
0. [...]
1. [...] -> deleted
2. [...] -> get rerendered even without changes
```

oliverschwendener (changes requested):
See my comment above.

src/components/Dropdown/DropdownMenuItem/DropdownMenuItem.tsx
```diff
@@ -0,0 +1,80 @@
...
+        >
+            {icon}
+            {subtitle ? (
+                <div className={contentClassNames}>
+                    <div data-test-id="dropdown-menu-item-title" title={title}>
+                        {title}
+                    </div>
+                    <div data-test-id="dropdown-menu-item-subtitle" className={css.subtitle} title={subtitle}>
+                        {subtitle}
+                    </div>
```
SamuelAlev, line 70: I would add a `user-select: none;` on both element

src/components/Dropdown/DropdownMenuItem/DropdownMenuItem.module.css
```diff
@@ -0,0 +1,92 @@
...
+    box-shadow: var(--f-shadow-element-focused-background-white);
+}
+
+.item svg {
+    flex-shrink: 0;
+}
+
+.content {
+    flex: 1;
+    padding-right: var(--f-spacing-1x);
```
SamuelAlev, line 51:
```suggestion
```

I don't think it is needed as we have the `grid-gap` set and I don't think it affects anything

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,94 @@
...
+                        onClick={() => onChange("")}
+                        disabled={disabled}
+                    >
+                        <IconReject size={IconSize.Size13} />
+                    </button>
+                )}
+                <button className={disabled ? css.disabled : ""} onClick={() => setOpen(!open)} disabled={disabled}>
+                    {open ? <IconCaretUp size={IconSize.Size16} /> : <IconCaretDown size={IconSize.Size16} />}
+                </button>
+            </div>
```
SamuelAlev, line 80: We miss the contextual options part

Capture d’écran 2021-07-15 à 16 45 37

SamuelAlev (changes requested):

src/components/Dropdown/DropdownMenu/DropdownMenu.tsx
dwirz: i told @NadiaPosch to do it like this for the moment, but you're right and we're with you on this, it's better to change the items from `MenuItem[][]` to something like `{id: string; items: MenutItem[] }[]` this it would also allow to potentially add a title per group or other props

src/components/Dropdown/Dropdown.tsx
dwirz: since atm we do not actually know what should happen if someone click on the dots i would add them later in a seperate PR wdyt?

src/components/Dropdown/DropdownMenu/DropdownMenu.tsx
NadiaPosch: Done 👌🏼

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,113 @@
...
+import { ReactComponent as IconCaretUp } from "@components/Icon/Svg/CaretUp.svg";
+import { ReactComponent as IconReject } from "@components/Icon/Svg/Reject.svg";
+import useClickOutside from "@hooks/useClickOutside";
+import { Size } from "@utilities/enum";
+import { ReactElement, useRef, useState } from "react";
+import css from "./Dropdown.module.css";
+import DropdownMenu from "./DropdownMenu/DropdownMenu";
+import DropdownMenuItem, { MenuItem } from "./DropdownMenuItem/DropdownMenuItem";
+
+export interface ItemBlock {
```
dwirz, line 14: Move this to `DropdownMenu` otherwise you have a circular dependency between `Dropdown` and `DropdownMenu`. Also I would name it `MenuBlock`.

NadiaPosch: Done 👌🏼

src/components/Dropdown/Dropdown.tsx
NadiaPosch: I agree with Dominique and have also talked to Simone about it. Maybe we can make a note somewhere to make sure it won't be overlooked when the design is ready.

oliverschwendener (approved):
I'll be off for a week starting today. I'll approve this PR so you don't have to wait for my approval. 😉

src/components/Dropdown/Dropdown.stories.tsx
```diff
@@ -0,0 +1,170 @@
...
+    const [activeItemId, setActiveItemId] = useState("");
+    return <Dropdown activeItemId={activeItemId} onChange={setActiveItemId} {...args} />;
+};
+
+export const SmallSelect = DropdownTemplate.bind({});
+
+SmallSelect.args = {
+    placeholder: "select item",
+    size: Size.Small,
+    itemBlocks: [
```
oliverschwendener, line 35: Shouldn't this be `menuBlocks`?

src/components/Dropdown/Dropdown.stories.tsx
```diff
@@ -0,0 +1,170 @@
...
+        },
+    ],
+};
+
+export const LargeSelect = DropdownTemplate.bind({});
+
+LargeSelect.args = {
+    placeholder: "select item",
+    size: Size.Large,
+    itemBlocks: [
```
oliverschwendener, line 98: And also this?

src/components/Dropdown/Dropdown.stories.tsx
NadiaPosch: Yes, I just fixed it 🙈

src/components/Dropdown/Dropdown.spec.tsx
```diff
@@ -0,0 +1,162 @@
...
+    placeholder?: string;
+    initialActiveId?: string;
+    clearable?: boolean;
+};
+
+const Component: FC<Props> = ({
+    menuBlocks,
+    size = Size.Small,
+    placeholder = "",
+    initialActiveId = "",
```
SamuelAlev, line 89:
```suggestion
    placeholder,
    initialActiveId,
```
It can be `undefined`

src/components/Dropdown/Dropdown.module.css
```diff
@@ -0,0 +1,73 @@
...
+
+.dropdown {
+    position: relative;
+    width: 100%;
+    font-family: var(--f-font-standard);
+    font-size: var(--f-font-14-size);
+    line-height: var(--f-font-14-line-height);
+}
+
+.trigger-wrapper {
```
SamuelAlev, line 11:
```suggestion
.triggerWrapper {
```

SamuelAlev (changes requested):
We should have `12px` instead of `20px` for the padding of the text (large and small) and right caret-down have `12px` instead of `8px` (small version).
Capture d’écran 2021-07-16 à 14 39 09

The `grid-gap`/`gap` should be `12px`
Capture d’écran 2021-07-16 à 14 51 10


[I also asked to Simone about MenuItem disabled vs hovering colors](https://www.figma.com/file/v00soRe37Tdb14Nno9VFEl?node-id=6086:190565#94717216) as they are the same and makes it weird for the eyes.

NadiaPosch:
@SamuelAlev Where did you find the spacing examples exactly? In the Figma file that is linked in the ticket, I measure `20px` for small and large items 🤔
https://www.figma.com/file/v00soRe37Tdb14Nno9VFEl/Arcade?node-id=6086%3A190565

NadiaPosch:
I just saw that there are two types of selection indicator icons, so I'll quickly add it as an option
image

SamuelAlev:
@NadiaPosch https://www.figma.com/file/v00soRe37Tdb14Nno9VFEl/Arcade?node-id=4588%3A211854

Seems the Dropdown has 12px but the MenuItem has 16px, problably a quick sync with Simone is worth.

PR #4, opened 2021-07-15
feat: Add form control

dwirz:


src/components/FormControl/FormControl.tsx
```diff
@@ -0,0 +1,72 @@
...
+        data-test-id="form-control-helper-text"
+        className={`${css.helper} ${disabled ? css.disabled : ""} ${
+            variant === HelperVariant.Error ? css.error : variant === HelperVariant.Success ? css.success : ""
+        }`}
+    >
+        {text}
+    </span>
+);
+
+export interface FormControlProps {
```
deniaz, line 29: The `HelperText` uses `type` for the prop-typing, whereas here it's an `interface`. I think we should avoid having a mixture of both. Personally I favour `type` over `interface` since we're only going for type-safety. Interfaces come from an OOP background, where classes implement `interfaces` so you can achieve loose coupling (is it Liskov substitution principle? I'm not sure after all these years without OOP :joy:). Interfaces can be extended and implemented, where as types are just that―types.

What do the two of you think? @dwirz @SamuelAlev

src/components/InputLabel/InputLabel.tsx
```diff
@@ -0,0 +1,73 @@
...
+                data-test-id="input-label"
+            >
+                {children}
+            </label>
+            {required && (
+                <span data-test-id="input-label-required" className={css.asterisk}>
+                    *
+                </span>
+            )}
+            {tooltip && (
```
deniaz, line 55: What about extracting the Tooltip to its own component?

src/components/InputLabel/InputLabel.tsx
```diff
@@ -0,0 +1,73 @@
...
+}: InputLabelProps): ReactElement<InputLabelProps> {
+    const tooltipIconRef = useRef(null);
+    const tooltipRef = useRef(null);
+    const { styles, attributes } = usePopper(tooltipIconRef.current, tooltipRef.current, {
+        placement: "auto-end",
+        modifiers: [
+            {
+                name: "offset",
+                options: {
+                    offset: [0, 9],
```
deniaz, line 35: use `const` over magic numbers maybe? :stuck_out_tongue_winking_eye:

src/components/InputLabel/InputLabel.tsx
```diff
@@ -0,0 +1,73 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
```
deniaz, line 1: @SamuelAlev wouldn't it make more sense to have a `LICENSE` file or something like that? I'm not sure what benefit this line actually has and it's way to easy to forget :see_no_evil:

deniaz (approved):

src/components/FormControl/FormControl.tsx
```diff
@@ -0,0 +1,72 @@
...
+    children: ReactNode;
+    direction?: "horizontal" | "vertical";
```
SamuelAlev, line 31: We usually use Enums for that

src/components/FormControl/FormControl.tsx
SamuelAlev: I could agree with you as type composition looks nicer with `type` than `interface`.
And yes `interface` comes from an OOP background so it makes more sense.

src/components/InputLabel/InputLabel.tsx
dwirz: dfee1e6

src/components/InputLabel/InputLabel.tsx
dwirz: dfee1e6

src/components/FormControl/FormControl.tsx
dwirz: I suggest that after this one and https://github.com/Frontify/arcade/pull/7 is merged I open a new PR and align all the `Props` definitions

src/components/FormControl/FormControl.tsx
dwirz: e962d0a

PR #6, opened 2021-07-16
feat: add tree component

triggertoo:


src/components/Tree/Tree.spec.tsx
```diff
@@ -0,0 +1,97 @@
...
+import { IconSize } from "@components/Icon/Icon";
+import { ReactComponent as IconDocument } from "@components/Icon/Svg/Document.svg";
+import { ReactComponent as IconGuidelines } from "@components/Icon/Svg/Guidelines.svg";
+import { mount } from "@cypress/react";
+import { FC, useState } from "react";
+import { TreeNode } from "./Node";
+import Tree from "./Tree";
+import css from "./Tree.module.css";
+
+const Tree_FOO = "bar";
```
dwirz, line 12: unused?

src/components/Tree/Tree.module.css
```diff
@@ -0,0 +1,62 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+.tree {
+    max-width: 800px;
```
dwirz, line 4: isn't there a var for that? or shouldn't the one who uses this component decide what the `max-width` is?

src/components/Tree/Tree.module.css
```diff
@@ -0,0 +1,62 @@
...
+    text-align: left;
+    list-style: none;
+    margin: 0;
+    padding: 0;
+    font-family: var(--f-font-standard);
+    font-weight: 400;
```
dwirz, line 10: isn't there a var for `font-weight`s?

src/components/Tree/Tree.spec.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+const Component: FC<{ nodes: TreeNode[] }> = ({ nodes }) => {
+    const [selectedId, setSelectedId] = useState<string | undefined>();
+    return <Tree nodes={nodes} selectedNodeId={selectedId} onSelect={(id: string) => setSelectedId(id)} />;
+};
+
+const TREE_ID = "[data-test-id=Tree]";
```
dwirz, line 19: we agreed on using kebap-case for `data-test-id`s

src/components/Tree/Tree.stories.tsx
```diff
@@ -0,0 +1,90 @@
...
+export default {
+    title: "Tree",
+    component: Tree,
+} as Meta;
+
+const TreeTemplate: Story<TreeProps> = (args) => <Tree {...args} />;
+
+export const Default = TreeTemplate.bind({});
+
+Default.args = {
```
dwirz, line 19: how about adding the other props? like `onSelect` and `selectedNodeId`

src/components/Tree/Tree.tsx
```diff
@@ -0,0 +1,27 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+import { ReactElement, useState } from "react";
+import Node, { TreeNode } from "./Node";
+import css from "./Tree.module.css";
+
+export interface TreeProps {
+    nodes: TreeNode[];
+    onSelect: (id: string) => void;
+    selectedNodeId?: string;
```
dwirz, line 10: why not using `activeNodeId` and rename it internally? this way it would follow the same "rule" as the slider and the dropdown of nadia

src/components/Tree/Tree.spec.tsx
triggertoo: 952bc7042225491892aaccd9af3d4533304d36b3

src/components/Tree/Tree.module.css
triggertoo: not yet

src/components/Tree/Tree.spec.tsx
triggertoo: 952bc7042225491892aaccd9af3d4533304d36b3

src/components/Tree/Tree.module.css
triggertoo: there isn't a variable for that one, I'm fine letting the user decide what it should be

src/components/Tree/Node.tsx
```diff
@@ -0,0 +1,91 @@
...
+    useEffect(() => setActive(id === activeNodeId), [activeNodeId]);
+
+    return (
+        <li data-test-id="node">
+            <a
+                className={getNodeClassNames(active, strong, !!value).join(" ")}
+                onClick={() => {
+                    if (value) {
+                        setActive(true);
+                        onClick && onClick(id);
```
SamuelAlev, line 58:
```suggestion
     return (
        <li data-test-id="node">
            <a
                className={getNodeClassNames(id === activeNodeId, strong, !!value).join(" ")}
                onClick={() => {
                    if (value) {
                        onClick && onClick(id);
```
I haven't tried but can't we bypass the `useEffect`?

src/components/Tree/Node.tsx
```diff
@@ -0,0 +1,91 @@
...
+        classNames.push(css.selectableNodeLink);
+    }
+    if (strong) {
+        classNames.push(css.strong);
+    }
+
+    return classNames;
+};
+
+export default function Node({
```
SamuelAlev, line 40: I wonder if it won't cause any conflict with [Node from the HTML API](https://developer.mozilla.org/docs/Web/API/Node) 😬

src/components/Tree/Tree.stories.tsx
triggertoo: e4ccece14380b7a3602047727defd88344e03feb

onSelect shows the selected id in storybook

src/components/Tree/Tree.tsx
triggertoo: 49065e5

src/components/Tree/Node.tsx
triggertoo: since it's used by the state, it's needed there

src/components/Tree/Node.tsx
triggertoo: to avoid any kind of conflict, i renamed it to TreeNode 6a5a786fe2e115deb012750be95160f80c31ad14  ;)

src/components/Tree/Tree.module.css
triggertoo: removed the max-width and just wrapped it in storybook: acd5ea3

src/components/Tree/Tree.module.css
dwirz: not sure if we need/want to add this to the css-vars? wdyt @SamuelAlev / @deniaz ?

src/components/Tree/Node.tsx
dwirz: @triggertoo I think @SamuelAlev is right since you pass in the `activeNodeId` and keep track of it outside within `Tree` you do not need to track it here with `useState` since if a new `activeNodeId` is passed in from outside the the `TreeNode` gets re-rendered.

src/components/Tree/Tree.tsx
```diff
@@ -0,0 +1,31 @@
...
+    onSelect: (id: string) => void;
+    activeNodeId?: string;
+}
+
+export default function Tree({
+    nodes,
+    onSelect,
+    activeNodeId: initialActiveNodeId,
+}: TreeProps): ReactElement<TreeProps> {
+    const [activeNodeId, setActiveNodeId] = useState<string | undefined>(initialActiveNodeId);
```
dwirz, line 18: here you need the `useEffect(() => setActiveNodeId(initialActiveNodeId), [initialActiveNodeId]);` logic otherwise the component does not rerender if you pass it a new `activeNodeId` 😉

src/components/Tree/Tree.tsx
```diff
@@ -0,0 +1,31 @@
...
+    const onClick = (id: string) => {
```
dwirz, line 19: why not inlining this function?

src/components/Tree/Node.tsx
triggertoo: ah, now i see it 👍  adaf607

src/components/Tree/Tree.tsx
triggertoo: thx, good catch: adaf607

src/components/Tree/Tree.tsx
triggertoo: done adaf607

PR #8, opened 2021-07-19
feat: Add Switch and Labelled Switch Components

deniaz:


src/components/Switch/Switch.tsx
```diff
@@ -0,0 +1,61 @@
...
+export type SwitchProps = {
+    on?: boolean;
+    disabled?: boolean;
+    label?: string;
+    name?: string;
+    size?: Size.Large | Size.Small;
+    onChange?: (e: MouseEvent) => void;
+};
+
+const merge = (classNames: (string | undefined | boolean)[]) => classNames.filter(Boolean).join(" ");
```
dwirz, line 18: nitty-gritty helper. how about extract it to a utility file and reuse it in the other please we do similar stuff? cc @SamuelAlev

SamuelAlev: Really great idea, remind me of [classnames](https://www.npmjs.com/package/classnames)

src/components/Switch/Switch.spec.tsx
```diff
@@ -0,0 +1,41 @@
...
+        cy.get("@Switch").should("have.value", "false");
+
+        cy.get("@Switch").click();
+        cy.get("@Switch").should("have.value", "true");
+    });
+
+    it("should do nothing if disabled", () => {
+        mount(<Component name="switch-test-disabled" disabled />);
+
+        cy.get("[data-test-id=Switch]").as("Switch");
```
SamuelAlev, line 30: We use kebab-case for those

src/components/Switch/Switch.tsx
deniaz: alright, done here: https://github.com/Frontify/arcade/pull/8/commits/cefe40c2b5089d3d87ebf13018faab91ebbfd61c

src/components/Switch/Switch.spec.tsx
deniaz: https://github.com/Frontify/arcade/pull/8/commits/7366c57ecc30e14003f3251c1bad7f8a70a5e7fd I adjusted the generator so it's now kebab-case by default ;)

src/components/Switch/Switch.stories.tsx
```diff
@@ -0,0 +1,46 @@
...
+    useEffect(() => {
+        setOn(args.on);
+    }, [args.on]);
+
+    return (
+        <Switch
+            {...args}
+            onChange={() => {
+                toggle();
+            }}
```
SamuelAlev, line 32:
```suggestion
            onChange={toggle}
```

src/components/Switch/Switch.tsx
```diff
@@ -0,0 +1,60 @@
...
+import Label from "../Form/Label/Label";
+import css from "./Switch.module.css";
+
+export type SwitchProps = {
+    on?: boolean;
+    disabled?: boolean;
+    label?: string;
+    name?: string;
+    size?: Size.Large | Size.Small;
+    onChange?: (e: MouseEvent) => void;
```
SamuelAlev, line 16:
```suggestion
    onChange?: (event: MouseEvent) => void;
```
nit: we decided to not use the shorthand `e` but `event` everywhere

SamuelAlev (approved):
Except the 2 little nit, works fine!

src/components/Switch/Switch.tsx
deniaz: @SamuelAlev are things like that documented somewhere like coding guidelines? cc @dwirz

dwirz (approved):

PR #13, opened 2021-07-22
feat: add story for icons

dwirz:


src/elements/Icon/IconTemplate.js
```diff
@@ -13,12 +13,11 @@ function IconTemplate({ template }, opts, { imports, interfaces, componentName,
       ${imports}
 
       import { IconSize } from '@elements/Icon/Icon';
-      import css from '@elements/Icon/Icon.module.css';
 
       ${interfaces}
 
       function ${componentName}(${props}) {
-        const customClassName = [css.icon, css[props.size || IconSize.Size16]].join(' ');
+        const customClassName = ["flex items-center justify-center fill-current stroke-current", IconSize[props.size || IconSize.Size16]].join(' ');
```
SamuelAlev, line 20: Does it automatically recognize the class? (to tree shake with jit)

dwirz: it goes through all your code and looks for occurences of tailwind classes for example `h-3` it should no matter where it occurs...

dwirz: I double checked the tailwind classes are within the "built" `index.css`

PR #17, opened 2021-07-26
feat: use tailwind classes within components

dwirz:


src/elements/Textarea/Textarea.tsx
```diff
@@ -30,16 +29,20 @@ export default function Textarea({
...
             )}
             <textarea
                 placeholder={placeholder}
                 required={required}
-                className={css.textarea}
+                className={`p-2 border rounded text-s outline-none transition placeholder-black-60 ${
+                    disabled
+                        ? "border-black-5 bg-black-5 text-black-40"
+                        : "text-black border-black-40 hover:border-black-90"
+                } ${decorator ? "pl-7 " : ""}`}
```
triggertoo, line 45: you could use the merge helper in that case so you don't have to fallback to empty string

src/elements/Divider/Divider.tsx
```diff
@@ -26,22 +24,11 @@ export type DividerProps = {
 export default function Divider({
     style = DividerStyle.Solid,
     height = DividerHeight.Height50,
-    color = "#CCC",
+    color: borderTopColor = "#CCC",
```
triggertoo, line 27: wouldn't it make sense to reuse an existing color in that case?

src/components/Slider/Slider.tsx
```diff
@@ -21,33 +20,27 @@ export type SliderProps = {
...
+        <ul
+            data-test-id="slider"
+            className="flex w-full justify-evenly p-0 border border-black-20 m-0 bg-black-0 rounded font-sans text-s list-none"
+        >
+            {items.map((item) => (
+                <li
+                    key={item.id}
+                    onClick={() => onChange(item.id)}
+                    className={`overflow-hidden flex-1 p-2 text-black-80 text-center overflow-ellipsis transition-colors whitespace-nowrap hover:text-black hover:cursor-pointer                    ${
+                        item.id === activeItemId
```
triggertoo, line 35: maybe use merge in this case?

deniaz (approved):

deniaz (approved):

src/elements/Divider/Divider.tsx
dwirz: atm i would leave it like that, since the devider isn't "designed" yet... so it could potentially change how we will use...

PR #19, opened 2021-07-26
feat: add framer motion to slider and dropdown

dwirz:


src/components/Dropdown/MenuItemContent/MenuItemContent.tsx
```diff
@@ -0,0 +1,84 @@
...
+          subtitle: string;
+          size: Size.Large;
+      });
+
+export type MenuItemContentProps = Omit<MenuItem, "id">;
+
+export default function MenuItemContent({
+    title,
+    icon,
+    subtitle = "",
```
triggertoo, line 40: not sure if we should set it empty, since it always will render the subtitle, or is that on purpose?

PR #21, opened 2021-07-26
fix: show form control helper text only once when set to before

dwirz:


src/compositions/FormControl/FormControl.spec.tsx
```diff
@@ -50,6 +50,8 @@ describe("FormControl Component", () => {
         );
 
         cy.get(FORM_CONTROL_HELPER_TEXT_ID).should("be.visible");
+        // check if the helper text is only presence once
```
triggertoo, line 53: // check if the helper text is only **present** once ;)

PR #27, opened 2021-07-29
feat(Fieldset Header): Add Fieldset Header Component

deniaz:


src/compositions/FieldsetHeader/FieldsetHeader.tsx
```diff
@@ -0,0 +1,105 @@
...
+    type?: Type;
+    disabled?: boolean;
+    bold?: boolean;
+    heading: string;
+    onClick?: () => void;
+    as?: keyof JSX.IntrinsicElements;
+};
+
+const renderType = (type: Type, id: string, size: Size, active: boolean, disabled: boolean) => {
+    if (type === Type.Switch) {
```
dwirz, line 39: why not using switch/case?

src/compositions/FieldsetHeader/FieldsetHeader.tsx
```diff
@@ -0,0 +1,105 @@
...
+}
+
+export type FieldsetHeaderProps = {
+    size?: Size;
+    active?: boolean;
+    decorator?: ReactNode;
+    type?: Type;
+    disabled?: boolean;
+    bold?: boolean;
+    heading: string;
```
dwirz, line 33: didn't we talk about using `children: string`?

src/compositions/FieldsetHeader/FieldsetHeader.tsx
deniaz: happy now? :smiling_face_with_three_hearts:

src/compositions/FieldsetHeader/FieldsetHeader.tsx
deniaz: good point, my attention span is currently roughly 30s, so I forgot. :facepalm:

PR #28, opened 2021-07-29
fix: slider styles and accessibility

NadiaPosch:


src/components/Slider/Slider.tsx
```diff
@@ -19,49 +24,74 @@ export type SliderProps = {
...
     return (
         <AnimateSharedLayout>
             <ul
                 data-test-id="slider"
                 className="w-full grid grid-flow-col auto-cols-fr justify-evenly p-0 border border-black-20 m-0 bg-black-0 rounded font-sans text-s list-none"
+                {...radioGroupProps}
             >
                 {items.map((item) => {
                     const isActive = item.id === activeItemId;
+                    const radioProps = {
```
dwirz, line 53: inline since only used once?

src/components/Slider/Slider.tsx
```diff
@@ -19,49 +24,74 @@ export type SliderProps = {
...
+                        value: item.id,
+                        "aria-label": isIconItem(item) ? item.ariaLabel : item.name,
```
dwirz, line 55: since `item.ariaLabel` is optional should we use a fallback like `item ${index}`?

NadiaPosch: For icon items it's not optional (only for the slider component) and for text items the name always has to be set as well, so there shouldn't be a case where a fallback is necessary.

package.json
```diff
@@ -107,6 +107,10 @@
         "react-router-dom": "^5.2.0"
     },
     "dependencies": {
+        "@react-aria/focus": "^3.4.0",
+        "@react-aria/radio": "^3.1.4",
+        "@react-aria/visually-hidden": "^3.2.2",
+        "@react-stately/radio": "^3.3.1",
```
SamuelAlev, line 113: You can install [react-aria](https://www.npmjs.com/package/react-aria) instead

dwirz (approved):

PR #30, opened 2021-07-30
feat(Badge): Add dark mode, styles, onClick and onDismiss

dwirz:


src/elements/Badge/Badge.tsx
```diff
@@ -44,7 +90,19 @@ const Badge: FC<BadgeProps> = ({ children, status, icon }: BadgeProps) => {
...
+                {onDismiss && (
+                    <button
+                        data-test-id="badge-icon"
+                        className={dismissClasses[style]}
+                        onClick={(event) => {
+                            event.stopPropagation();
+                            onDismiss();
+                        }}
+                    >
+                        <Reject size={IconSize.Size12} />
```
NadiaPosch, line 102: Shouldn't the icon disappear as soon as `onDismiss` is called? If the Badge works as a kind of selectable tag, this would be the expected behaviour in my opinion.

dwirz: since I do not know what the use case of the dismiss "button" is or how the badge is going to be used and also the design just states that it is possible to add a dismiss cross I would suggest to leave it like that for the moment.

src/elements/Badge/Badge.tsx
```diff
@@ -1,41 +1,87 @@
...
+
+const styleClasses: Record<Style, string> = {
+    [Style.Primary]: "bg-black-5 text-black-90 dark:text-black-10",
+    [Style.Positive]: "bg-green-20 text-green-90 dark:text-green-50",
+    [Style.Progress]: "bg-violet-20 text-violet-90 dark:text-violet-50",
+    [Style.Warning]: "bg-yellow-20 text-yellow-90 dark:text-yellow-50",
+    [Style.Danger]: "bg-red-20 text-red-90 dark:text-red-50",
+};
+
+const hoverStyleClasses: Record<Style, string> = {
```
NadiaPosch, line 38: I'd add a simple `transition-colors` here, wdyt?

src/elements/Badge/Badge.tsx
```diff
@@ -1,41 +1,87 @@
...
-                    "float-left h-6 bg-black-10 text-black-80 dark:bg-black-100 dark:text-white inline-flex items-center justify-center rounded-full",
-                    !children && (!status || !icon) ? "w-6" : "gap-x-1 pl-2 pr-[0.625rem]",
+                    "float-left h-6 inline-flex items-center justify-center rounded-full",
+                    !children && (!status || !icon)
+                        ? "w-6"
+                        : children && !icon && !status
+                        ? "gap-x-1 px-2"
+                        : "gap-x-1 pl-2 pr-[0.625rem]",
+                    styleClasses[style],
+                    onClick && `${hoverStyleClasses[style]} hover:cursor-pointer dark:hover:bg-black-superdark`,
```
NadiaPosch, line 83: In the Figma file there always seem to be hover styles, so I'd suggest to include them per default and only add `cursor-pointer` if `onClick` is defined

src/elements/Badge/Badge.stories.tsx
```diff
@@ -1,37 +1,79 @@
...
 
 export default {
     title: "Elements/Badge",
     component: Badge,
+    argTypes: {
+        style: {
+            options: Object.values(Style),
+            control: { type: "select" },
+        },
+    },
```
NadiaPosch, line 13:
```suggestion
    },
    args: {
        style: Style.Primary
    },
```

src/elements/Badge/Badge.tsx
```diff
@@ -1,41 +1,87 @@
...
+    Positive = "Positive",
+    Progress = "Progress",
+    Warning = "Warning",
+    Danger = "Danger",
+}
+
+const styleClasses: Record<Style, string> = {
+    [Style.Primary]: "bg-black-5 text-black-90 dark:text-black-10",
+    [Style.Positive]: "bg-green-20 text-green-90 dark:text-green-50",
+    [Style.Progress]: "bg-violet-20 text-violet-90 dark:text-violet-50",
```
NadiaPosch, line 33: The value for `violet-20` in the tailwind config (#e3d4f6) is different from the value in Figma (#E3E8F6). Maybe ask Johanna/responsible person which one to use and adjust accordingly

NadiaPosch: Also, the dark theme version is `violet-40` in Figma

src/elements/Badge/Badge.tsx
dwirz: 41d3718

src/elements/Badge/Badge.tsx
dwirz: 41d3718

src/elements/Badge/Badge.stories.tsx
dwirz: 41d3718

src/elements/Badge/Badge.tsx
dwirz: 41d3718

dwirz: thx for the color "bug" 👍  didn't notice the color difference 🙈  figma is single source of truth when it comes to colors and typography, that's why i  just the colors in the tailwind config

tailwind.config.js
```diff
@@ -58,7 +58,7 @@ module.exports = {
                 60: "#825fff",
```
NadiaPosch, line 58: `violet-90` is missing in the colour palette but is needed for the progress text color
```suggestion
                90: "#443185",
                70: "#6449c4",
                60: "#825fff",
```

dwirz: 83da75b

PR #38, opened 2021-07-30
fix(Text Input): Add additional stories and styles

dwirz:


src/elements/TextInput/TextInput.stories.tsx
```diff
@@ -1,28 +1,98 @@
...
 
 export default {
     title: "Elements/Text Input",
-    component: TextInputComponent,
+    component: TextInput,
     args: {
         clearable: false,
         disabled: false,
         required: false,
         type: TextInputType.Text,
```
NadiaPosch, line 15:
```suggestion
        type: TextInputType.Text,
        validation: Validation.Default,
```

src/elements/TextInput/TextInput.stories.tsx
```diff
@@ -1,28 +1,98 @@
...
     },
     argTypes: {
-        type: {
-            options: Object.values(TextInputType),
+        validation: {
+            options: Object.values(Validation),
             type: "select",
         },
-        decorator: { type: "string" },
         placeholder: { type: "string" },
```
NadiaPosch, line 22:
```suggestion
        placeholder: { type: "string" },
        defaultValue: { type: "string" },
```

src/elements/TextInput/TextInput.stories.tsx
```diff
@@ -1,28 +1,97 @@
...
+PasswordWithCharacterDecorator.args = {
+    type: TextInputType.Password,
+    defaultValue: "Secret",
+    decorator: "P",
+};
+
+export const PasswordVisible = TextInputTemplate.bind({});
+
+PasswordVisible.argTypes = {
+    obfuscated: { table: { disable: false } },
```
NadiaPosch, line 90: I'd remove the `PasswordVisible` story and instead add the `obfuscated` option to all other password stories, wdyt?

src/elements/TextInput/TextInput.stories.tsx
```diff
@@ -1,28 +1,97 @@
...
 
-export const TextInput: Story<TextInputProps> = (args: TextInputProps) => <TextInputComponent {...args} />;
+const TextInputTemplate: Story<TextInputProps> = (args) => <TextInput {...args} />;
+
+export const Text = TextInputTemplate.bind({});
+
+export const Prefilled = TextInputTemplate.bind({});
+
+Prefilled.args = {
+    defaultValue: "Value text",
```
NadiaPosch, line 42: Somehow when a `defaultValue` is set, it doesn't get updated if it's changed via the storybook control.

src/elements/TextInput/TextInput.stories.tsx
dwirz: per default the password is obfuscated, i added the story to show it is possible to set it to "visible" that's why i would keep this one, but adding the obfuscate option to the other stories is a good idea

src/elements/TextInput/TextInput.stories.tsx
dwirz: since the underlying `input` does not re-render if it is initially set with a `defaultValue` this does not work as expected i changed it to only allow to change the `defaultValue` if it is not set initially see b3ee7fc

PR #45, opened 2021-08-03
chore(Component Generator): Use Named Exports for Components

deniaz:


scripts/templates/component.ts
```diff
@@ -4,15 +4,15 @@ import ComponentFileBuilderResponse from "./ComponentFileBuilderResponse";
...
+import { FC } from "react";
 
-export interface ${componentName}Props {
+export type ${componentName}Props = {
     foo: string;
-}
+};
 
-export default function ${componentName}({ foo }: ${componentName}Props): ReactElement<${componentName}Props> {
+export const ${componentName}: FC<${componentName}Props> = ({ foo }) => {
```
SamuelAlev, line 13:
```suggestion
export const ${componentName}: FC<${componentName}Props> = ({ foo }: ${componentName}Props) => {
```

scripts/templates/component.stories.ts
```diff
@@ -6,17 +6,22 @@ export default (componentName: string, componentType: string): ComponentFileBuil
...
+export const Default = Template.bind({});
+
+Default.args = {
     args: {
         foo: "bar",
     },
-} as Meta<${componentName}Props>;
-
-export const ${componentName}: Story<${componentName}Props> = (args) => <${componentName}Component {...args} />;
+};
```
SamuelAlev, line 24: One `args` too much :)

.eslintrc.js
```diff
@@ -6,4 +6,7 @@ module.exports = {
         },
     },
     plugins: ["jsx-a11y"],
+    rules: {
+        "react/prop-types": "off",
+    },
```
SamuelAlev, line 11: Can be removed

dwirz: works 👍  removed the rule ee61723

scripts/templates/component.ts
dwirz: no need since the `FC` type ensures that props have the correct typing

PR #47, opened 2021-08-04
refactor: rename key to `FieldsetHeader` to prevent overlaps

SamuelAlev:


src/compositions/FieldsetHeader/FieldsetHeader.stories.tsx
```diff
@@ -2,26 +2,31 @@
...
             control: { type: "radio" },
         },
         type: {
-            options: [Type.Default, Type.Switch, Type.Accordion, Type.AddRemove],
+            options: [
+                FieldsetHeaderType.Default,
+                FieldsetHeaderType.Switch,
+                FieldsetHeaderType.Accordion,
+                FieldsetHeaderType.AddRemove,
+            ],
```
dwirz, line 22: you could use `Object.values(FieldsetHeaderType)`

PR #67, opened 2021-08-06
chore: add prefix to tailwind

SamuelAlev:


src/compositions/Accordion/Accordion.tsx
```diff
@@ -72,8 +72,7 @@ const AriaAccordionItem: FC<AriaAccordionItemProps> = ({ item, state, header })
                 >
                     <div
                         {...regionProps}
-                        className="px-8 pb-7 mt--1"
-                        style={{ visibility: isOpen ? "visible" : "hidden" }}
+                        className={`tw-px-8 tw-pb-7 tw-mt-1 ${isOpen ? "tw-visible" : "tw-invisible"}`}
```
SamuelAlev, line 75: > Tailwind is smart enough to generate classes like -mx-72 when it sees the leading dash, not mx--72 like you might expect.

dwirz: image
actually it's not a typo 😉

dwirz: i use to not need to change the y-padding on the accordion header, because this lead to a jumping effect with framer-motion 🙈

dwirz: why i needed to this, compare the whole header/accordion with the figma design...

src/compositions/Accordion/Accordion.tsx
```diff
@@ -72,7 +72,7 @@ const AriaAccordionItem: FC<AriaAccordionItemProps> = ({ item, state, header })
                 >
                     <div
                         {...regionProps}
-                        className={`tw-px-8 tw-pb-7 tw-mt-1 ${isOpen ? "tw-visible" : "tw-invisible"}`}
+                        className={`tw-px-8 tw-pb-7 tw--mt-1 ${isOpen ? "tw-visible" : "tw-invisible"}`}
```
dwirz, line 75: shouldn't it be `tw-mt--1`? 🤔

SamuelAlev: >Tailwind is smart enough to generate classes like -mx-72 when it sees the leading dash, not mx--72 like you might expect.

Nope

dwirz: nice 👍

PR #91, opened 2021-08-13
fix(Checkbox): scale and tooltip aria

NadiaPosch:
Clickup Ticket: https://app.clickup.com/t/pq8qe3

TODO: add Tooltip focus styles

### DIFF
--- a/src/elements/Checkbox/Checkbox.tsx
+++ b/src/elements/Checkbox/Checkbox.tsx
@@ -1,6 +1,5 @@
 /* (c) Copyright Frontify Ltd., all rights reserved. */
 
-import React, { FC, useRef } from "react";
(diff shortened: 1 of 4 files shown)


src/elements/Checkbox/Checkbox.tsx
```diff
@@ -55,7 +55,7 @@ const isCheckedOrMixed = (checked: CheckboxState): boolean => {
 };
 
 export const Checkbox: FC<CheckboxProps> = (props) => {
-    const id = props.id || generateRandomId();
+    const [id] = useState(props.id || generateRandomId());
```
deniaz, line 58: Same question here: Why wrap it in a state?

PR #99, opened 2021-08-17
test: re-activate tests for Dropdown and Menu

dwirz:


rollup.config.ts
```diff
@@ -59,6 +59,7 @@ const rollupConfig = [
     }),
     bundle({
         plugins: [dts()],
+        external: ["tailwindcss/tailwind.css"],
```
triggertoo, line 62: do we need the css when creating the type definition files?

dwirz: added this to solve the warning when generating the typings 😉 
```
src/index.ts → dist/index.d.ts...
(!) Unresolved dependencies
https://rollupjs.org/guide/en/#warning-treating-module-as-external-dependency
tailwindcss/tailwind.css (imported by src/index.ts)
created dist/index.d.ts in 6.3s
```

triggertoo: 👍

PR #108, opened 2021-08-20
feat(Flyout): add basic component

NadiaPosch:


triggertoo:
Badges seem to have the wrong background color in dark mode, but I'm also fine when we fix that in another MR:
badges

src/components/Flyout/Flyout.tsx
```diff
@@ -0,0 +1,165 @@
...
+            >
+                {title && (
+                    <>
+                        <div className="tw-flex tw-justify-between tw-flex-wrap tw-gap-3 tw-p-8">
+                            <div {...titleProps} className="tw-inline-flex">
+                                <FieldsetHeader decorator={decorator}>{title}</FieldsetHeader>
+                            </div>
+                            <div className="tw-inline-flex tw-gap-2 tw-flex-wrap">
+                                {badges &&
+                                    badges.map((badge, index) => <Badge key={`flyout-badge-${index}`} {...badge} />)}
```
triggertoo, line 65: rename badge to badgeProps

```suggestion
                                    badges.map((badgeProps, index) => <Badge key={`flyout-badge-${index}`} {...badgeProps} />)}
```

src/components/Flyout/Flyout.tsx
```diff
@@ -0,0 +1,165 @@
...
+                )}
+                {Children.map(children, (child, index) => (
+                    <>
+                        {index !== 0 && <Divider color={FLYOUT_DIVIDER_COLOR} height={DividerHeight.Height10} />}
+
+                        <div key={index} className="tw-p-8">
+                            {child}
+                        </div>
+
+                        {index === Children.count(children) - 1 && (
```
triggertoo, line 79: I'd append the Divider after each block/iteration, so you can actually remove line 73

dwirz:
> Badges seem to have the wrong background color in dark mode, but I'm also fine when we fix that in another MR:
> badges

the flyout does not have a dark "theme" atm so the bg of the flyout is probably not optimal

triggertoo (commented):
noice

PR #145, opened 2021-09-08
feat: hide clear icon when input is empty

SamuelAlev:
Simone asked to hide the clear icon when there is no text in the input


oliverschwendener:
Code looks good 👍 

A issue that I noticed: the clear icon will hide when you remove all text from the input. If you clear the text via clear icon then the icon will stay visible. Is this intended behavior?

oliverschwendener:
Here is a screen recording to show what I mean:

https://user-images.githubusercontent.com/15727229/132666580-4e17dba7-864d-4562-aba1-fec48ed47bd4.mov

oliverschwendener (approved):
Works as expected 💪

PR #160, opened 2021-09-14
fix: Accordion focus styles and selectable input

NadiaPosch:


SamuelAlev (approved):
Really nice, thanks for that!

PR #165, opened 2021-09-15
feat(Tooltip): Add Component

CelienB:
This PR updates the Tooltip v2 component along with its related stories.


src/components/Tooltip/Tooltip.spec.tsx
```diff
@@ -12,12 +12,9 @@ describe("Tooltip Component", () => {
     it("should render a tooltip", () => {
         mount(<Tooltip tooltip={TOOLTIP_TEXT} />);
 
-        cy.get(TOOLTIP_ICON_ID).should("be.visible");
-        cy.get(TOOLTIP_ICON_ID).trigger("mouseover");
-
         cy.get(TOOLTIP_ID).should("contain", TOOLTIP_TEXT);
     });
-    it("should focus with keyboard", () => {
+    it.skip("should focus with keyboard", () => {
```
NadiaPosch, line 17: I suspect this is just an overlook :)
```suggestion
    it("should focus with keyboard", () => {
```

src/components/Tooltip/Tooltip.stories.tsx
```diff
@@ -3,18 +3,139 @@
...
+    },
+    argTypes: {
+        brightHeader: {
+            options: [
+                null,
+                BrightHeaderVariants.Information,
+                BrightHeaderVariants.Warning,
+                BrightHeaderVariants.Tip,
+                BrightHeaderVariants.Note,
+            ],
```
SamuelAlev, line 25: You could do the same mapping as the following `argType` so we have better info than `null` when selecting the header type

src/components/Tooltip/Tooltip.stories.tsx
```diff
@@ -3,18 +3,139 @@
...
+import IconIcons from "@elements/Icon/Generated/IconIcons";
 
 export default {
     title: "Components/Tooltip",
     component: TooltipComponent,
     args: {
-        tooltip: "Tooltip Content (Reactnode)",
+        tooltip: "Your text here",
+        heading: "",
+        buttons: [],
```
NadiaPosch, line 15: I'd remove the default value for `heading` and `buttons` as you've got all possibilities nicely covered in the different stories.

src/components/Tooltip/Tooltip.stories.tsx
```diff
@@ -3,18 +3,139 @@
...
-        <p>Some other text.</p>
-    </div>
-);
+const loremIpsum = "Cupcake ipsum dolor sit amet ice cream.";
+
+export const TooltipTemplate: Story<TooltipProps> = (args: TooltipProps) => <TooltipComponent {...args} />;
+
+export const Default = TooltipTemplate.bind({});
+Default.args = {
+    brightHeader: undefined,
```
NadiaPosch, line 53: You don't have to set this explicitly if it's not needed. Same for the `buttons` control below.

src/components/Tooltip/Tooltip.tsx
```diff
@@ -1,69 +1,81 @@
...
+    tooltip,
+    tooltipIcon,
+    heading,
+    headingIcon,
+    linkUrl,
+    linkLabel,
+    brightHeader,
+    buttons,
+}) => {
+    const showButtons = buttons !== undefined && Object.keys(buttons).length === 2;
```
SamuelAlev, line 34:
```suggestion
    buttons = [],
}) => {
    const showButtons = Object.keys(buttons).length === 2;
```

It can only be `undefined | [TooltipButton, TooltipButton]`
`undefined` is falsy and `[TooltipButton, TooltipButton]` is truthy

src/components/Tooltip/Tooltip.tsx
```diff
@@ -1,69 +1,81 @@
...
 
 export type TooltipProps = {
     tooltip: ReactNode;
+    tooltipIcon?: ReactElement;
+    heading?: ReactNode;
+    headingIcon?: ReactElement;
+    linkUrl?: string;
+    linkLabel?: string;
+    brightHeader?: BrightHeaderVariants;
+    buttons?: [TooltipButton, TooltipButton];
```
NadiaPosch, line 21: I'd suggest typing it as follows and only displaying the first two buttons as you've already done, wdyt?

```suggestion
    buttons?: TooltipButton[];
```

src/components/Tooltip/Tooltip.tsx
NadiaPosch, line 34: Since you only use it once there's no need to define a variable

src/components/Tooltip/Tooltip.tsx
```diff
@@ -1,69 +1,81 @@
...
+                        <a
+                            href={linkUrl}
+                            target="_blank"
+                            rel="noopener noreferrer"
+                            className="tw-text-xs tw-text-black-40 dark:tw-text-black-80 tw-underline tw-mt-1"
+                        >
+                            {linkLabel ?? "Click here to learn more."}
+                        </a>
+                    )}
+                    {showButtons && (
```
NadiaPosch, line 67:
```suggestion
                    {buttons?.length > 0 && (
```

src/components/Tooltip/partials/BrightHeader.tsx
```diff
@@ -0,0 +1,47 @@
...
+    [BrightHeaderVariants.Warning]: "tw-bg-red-60",
+    [BrightHeaderVariants.Tip]: "tw-bg-green-60",
+    [BrightHeaderVariants.Note]: "tw-bg-yellow-60",
+};
+
+const brightHeaderIcon: Record<BrightHeaderVariants, ReactElement> = {
+    [BrightHeaderVariants.Information]: <IconInfo size={IconSize.Size20} />,
+    [BrightHeaderVariants.Warning]: <IconCallout size={IconSize.Size20} />,
+    [BrightHeaderVariants.Tip]: <IconCheck size={IconSize.Size20} />,
+    [BrightHeaderVariants.Note]: <IconDocument size={IconSize.Size20} />,
```
SamuelAlev, line 29:
```suggestion
    [BrightHeaderVariants.Information]: <IconInfo />,
    [BrightHeaderVariants.Warning]: <IconCallout />,
    [BrightHeaderVariants.Tip]: <IconCheck />,
    [BrightHeaderVariants.Note]: <IconDocument />,
```

As you already do `cloneElement(headingIcon, { size: IconSize.Size20 })`

src/components/Tooltip/Tooltip.tsx
SamuelAlev: The current one has the advantage to force the prop to have 2 element of `TooltipButton` type or `undefined`

src/components/Tooltip/Tooltip.tsx
```diff
@@ -1,69 +1,81 @@
...
+                        <div className="tw-grid tw-grid-cols-2 tw-gap-x-1 tw-mt-4">
+                            <Button style={ButtonStyle.Secondary} size={ButtonSize.Small} onClick={buttons[0].action}>
+                                {buttons[0].label}
+                            </Button>
+                            <Button style={ButtonStyle.Primary} size={ButtonSize.Small} onClick={buttons[1].action}>
+                                {buttons[1].label}
+                            </Button>
```
NadiaPosch, line 74: I see two cases that should be covered here:
- there's only one button defined which would be displayed _primary_
- there are >= 2 buttons, in which case the first would be displayed _primary_ and the second _secondary_

src/components/Tooltip/partials/BrightHeader.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+import React, { FC, ReactElement } from "react";
+import { merge } from "@utilities/merge";
+import { IconSize } from "@elements/Icon/IconSize";
+import IconInfo from "@elements/Icon/Generated/IconInfo";
+import IconCheck from "@elements/Icon/Generated/IconCheck";
+import IconCallout from "@elements/Icon/Generated/IconCallout";
+import IconDocument from "@elements/Icon/Generated/IconDocument";
+
+export enum BrightHeaderVariants {
```
NadiaPosch, line 11: We haven't used variants in the other components, so I'd suggest renaming this to `BrightHeaderStyle` to keep it more consistent and capitalise the values (see e.g. `BadgeStyle`)

src/components/Tooltip/Tooltip.tsx
NadiaPosch: Is it possible to only define one button which is then primary? At the moment this possibility is not covered.

SamuelAlev: The design don't show anything, maybe you can ask Simone @CelienB

src/components/Tooltip/Tooltip.spec.tsx
CelienB: I decided to skip it for the time being as I didn't get around to adding a11y functionalities yet. I will update it once this is done.

src/components/Tooltip/Tooltip.tsx
CelienB: There are always two buttons on the mockups. I'd have to check with the designer first.

src/components/Tooltip/Tooltip.tsx
CelienB: Do you mean moving the condition inside the template instead? Wouldn't that be harder to read then? Why not, though.

src/components/Tooltip/Tooltip.tsx
CelienB: On the mockups, the secondary button comes first and the primary one comes second. Did I understand that correctly?

src/components/Tooltip/Tooltip.tsx
SamuelAlev: You shouldn't need it anymore with her next comment (`{buttons?.length > 0 && (`)

src/components/Tooltip/Tooltip.tsx
NadiaPosch: Yes, I just thought that from a developers perspective I'd put the button I wanted to be primary first in the array and the secondary one in second place. In the case where only one button is defined, I'd expect it to be primary and pass in either `[button]` or `[button, undefined]`.

CelienB: That makes sense. Simone also told me she doesn't expect a situation where there's only one button displayed. Should we be proactive and make it future-proof by allowing this specific use-case, though?

src/components/Tooltip/Tooltip.spec.tsx
```diff
@@ -2,18 +2,112 @@
...
+        mount(<Tooltip tooltip={TOOLTIP_TEXT} brightHeader={BrightHeaderStyle.Tip} />);
+
+        cy.get(BRIGHT_HEADER_ID).should("exist");
+    });
+
+    it("should renders a bright header (note)", () => {
+        mount(<Tooltip tooltip={TOOLTIP_TEXT} brightHeader={BrightHeaderStyle.Note} />);
+
+        cy.get(BRIGHT_HEADER_ID).should("exist");
+    });
```
SamuelAlev, line 92:
```suggestion
    Object.values(BrightHeaderStyle).forEach((brightHeaderStyle) => it(`should renders a bright header (${brightHeaderStyle})`, () => {
        mount(<Tooltip tooltip={TOOLTIP_TEXT} brightHeader={brightHeaderStyle} />);

        cy.get(BRIGHT_HEADER_ID).should("exist");
    }))
```

src/components/Tooltip/Tooltip.spec.tsx
```diff
@@ -2,22 +2,113 @@
...
-import { Tooltip } from "./Tooltip";
+import { Tooltip, TooltipButton } from "./Tooltip";
+import IconIcons from "@elements/Icon/Generated/IconIcons";
+import { BrightHeaderStyle } from "./BrightHeader";
 
 const TOOLTIP_TEXT = "This is a tooltip";
 const TOOLTIP_ICON_ID = "[data-test-id=tooltip-icon]";
 const TOOLTIP_ID = "[data-test-id=tooltip]";
+const TOOLTIP_LINK_ID = "[data-test-id=tooltip-link]";
+const TOOLTIP_LINK_URL = "http://www.frontify.com";
```
SamuelAlev, line 13:
```suggestion
const TOOLTIP_LINK_URL = "https://www.frontify.com";
```
To fix SonarCloud issue

src/components/Tooltip/Tooltip.stories.tsx
```diff
@@ -3,18 +3,133 @@
...
-        <p>Some other text.</p>
-    </div>
-);
+const loremIpsum = "Cupcake ipsum dolor sit amet ice cream.";
+
+export const TooltipTemplate: Story<TooltipProps> = (args: TooltipProps) => <TooltipComponent {...args} />;
+
+export const Default = TooltipTemplate.bind({});
+Default.args = {
+    tooltip: loremIpsum,
```
SamuelAlev, line 56: You defined `tooltip` on the global level (L13) so you can remove all `tooltip` from the stories

src/components/Tooltip/Tooltip.spec.tsx
```diff
@@ -2,22 +2,113 @@
...
+
+    it("should render a link with custom text", () => {
+        const linkLabel = "Upgrade your plan.";
+
+        mount(<Tooltip tooltip={TOOLTIP_TEXT} linkUrl={TOOLTIP_LINK_URL} linkLabel={linkLabel} />);
+
+        cy.get(TOOLTIP_LINK_ID).should("be.visible").and("contain", linkLabel);
+    });
+
+    it("should not render the heading if the heading icon is missing", () => {
```
SamuelAlev, line 45: This shouldn't happen you may want to have the heading without any icon

src/components/Tooltip/Tooltip.tsx
```diff
@@ -1,69 +1,80 @@
...
+                {brightHeader && <BrightHeader headerStyle={brightHeader} />}
+                <div data-test-id="tooltip" className="tw-p-4">
+                    {heading && headingIcon && (
+                        <h4 className="tw-flex tw-text-m tw-font-bold tw-mb-1">
+                            <span className="tw-mr-1.5">
+                                {headingIcon && cloneElement(headingIcon, { size: IconSize.Size20 })}
+                            </span>
+                            {heading}
+                        </h4>
+                    )}
```
SamuelAlev, line 46:
```suggestion
                    {heading && (
                        <h4 className="tw-flex tw-text-m tw-font-bold tw-mb-1">
                            <span className="tw-mr-1.5">
                                {headingIcon && cloneElement(headingIcon, { size: IconSize.Size20 })}
                            </span>
                            {heading}
                        </h4>
                    )}
```

This shouldn't happen you may want to have the heading without any icon (this behavior already have been asked to Simone for previous components)

NadiaPosch:
The Tooltip is also used in the `InputLabel` component. Please make sure to update it as well and use the appropriate react-aria hooks `useTooltipTrigger` and `useTooltip` as they are implemented in the current tooltip.

src/components/Tooltip/Tooltip.stories.tsx
```diff
@@ -3,18 +3,129 @@
...
     title: "Components/Tooltip",
     component: TooltipComponent,
     args: {
-        tooltip: "Tooltip Content (Reactnode)",
+        tooltip: "Cupcake ipsum dolor sit amet ice cream.",
+        heading: "",
+    },
+    argTypes: {
+        brightHeader: {
+            options: ["None", "Information", "Warning", "Tip", "Note"],
```
NadiaPosch, line 18: Right now, it doesn't show which option is selected, this should fix it:
```suggestion
            options: ["None", ...Object.values(BrightHeaderStyle)],
```

src/components/Tooltip/Tooltip.spec.tsx
```diff
@@ -3,27 +3,145 @@
...
-        cy.get(TOOLTIP_ICON_ID).trigger("mouseover");
-
         cy.get(TOOLTIP_ID).should("contain", TOOLTIP_TEXT);
     });
-    it("should focus with keyboard", () => {
+
+    it("should render an icon next to the tooltip", () => {
+        mount(<Tooltip tooltip={TOOLTIP_TEXT} tooltipIcon={GENERIC_ICON} />);
+
+        cy.get("svg[name=IconIcons]").should("be.visible");
```
NadiaPosch, line 27: You could create a reusable variable, e.g. `const SVG_ICON_NAME = "svg[name=IconIcons]"`

src/components/Tooltip/Tooltip.spec.tsx
```diff
@@ -3,27 +3,145 @@
...
+    });
+
+    it("should render the heading and the heading icon when both are present", () => {
+        mount(<Tooltip tooltip={TOOLTIP_TEXT} heading={TOOLTIP_HEADING_TEXT} headingIcon={GENERIC_ICON} />);
+
+        cy.get(TOOLTIP_ID).should("contain.text", TOOLTIP_HEADING_TEXT);
+        cy.get("svg[name=IconIcons]").should("be.visible");
+    });
+
+    it("should not render a bright header by default", () => {
```
NadiaPosch, line 57: I'd combine this with the first one and maybe call it _should render a default tooltip_

src/components/Tooltip/Tooltip.spec.tsx
```diff
@@ -3,27 +3,145 @@
...
         mount(<Tooltip tooltip={TOOLTIP_TEXT} />);
 
-        cy.get(TOOLTIP_ICON_ID).should("not.be.focused");
-        cy.get(TOOLTIP_ID).should("not.exist");
+        cy.get(BRIGHT_HEADER_ID).should("not.exist");
+    });
+
+    Object.values(BrightHeaderStyle).forEach((brightHeaderStyle) => {
+        it(`should render a bright header (${brightHeaderStyle})`, () => {
```
NadiaPosch, line 64: You don't test if the style is actually correct. You could for example check if the icon or the color class are the right ones. Or leave it like this but only check for one style if the header id exists.

src/components/Tooltip/Tooltip.spec.tsx
```diff
@@ -3,27 +3,145 @@
...
+            mount(<Tooltip tooltip={TOOLTIP_TEXT} brightHeader={brightHeaderStyle} />);
+
+            cy.get(BRIGHT_HEADER_ID).should("exist");
+        });
+    });
+
+    it("should not render buttons by default", () => {
```
NadiaPosch, line 71: Same here, I'd move this to the first test

src/components/Tooltip/Tooltip.spec.tsx
```diff
@@ -3,27 +3,145 @@
...
+            <Tooltip
+                tooltip={TOOLTIP_TEXT}
+                buttons={[
+                    { label: "Primary", action: () => null },
+                    { label: "Secondary", action: () => null },
+                ]}
+            />,
+        );
+
+        cy.get("button").should("exist");
```
NadiaPosch, line 94:
```suggestion
        cy.get("button").should("have.length", 2);
```

src/components/Tooltip/Tooltip.spec.tsx
CelienB: Alright, I wasn't sure if it was a legit way to check if a specific icon is visible or if there was a more elegant way to do it.

src/components/Tooltip/Tooltip.spec.tsx
```diff
@@ -3,27 +3,145 @@
...
+    it("should not render buttons by default", () => {
+        mount(<Tooltip tooltip={TOOLTIP_TEXT} />);
+
+        cy.get("button").should("not.exist");
+    });
+
+    it("should render one button", () => {
+        mount(<Tooltip tooltip={TOOLTIP_TEXT} buttons={[{ label: "Primary", action: () => null }]} />);
+
+        cy.get("button").should("exist");
```
NadiaPosch, line 80:
```suggestion
        cy.get("button").should("have.length", 1);
```

src/components/Tooltip/Tooltip.spec.tsx
CelienB: You mean you would move the assertions to the default tooltip test and not specifically have a test for this use case?

src/components/Tooltip/Tooltip.spec.tsx
```diff
@@ -3,27 +3,145 @@
...
+                    { label: "Primary", action: () => null },
+                    { label: "Secondary", action: () => null },
+                ]}
+            />,
+        );
+
+        cy.get("button").should("exist");
+    });
+
+    it("should focus the link using a keyboard", () => {
```
NadiaPosch, line 97: I'd remove this in favour of the last one which tests the same thing

src/components/Tooltip/Tooltip.spec.tsx
```diff
@@ -3,27 +3,145 @@
...
+
+        cy.get(TOOLTIP_LINK_ID).should("be.visible");
+
         cy.get("body").tab();
-        cy.get(TOOLTIP_ICON_ID).should("be.focused");
-        cy.get(TOOLTIP_ID).should("exist");
+        cy.get(TOOLTIP_LINK_ID).should("be.focused");
+    });
+
+    it("should focus the buttons using a keyboard", () => {
```
NadiaPosch, line 106: I'd remove this in favour of the last one which tests the same thing

src/components/Tooltip/Tooltip.spec.tsx
```diff
@@ -3,27 +3,145 @@
...
+        cy.get("svg[name=IconIcons]").should("be.visible");
+    });
+
+    it("should render a link", () => {
+        mount(<Tooltip tooltip={TOOLTIP_TEXT} linkUrl={TOOLTIP_LINK_URL} />);
+
+        cy.get(TOOLTIP_LINK_ID).should("be.visible").and("have.attr", "href").and("include", TOOLTIP_LINK_URL);
+    });
+
+    it("should render a link with custom text", () => {
```
NadiaPosch, line 36: I'd combine this with the one above

src/elements/Button/Button.spec.tsx
```diff
@@ -51,6 +51,28 @@ describe("Button component", () => {
         });
     });
 
+    it(`renders in solid Primary inverted with only text.`, () => {
```
NadiaPosch, line 54: You have the same test twice. Maybe add a data-test-id that reflects if the button is inverted or check if the inverted styles are correct.

src/elements/Button/Button.stories.tsx
```diff
@@ -17,6 +17,11 @@ export default {
             options: [ButtonStyle.Primary, ButtonStyle.Secondary, ButtonStyle.Danger, ButtonStyle.Positive],
             control: { type: "select" },
         },
+        inverted: {
```
NadiaPosch, line 20: If you add it as an arg storybook automatically renders a toggle:
```
args: {
        inverted: false,
    },
```

src/components/Tooltip/Tooltip.spec.tsx
NadiaPosch: Yes exactly

src/components/Tooltip/BrightHeader.tsx
```diff
@@ -0,0 +1,48 @@
...
+import React, { cloneElement, FC, ReactElement } from "react";
+import { merge } from "@utilities/merge";
+import { IconSize } from "@elements/Icon/IconSize";
+import IconInfo from "@elements/Icon/Generated/IconInfo";
+import IconCheck from "@elements/Icon/Generated/IconCheck";
+import IconCallout from "@elements/Icon/Generated/IconCallout";
+import IconDocument from "@elements/Icon/Generated/IconDocument";
+
+export enum BrightHeaderStyle {
+    Information = "information",
```
NadiaPosch, line 12: Please use upper case letters for enums, e.g. `Information = "Information"`

src/components/Tooltip/TooltipArrow.css
```diff
@@ -0,0 +1,15 @@
+.arcade-tooltip[data-popper-placement^="top"] > .arcade-tooltip-arrow {
```
NadiaPosch, line 1: Please move custom styles into `src/styles.css` and maybe add a `tw-` prefix to keep it consistent

src/components/Tooltip/TooltipArrow.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+export type TooltipArrowProps = {
+    style: CSSProperties;
+    headerColor?: BrightHeaderStyle;
+    placement?: string;
+};
+
+export const TooltipArrow = forwardRef<HTMLDivElement, TooltipArrowProps>(({ style, headerColor, placement }, ref) => {
+    // We want the arrow to be the same color as the header
+    // when tooltip is positionned at the bottom
```
NadiaPosch, line 16: As the code is self-explanatory I think you don't need the comment here

PR #168, opened 2021-09-16
fix: TextInput styling fixes

NadiaPosch:


src/elements/TextInput/TextInput.tsx
```diff
@@ -12,7 +12,7 @@ import React, { FC, FormEvent, KeyboardEvent, ReactElement, ReactNode, useEffect
 export enum TextInputType {
     Text = "text",
     Password = "password",
-    Number = "number",
+    Number = "numeric",
```
SamuelAlev, line 15: `number` is the correct type, no?
If this to remove the number increment arrows then I would go with CSS
https://www.w3schools.com/howto/howto_css_hide_arrow_number.asp

deniaz: another way would be to use the React TypeScript definition for the input type attribute:

```
type?: InputHTMLAttributes<HTMLInputElement>["type"];
```

(in the prop types, rather than enum)

SamuelAlev: @deniaz IMO we should not go with this solution as everything related to styles or sizes already is an `enum`, it wouldn't make sense to have this one as `string`.

src/styles.css
```diff
@@ -0,0 +1,15 @@
...
+@tailwind utilities;
+
+@layer utilities {
+    input.hide-input-arrows::-webkit-outer-spin-button,
+    input.hide-input-arrows::-webkit-inner-spin-button {
+        -webkit-appearance: none;
+        margin: 0;
+    }
+
+    input[type="number"].hide-input-arrows {
```
SamuelAlev, line 12:
```suggestion
    input.tw-hide-input-arrows::-webkit-outer-spin-button,
    input.tw-hide-input-arrows::-webkit-inner-spin-button {
        -webkit-appearance: none;
        margin: 0;
    }

    input[type="number"].tw-hide-input-arrows {
```

What do you think about adding the prefix? 
[Btw no automatic way to do it](https://github.com/tailwindlabs/tailwindcss/discussions/2128)

SamuelAlev (approved):

src/styles.css
NadiaPosch: Done 👌

SamuelAlev (approved):

deniaz (approved):

PR #170, opened 2021-09-17
style: remove fill and fillRule properties from svg

jprats89:
**Issue**
IconLightningFilled is not displayed. It can be checked in storybook

**Why**
I guess `fill='none'` path property is not helping

**Solution**
Remove this property

**Notes for the reviewer**
I also removed the `fillRule` property since I did not see it in other Icons, I guess that's fine


SamuelAlev (approved):
I replicated the change you did on the generated file to the main .svg one
Everything looks good now 👍

PR #182, opened 2021-09-22
fix: breadcrumbs accessibility

NadiaPosch:


src/components/Breadcrumbs/CurrentBreadcrumbItem.tsx
```diff
@@ -41,10 +63,18 @@ export const CurrentBreadcrumbItem: FC<Breadcrumb> = ({ label, badges, bold, dec
                 <>
                     {decorator}
                     <ItemWithBadges badges={badges}>
-                        <span className={bold ? "tw-font-bold" : ""}>{label}</span>
+                        <span
+                            ref={ref as RefObject<HTMLSpanElement>}
+                            {...props}
+                            className={bold ? "tw-font-bold" : isFocusVisible ? FOCUS_STYLE : ""}
```
triggertoo, line 69: maybe we should rename bold to active, so that the focus is not depending on bold

PR #196, opened 2021-09-23
fix(Accordion): focus accordion items and surround content

triggertoo:


src/compositions/Accordion/Accordion.tsx
```diff
@@ -27,7 +27,7 @@ const AriaAccordionItem: FC<AriaAccordionItemProps> = ({ item, state, header })
     const { isFocusVisible, focusProps } = useFocusRing();
 
     return (
-        <div key={item.key}>
+        <div key={item.key} className={isFocusVisible ? FOCUS_STYLE_INSET : ""}>
```
deniaz, line 30:
```suggestion
        <div key={item.key} className={isFocusVisible && FOCUS_STYLE_INSET}>
```

triggertoo: this would result in `undefined`, since it's not within the merge helper

PR #198, opened 2021-09-23
fix(Color Picker): Update Input Values When Changing Color With Drag and Drop

deniaz:
⚠️ **BREAKING CHANGES INCOMING** 

The `TextInput` was an uncontrolled component, which means that it handled the input value itself. This way we could only pass the `defaultValue`. When this one was updated, the changes weren't reflected in the input itself.

In our experience with React (and according to the docs) in most cases controlled form components are preferred over uncontrolled. Therefore this change does exactly this.

More info:

- Uncontrolled Components: https://reactjs.org/docs/uncontrolled-components.html
- Controlled Components: https://reactjs.org/docs/forms.html#controlled-components

### DIFF
--- a/src/elements/TextInput/TextInput.tsx
+++ b/src/elements/TextInput/TextInput.tsx
@@ -52,8 +52,8 @@ type TextInputBaseProps = {
     validation?: Validation;
     value?: string;
     onChange?: (value: string) => void;
-    onEnterPressed?: (e: KeyboardEvent<HTMLInputElement>) => void;
-    onBlur?: (e: FocusEvent<HTMLInputElement>) => void;
+    onEnterPressed?: (event: KeyboardEvent<HTMLInputElement>) => void;
+    onBlur?: (event: FocusEvent<HTMLInputElement>) => void;
     onClear?: () => void;
     size?: number;
 };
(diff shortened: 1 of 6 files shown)


src/elements/TextInput/TextInput.tsx
```diff
@@ -46,14 +46,14 @@ type TextInputBaseProps = {
...
     required?: boolean;
     disabled?: boolean;
     validation?: Validation;
-    onInput?: (value: string) => void;
-    onEnterPressed?: (value: string) => void;
-    onBlur?: (value: string) => void;
+    value?: string;
+    onChange?: (value: string) => void;
+    onEnterPressed?: (e: KeyboardEvent<HTMLInputElement>) => void;
+    onBlur?: (e: FocusEvent<HTMLInputElement>) => void;
```
oliverschwendener, line 56: Can we rename `e` to `event` or even `keyboardEvent` and `focusEvent`?

oliverschwendener (changes requested):
Beside my comment it looks great, thanks for taking care of this! 💪

oliverschwendener (approved):

PR #200, opened 2021-09-24
fix(Color Picker): Don't Allow Cursor to Leave Color Palette

deniaz:


src/components/ColorPicker/CustomColorPicker.tsx
```diff
@@ -16,7 +16,15 @@ import { Alpha, Saturation } from "react-color/lib/components/common";
 import { isValidHex, toState } from "react-color/lib/helpers/color";
 import { ColorFormat, ColorPickerProps } from "./ColorPicker";
 
-const ColorPointer: FC = () => <div className="tw-w-4 tw-h-4 tw-rounded-full tw-border-2 tw-border-white" />;
+const ColorPointer: FC<{ offsetX?: boolean; offsetY?: boolean }> = ({ offsetX = true, offsetY = true }) => (
```
NadiaPosch, line 19: Is the `offsetY` prop needed? At the moment there's no case where it's set to false so it could be removed.

PR #201, opened 2021-09-27
fix(Color Picker): Only Display HEX or RGBA Color

deniaz:


src/utilities/colors.ts
```diff
@@ -35,6 +35,7 @@ export const toColor = (current: Color, { name, ...diff }: DiffColor): Color =>
     const [r, g, b] = [rgb.r, rgb.g, rgb.b].map(getValidRgbColorValue);
     const rgba = { r, g, b, a: alpha };
     const hex = toState(rgba).hex;
+
```
triggertoo:
```suggestion
```

PR #206, opened 2021-09-27
feat(TextInput): Improve component

CelienB:
This PR adds a few improvements to the TextInput component.

- Add hover effect on the clear input button
- Add hover effect on the password show/hide button
- Add correct styles when the clear and password show/hide buttons are focused using a keyboard
- Remove the forbidden cursor on hover when input is disabled
- Refactoring to reduce complexity

### DIFF
--- a/src/elements/TextInput/TextInput.tsx
+++ b/src/elements/TextInput/TextInput.tsx
@@ -113,10 +113,8 @@ export const TextInput: FC<TextInputProps> = ({
     };
 
     const getInputType = () => {
-        if (type === TextInputType.Password && isObfuscated) {
-            return TextInputType.Password;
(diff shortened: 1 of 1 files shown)


src/elements/TextInput/TextInput.tsx
```diff
@@ -110,6 +112,15 @@ export const TextInput: FC<TextInputProps> = ({
         }
     };
 
+    const getInputType = () => {
+        if (type === TextInputType.Password && isObfuscated) {
```
NadiaPosch, line 116: You could simplify in the following way but it's up to you:
```
        if (type === TextInputType.Password) {
            return isObfuscated ? TextInputType.Password : TextInputType.Text;
        } 
        return type;
```

PR #231, opened 2021-09-30
feat(Tag): add component

CelienB:
This draft PR adds a `tag` component.


src/elements/Tag/Tag.tsx
```diff
@@ -0,0 +1,52 @@
...
+};
+
+export type TagProps = {
+    type: TagType;
+    label: string;
+    onClick?: () => void;
+};
+
+export const Tag: FC<TagProps> = ({ type, label, onClick }) => {
+    const isASelectedType = type === TagType.Selected || type === TagType.SelectedWithFocus;
```
NadiaPosch, line 33: I'd rename this to `isSelectedType`. It would be easier to read and not really affect its meaning, wdyt?

src/elements/Tag/Tag.tsx
```diff
@@ -0,0 +1,52 @@
...
+};
+
+export const Tag: FC<TagProps> = ({ type, label, onClick }) => {
+    const isSelectedType = type === TagType.Selected || type === TagType.SelectedWithFocus;
+
+    return (
+        <button
+            data-test-id="tag"
+            className={merge([
+                "tw-inline-flex tw-items-center tw-border tw-border-solid tw-rounded-full tw-text-xs tw-transition-colors tw-cursor-pointer tw-px-2.5 tw-py-1",
```
NadiaPosch, line 39: If the `onClick` prop is optional, I'd only set `tw-cursor-pointer` if `onClick` is defined, otherwise it's misleading

src/elements/Tag/Tag.stories.tsx
```diff
@@ -0,0 +1,47 @@
...
+export default {
+    title: "Elements/Tag",
+    component: Tag,
+    args: {
+        label: "Label",
+    },
+    argTypes: {
+        type: {
+            options: [...Object.keys(TagType)],
+            mappings: [...Object.values(TagType)],
```
NadiaPosch, line 16: `Object.keys` and `Object.values` already return arrays
```suggestion
            options: Object.keys(TagType),
            mappings: Object.values(TagType),
```

src/elements/Tag/Tag.tsx
```diff
@@ -0,0 +1,53 @@
...
+export const Tag: FC<TagProps> = ({ type, label, onClick }) => {
+    const isClickable = (type === TagType.Selected || type === TagType.SelectedWithFocus) && onClick;
+
+    return (
+        <button
+            data-test-id="tag"
+            className={merge([
+                "tw-inline-flex tw-items-center tw-border tw-border-solid tw-rounded-full tw-text-xs tw-transition-colors tw-cursor-default tw-px-2.5 tw-py-1",
+                tagStyles[type],
+                isClickable && "!tw-cursor-pointer",
```
NadiaPosch, line 41:
```suggestion
                isClickable && "tw-cursor-pointer",
```

src/elements/Tag/Tag.spec.tsx
```diff
@@ -0,0 +1,59 @@
...
+const TAG_LABEL = "My awesome tag";
+const TAG_REJECT_ICON_ATTRIBUTE = "svg[name=IconReject]";
+
+describe("Tag Component", () => {
+    it("should render a label", () => {
+        mount(<Tag type={TagType.Suggested} label={TAG_LABEL} />);
+
+        cy.get(TAG_ID).contains(TAG_LABEL);
+    });
+    it("should render a suggested tag", () => {
```
SamuelAlev, line 17: Maybe a bit of spacing for better readability?

```suggestion
    });

    it("should render a suggested tag", () => {
```

src/elements/Tag/Tag.stories.tsx
```diff
@@ -0,0 +1,47 @@
...
+        label: "Label",
+    },
+    argTypes: {
+        type: {
+            options: Object.keys(TagType),
+            mappings: Object.values(TagType),
+            defaultValue: TagType.Suggested,
+            control: { type: "select" },
+        },
+    },
```
SamuelAlev, line 20:
```suggestion
    argTypes: {
        type: {
            options: Object.keys(TagType),
            mappings: Object.values(TagType),
            defaultValue: TagType.Suggested,
            control: { type: "select" },
        },
        onClick: { action: "onClick" },
    },
```

And then you can delete all `onClick` from the `.args = { ... }`

NadiaPosch: The onClick is not set for all stories on purpose because the cursor style is different if onClick is undefined.

CelienB: It still works and shows the correct pointer since the type isn't `Selectable` or `SelectableWithFocus`, though.

NadiaPosch: It's just that onClick is now always set even though it's not intended for the `Selectable` and `SelectableWithFocus` types. But you can leave it as it is, if you prefer this version.

PR #247, opened 2021-10-05
feat: add ellipsis to input label

Walkeezy:
In regards to this: https://app.clickup.com/t/1kge6zu


Walkeezy:
@NadiaPosch: Now it should work on the single input label component too.

PR #254, opened 2021-10-06
fix: checkboxes states

CelienB:
This PR adds checkboxes state management back in the Checklist component stories.

### DIFF
--- a/src/compositions/Checklist/Checklist.stories.tsx
+++ b/src/compositions/Checklist/Checklist.stories.tsx
@@ -28,7 +28,7 @@ export const Checklist: Story<ChecklistProps> = (args) => {
         CheckboxState.Mixed,
     ]);
 
-    const updateCheckState = (index: number) => (isChecked: boolean) => {
+    const updateCheckState = (index: number, isChecked: boolean) => {
         checked[index] = isChecked ? CheckboxState.Checked : CheckboxState.Unchecked;
         setChecked([...checked]);
     };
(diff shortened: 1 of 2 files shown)


src/compositions/Checklist/Checklist.stories.tsx
```diff
@@ -54,12 +54,22 @@ export const Checklist: Story<ChecklistProps> = (args) => {
 };
 
 export const MultipleColumns: Story<ChecklistProps> = (args) => {
-    const [checked] = useState<CheckboxState[]>(new Array(8).fill(CheckboxState.Unchecked));
+    const [checked, setChecked] = useState<CheckboxState[]>(new Array(8).fill(CheckboxState.Unchecked));
+
+    const updateCheckState = (index: number) => (isChecked: boolean) => {
+        checked[index] = isChecked ? CheckboxState.Checked : CheckboxState.Unchecked;
+        setChecked([...checked]);
+    };
```
SamuelAlev, line 62:
```suggestion
    const updateCheckState = (index: number, isChecked: boolean) => {
        checked[index] = isChecked ? CheckboxState.Checked : CheckboxState.Unchecked;
        setChecked([...checked]);
    };
```
I don't think we ever used this way of to handle event (function returning function) in Arcade, I would then align with the rest and not introduce any other way now.
What do you think?

CelienB: Hey so this is just the way the function was written in the first story, I just borrowed it. Btw it won't work modifying the function the way you suggested.

NadiaPosch (approved):

SamuelAlev (approved):

PR #355, opened 2021-11-01
ci: Add Semantic Release with GitHub Actions Workflow

deniaz:


.releaserc
```diff
@@ -0,0 +1,16 @@
...
+            "name": "next",
+            "prerelease": true
+        }
+    ],
+    "plugins": [
+        "@semantic-release/commit-analyzer",
+        "@semantic-release/release-notes-generator",
+        "@semantic-release/npm",
+        "@semantic-release/github"
+    ],
```
SamuelAlev, line 14: Does it create a new release on github with the changelog?

.github/workflows/release.yml
```diff
@@ -0,0 +1,47 @@
+name: Release
+
+on:
+    workflow_dispatch:
+        branches:
+            - "main"
+    push:
+        branches:
+            - "next"
```
SamuelAlev, line 9: We may not want to have push on branch `main` (stable) and `next` (alpha) in the same workflow.
How do we distinguish them? Like `npm i @frontify/arcade@latest` should install a stable release and the `npm i @frontify/arcade@next` should install the alpha.

.github/workflows/release.yml
```diff
@@ -0,0 +1,47 @@
...
+
+        steps:
+            - name: Checkout default branch
+              uses: actions/checkout@v2
+
+            - name: Use Node.js 16 (npm repository)
+              uses: actions/setup-node@v2
+              with:
+                  node-version: "16"
+                  registry-url: "https://registry.npmjs.org"
```
SamuelAlev, line 27: We are missing the publish on github packages

.releaserc
deniaz: Exactly. :)

PR #369, opened 2021-11-07
style: use correct font size in heading

SamuelAlev:


src/components/FieldsetHeader/FieldsetHeader.tsx
```diff
@@ -10,15 +10,34 @@ import { merge } from "@utilities/merge";
...
+
 export enum FieldsetHeaderType {
-    Default = "Default",
-    Switch = "Switch",
-    Accordion = "Accordion",
-    AddRemove = "AddRemove",
+    Default = "default",
+    Switch = "switch",
+    Accordion = "accordion",
+    AddRemove = "add-remove",
```
Walkeezy, line 40: Isn't that kebab-case another new way of writing enum values? 😬

Walkeezy (approved):

PR #391, opened 2021-11-17
feat: Add TooltipIcon Component

marguslant0:
**Changes:**
- Extracting the logic of tooltip and the icon that triggers it out of `InputLabel` and into a separate `TooltipIcon` component
- Using the newly created `TooltipIcon` component inside `InputLabel` in place of the previous code
- Allowing the new component to set the size of the icon
- Updating tests

Storybook: http://localhost:6006/?path=/story/components-tooltipicon--with-default-icon-size

### DIFF
--- a/src/components/TooltipIcon/TooltipIcon.spec.tsx
+++ b/src/components/TooltipIcon/TooltipIcon.spec.tsx
@@ -1,16 +1,34 @@
 /* (c) Copyright Frontify Ltd., all rights reserved. */
 
-/* import { mount } from "@cypress/react";
 import React from "react";
+import { mount } from "@cypress/react";
 import { TooltipIcon } from "./TooltipIcon";
+import { IconSize } from "@foundation/Icon/IconSize";
 
-const TOOLTIP_COMPONENT = <Tooltip content="Cupcake ipsum dolor sit amet ice cream." />;
+const TOOLTIP_PROPS = {
+    content: "Lorem ipsum dolor sit amet.",
+};
 const TOOLTIP_ICON_ID = "[data-test-id=tooltip-icon]";
+const TOOLTIP_ICON_TRIGGER_ID = "[data-test-id=tooltip-icon-trigger]";
+const TOOLTIP_ID = "[data-test-id=tooltip]";
 
 describe("TooltipIcon Component", () => {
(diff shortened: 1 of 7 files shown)


src/components/TooltipIcon/TooltipIcon.spec.tsx
```diff
@@ -0,0 +1,34 @@
...
+    it("should render an icon with custom size", () => {
+        mount(<TooltipIcon tooltip={TOOLTIP_PROPS} iconSize={IconSize.Size20} />);
+
+        cy.get(TOOLTIP_ICON_ID).find(TOOLTIP_ICON_TRIGGER_ID).children().should("have.class", "tw-h-5");
+    });
+
+    it("should render a tooltip when the icon is hovered", () => {
+        mount(<TooltipIcon tooltip={TOOLTIP_PROPS} />);
+
+        cy.get(TOOLTIP_ICON_ID).realHover({ position: "top" });
```
jonasbuechel, line 31:
```suggestion
        cy.get(TOOLTIP_ICON_TRIGGER_ID).realHover({ position: "top" });
```

jonasbuechel (changes requested):
The change looks fine, just the test did not run so I added a suggestion 👌

src/components/TooltipIcon/TooltipIcon.spec.tsx
marguslant0: Thanks for the review, @jonasbuechel! 💪 

Using `TOOLTIP_ICON_TRIGGER_ID`, can you get the test to pass every time? I tried but got flaky results, and I'm trying now to understand what causes it 🤷

marguslant0: Looks like focus can randomly get lost in some cases. To ensure it won't, I added a click on the entire "body" before the hover action is performed. This seems to make it stable.

jonasbuechel: I think it also depends if you run the tests headless (`npm run test`) or with an open browser (`npm run test:browser`) 🤔

jonasbuechel (approved):
I was able to run the tests 3x with open and with closed browser and it's passing now 👍

PR #396, opened 2021-11-19
Feature - Add persisted icon to dropdown

msiric:
Allows the parent to pass the persistedIcon prop to the dropdown to display an icon at the beginning of the text input that doesn't get affected by the currently selected item.


src/components/Dropdown/Dropdown.tsx
```diff
@@ -30,6 +30,7 @@ export type DropdownProps = {
     disabled?: boolean;
     clearable?: boolean;
     ariaLabel?: string;
+    persistedIcon?: ReactElement;
```
SamuelAlev, line 33:
```suggestion
    persistedDecorator?: ReactElement;
```

SamuelAlev (approved):

SamuelAlev:
Nadia is on holidays, so I'll confirm it

PR #399, opened 2021-11-22
fix: add min-width to RTE when placeholder is displayed

Kenny806:


src/components/RichTextEditor/RichTextEditor.tsx
```diff
@@ -61,6 +62,8 @@ export const RichTextEditor: FC<RichTextEditorProps> = ({
 }) => {
     const [value, setValue] = useState<Descendant[]>(() => parseRawValue(initialValue));
     const debouncedValue = useDebounce(value, ON_SAVE_DELAY_IN_MS);
+    const wrapperRef = createRef<HTMLDivElement>();
+    const [placeholderElement, setPlaceholderElement] = useState<Element | null>();
```
triggertoo, line 66: you could set the wrapperStyle in the state directly, so you don't have to define a `placeholderElement`

triggertoo:
When removing the text, the placeholder gets displayed vertically for a short time:
https://user-images.githubusercontent.com/154760/142872770-7856ef90-9d8f-40b9-b6b3-9f18dab47414.mov

src/components/RichTextEditor/RichTextEditor.tsx
Kenny806: Done

Kenny806:
> When removing the text, the placeholder gets displayed vertically for a short time: https://user-images.githubusercontent.com/154760/142872770-7856ef90-9d8f-40b9-b6b3-9f18dab47414.mov

@triggertoo Yes. Because to be able to get the text size, we need the placeholder to be actually displayed first. Also, the placeholder element being displayed is the only indication that the editor content is empty. We wanted to check the `value` for this, but the object has such a complicated structure that @SamuelAlev suggested we just do it this way.

src/components/RichTextEditor/RichTextEditor.tsx
```diff
@@ -77,6 +80,15 @@ export const RichTextEditor: FC<RichTextEditorProps> = ({
         send("TEXT_UPDATED", { data: { value } });
     }, [debouncedValue]);
 
+    useEffect(() => {
+        const placeholderElement = wrapperRef.current?.querySelector('[data-slate-placeholder="true"]');
+        if (placeholderElement) {
+            setWrapperStyle({ minWidth: `${getTextWidth(placeholder, getCanvasFontSize(placeholderElement))}px` });
```
triggertoo, line 86:
```suggestion
            setWrapperStyle(placeholderElement ? { minWidth: `${getTextWidth(placeholder, getCanvasFontSize(placeholderElement))}px` } : undefined);
```

triggertoo:
> > When removing the text, the placeholder gets displayed vertically for a short time: https://user-images.githubusercontent.com/154760/142872770-7856ef90-9d8f-40b9-b6b3-9f18dab47414.mov
> 
> @triggertoo Yes. Because to be able to get the text size, we need the placeholder to be actually displayed first. Also, the placeholder element being displayed is the only indication that the editor content is empty. We wanted to check the `value` for this, but the object has such a complicated structure that @SamuelAlev suggested we just do it this way.

This would mean we have content jumping when it's loaded with content or being clear.
What about something like:
```ts
const placeholderRef = useRef<HTMLDivElement | null>(null);

    useEffect(() => {
        if (placeholderRef?.current?.style.position) {
            const isEmpty = ...;
            placeholderRef.current.style.position = matches(States.Readonly) && isEmpty ? "relative" : "absolute";
        }
    }, [matches, placeholderRef]);

    return (
        <div data-test-id="rich-text-editor" className="tw-min-w-[20px]">
            <Slate editor={editor} value={value} onChange={(value) => setValue(value)}>
                <Editable
                    placeholder={placeholder}
                    renderPlaceholder={({ children, attributes }) => {
                        attributes.style.position = "relative";
                        return (
                            <div {...attributes} ref={placeholderRef}>
                                <p>{children}</p>
                            </div>
                        );
                    }}
```

That way, you can always make sure that it's displayed with the right styles straight away. cc @SamuelAlev

triggertoo:
> > When removing the text, the placeholder gets displayed vertically for a short time: https://user-images.githubusercontent.com/154760/142872770-7856ef90-9d8f-40b9-b6b3-9f18dab47414.mov
> 
> @triggertoo Yes. Because to be able to get the text size, we need the placeholder to be actually displayed first. Also, the placeholder element being displayed is the only indication that the editor content is empty. We wanted to check the `value` for this, but the object has such a complicated structure that @SamuelAlev suggested we just do it this way.

To check if the Editor is empty, you can use [Editor.nodes](https://docs.slatejs.org/api/nodes/editor#editor.nodes-less-than-t-extends-node-greater-than-editor-editor-options-greater-than-generator-less) and check if there is a at least one text element with text.length > 0 with the match criteria

src/components/RichTextEditor/utils/setMinWidthIfEmpty.ts
```diff
@@ -0,0 +1,19 @@
...
+import { BaseEditor } from "slate";
+import { ReactEditor } from "slate-react";
+
+import { isEditorEmpty } from "@components/RichTextEditor/utils/isEditorEmpty";
+import { getCanvasFontSize, getTextWidth } from "@components/RichTextEditor/utils/getTextWidth";
+
+export const setMinWidthIfEmpty = (
+    editor: BaseEditor & ReactEditor,
+    placeholder: string,
+    setWrapperStyle: React.Dispatch<React.SetStateAction<React.CSSProperties | undefined>>,
```
SamuelAlev, line 11: Don't really like idea of passing this one, as the function complexity is quite low, could you move it back to RTE.tsx?

src/components/RichTextEditor/utils/getTextWidth.ts
```diff
@@ -0,0 +1,23 @@
+export function getTextWidth(text: string, font: string): number | null {
+    const canvas = document.createElement("canvas");
+    const context = canvas.getContext("2d");
+    if (context) {
+        context.font = font;
+        const metrics = context.measureText(text);
+
+        return Math.ceil(metrics.width * 1.05); // add a 5% safety margin
```
SamuelAlev, line 8: Could you put an explanation of why we need it (need to load font first, this is based on first fallback, ...)

src/components/RichTextEditor/utils/setMinWidthIfEmpty.ts
```diff
@@ -0,0 +1,19 @@
+import React from "react";
```
SamuelAlev, line 1: Missing copyright on files

src/components/RichTextEditor/utils/getTextWidth.ts
Kenny806: Added

src/components/RichTextEditor/utils/setMinWidthIfEmpty.ts
Kenny806: Returning `CSSProperties | undefined` now and setting the state variable inside the component

src/components/RichTextEditor/utils/setMinWidthIfEmpty.ts
Kenny806: Added

src/components/RichTextEditor/RichTextEditor.tsx
```diff
@@ -77,6 +84,11 @@ export const RichTextEditor: FC<RichTextEditorProps> = ({
         send("TEXT_UPDATED", { data: { value } });
     }, [debouncedValue]);
 
+    const onValueChanged = (value: Descendant[]): void => {
```
triggertoo, line 87: could you add a useCallback here?

src/components/RichTextEditor/utils/getMinWidthIfEmpty.ts
```diff
@@ -0,0 +1,19 @@
...
+
+import { isEditorEmpty } from "@components/RichTextEditor/utils/isEditorEmpty";
+import { getCanvasFontSize, getTextWidth } from "@components/RichTextEditor/utils/getTextWidth";
+
+export const getMinWidthIfEmpty = (
+    editor: BaseEditor & ReactEditor,
+    placeholder: string,
+    wrapperEl: HTMLDivElement | null,
+): CSSProperties | undefined => {
+    if (isEditorEmpty(editor) && wrapperEl) {
```
triggertoo, line 15:
```suggestion
    if (wrapperEl && isEditorEmpty(editor)) {
```
for performance reasons

SamuelAlev (approved):

src/components/RichTextEditor/RichTextEditor.tsx
```diff
@@ -77,6 +84,11 @@ export const RichTextEditor: FC<RichTextEditorProps> = ({
...
+        setValue(value);
+        setWrapperStyle(getMinWidthIfEmpty(editor, placeholder, wrapperRef.current));
```
triggertoo, line 89: could you add a check if there is placelholder to only handle it in this case?

Kenny806: Added the check inside `getMinWidthIfEmpty`

src/components/RichTextEditor/RichTextEditor.tsx
Kenny806: Done

PR #401, opened 2021-11-23
Fix unclickable areas inside dropdown

msiric:
This fix eliminates gaps that don't handle click events around the dropdown text and the caret icon on the right-hand side.

### DIFF
--- a/src/components/Dropdown/Dropdown.tsx
+++ b/src/components/Dropdown/Dropdown.tsx
@@ -69,7 +69,7 @@ export const Dropdown: FC<DropdownProps> = ({
     );
 
     return (
-        <div className="tw-relative tw-w-full tw-font-sans tw-text-s ">
+        <div className="tw-relative tw-w-full tw-font-sans tw-text-s">
             <Trigger
                 disabled={disabled}
                 buttonProps={buttonProps}
@@ -96,7 +96,7 @@ export const Dropdown: FC<DropdownProps> = ({
                     data-test-id="dropdown-trigger"
                     className={merge([
                         "tw-overflow-hidden tw-flex-auto tw-h-full tw-rounded tw-text-left tw-outline-none",
-                        size === DropdownSize.Small ? "tw-py-1 tw-px-3" : "tw-p-5 tw-min-h-[60px]",
+                        size === DropdownSize.Small ? "tw-py-2 tw-px-3 tw-min-h-[34px]" : "tw-p-5 tw-min-h-[60px]",
                         !activeItem && "tw-text-black-60",
                         disabled && "tw-text-black-40",
                     ])}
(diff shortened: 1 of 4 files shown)


src/components/MenuItem/MenuItemContent.tsx
```diff
@@ -32,19 +32,24 @@ export const MenuItemContent: FC<MenuItemContentProps> = ({
     <div
         {...ariaProps}
         data-test-id="menu-item-content"
-        className={merge([
-            "tw-flex tw-box-border tw-items-center tw-font-sans tw-text-s",
-            size === MenuItemContentSize.Large ? "tw-gap-3" : "tw-gap-2",
-        ])}
+        className={merge(["tw-flex tw-box-border tw-items-center tw-font-sans tw-text-s"])}
```
SamuelAlev, line 35:
```suggestion
        className="tw-flex tw-box-border tw-items-center tw-font-sans tw-text-s"
```

SamuelAlev:
Could you also rebase on `main`/ merge `main` as there is a fix for the storybook pr preview?

src/components/Dropdown/Dropdown.tsx
```diff
@@ -96,7 +96,7 @@ export const Dropdown: FC<DropdownProps> = ({
                     data-test-id="dropdown-trigger"
                     className={merge([
                         "tw-overflow-hidden tw-flex-auto tw-h-full tw-rounded tw-text-left tw-outline-none",
-                        size === DropdownSize.Small ? "tw-py-1 tw-px-3" : "tw-p-5 tw-min-h-[60px]",
+                        size === DropdownSize.Small ? "tw-py-2 tw-px-3 tw-min-h-[34px]" : "tw-p-5 tw-min-h-[60px]",
```
SamuelAlev, line 99: We don't have the cursor in the trigger of the color picker flyout https://deploy-preview-401--arcade-components.netlify.app/?path=/story/components-color-picker-flyout--flyout

msiric: Could you elaborate? I tried it out and couldn't replicate the issue.

SamuelAlev: https://deploy-preview-401--arcade-components.netlify.app/?path=/story/components-color-picker-flyout--flyout

On the top near the border, the cursor is not a pointer but click seems to work (Firefox)

msiric: Ah I see now. You're correct and it's been fixed. 
But I'm not sure if this is the best way to do it since the padding is coming from the wrapping `button` element.
This means that whenever the `MenuItemContent` component is used, we will expect the parent to have the appropriate padding styling.

SamuelAlev (approved):

PR #405, opened 2021-11-24
feat: Add card component

marguslant0:
**Changes:**
- Add Card component

In the future, this component should also take `interactive` prop to handle `onClick` and hover state, but omitting them for now, since none of it is currently needed for the Brand Listening use case.


src/components/Card/Card.tsx
```diff
@@ -0,0 +1,15 @@
...
+
+export type CardProps = {
+    children?: ReactChild;
+};
+
+export const Card: FC<CardProps> = ({ children }) => {
+    return (
+        <div data-test-id="card" className="tw-w-full tw-bg-white tw-border tw-border-black-10 tw-rounded">
+            {children}
+        </div>
```
SamuelAlev, line 13: Could you ask Jamie about making a component?
We can't accept component that are not inside of the Figma Arcade file as components need to respect the Figma specs (aka the props used in Figma should be used in code)

marguslant0: Hi @SamuelAlev 👋 

Thanks for the info!

Using [this task](https://app.clickup.com/t/1u470ae) as the input, and our current Brand Listening use case on [this prototype](https://www.figma.com/proto/F571BXcImisdo3ETFC3TSV/Social-Listening?node-id=88%3A104198&starting-point-node-id=286%3A192058), we have currently agreed with Jonas that this component won't have a header and divider.

But I will check with Jamie regarding getting it to the Figma Arcade file, thanks!

marguslant0: Btw, Jamie has the Card designs in the WIP file here:
https://www.figma.com/file/pevuvey1joCsopTRr9f4OE/Arcade-Components?node-id=11443%3A363037

src/components/Card/Card.tsx
```diff
@@ -0,0 +1,15 @@
...
+
+import React, { FC, ReactChild } from "react";
+
+export type CardProps = {
+    children?: ReactChild;
+};
+
+export const Card: FC<CardProps> = ({ children }) => {
+    return (
+        <div data-test-id="card" className="tw-w-full tw-bg-white tw-border tw-border-black-10 tw-rounded">
```
jonasbuechel, line 11: @thisguyscode should the card also include the 12px padding or does that belong to the content?

marguslant0: Jumping in here with regards to the reason why I currently left it out - as some of our use cases are going to have a Divider that stretches across the entire card width, we'd either need to start overriding the original padding or make it optional. But I'd be happy to change it however we decide :)

thisguyscode: @jonasbuechel @marguslant0 

0 padding should definitely be an option. If we also want padded options, we can provide it as a prop. I'd suggest that 0 padding remains the default.

src/components/Card/Card.tsx
```diff
@@ -0,0 +1,15 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+import React, { FC, ReactChild } from "react";
+
+export type CardProps = {
+    children?: ReactChild;
+};
```
SamuelAlev, line 7:
```suggestion
// import { PropsWithChildren } from 'react';
export type CardProps = PropsWithChildren<{}>
```

marguslant0: Thanks, @SamuelAlev! This could definitely be done when there are more props in the future. Currently, eslint doesn't like it empty.

src/components/Card/Card.tsx
marguslant0: Marking it as resolved, then, if you agree, @jonasbuechel?

SamuelAlev (approved):

jonasbuechel (approved):

PR #410, opened 2021-11-25
fix: FormControl children vertical spacing

CelienB:
This PR fixes the spacing issue for FormControl children components when direction is set to vertical.
Ticket: https://app.clickup.com/t/yyc0fd


src/components/FormControl/FormControl.tsx
```diff
@@ -70,6 +70,8 @@ export const FormControl: FC<FormControlProps> = ({
 }) => {
     const isHelperBefore = helper?.position === HelperPosition.Before;
 
+    const directionVerticalClasses = "tw-w-full tw-grid tw-gap-5";
```
SamuelAlev, line 73: You can inline them

PR #428, opened 2021-11-30
fix: input focus in accordion

SamuelAlev:
Fixes https://app.clickup.com/t/1vc80g2


SamCreasey (approved):
It works! So much nicer that it's just a prop to turn off 😄

PR #431, opened 2021-12-01
fix(RichTextEditor): Update content when value property is updated

triggertoo:


Walkeezy (approved):

src/components/RichTextEditor/hooks/useEditorValueUpdates.ts
```diff
@@ -0,0 +1,26 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+import { useEffect, useRef } from "react";
+import { Editor, Transforms } from "slate";
+import { clearEditor } from "../utils/editor/clear";
+import { parseRawValue } from "../utils/editor/parseRawContent";
+
+/**
+ * Updates the editor value when the initival value changes
```
SamuelAlev, line 9: Suggestions is broken on GH for now 🤷 
There is a typo "when the initival value changes"

src/components/RichTextEditor/RichTextEditor.tsx
```diff
@@ -71,6 +72,7 @@ export const RichTextEditor: FC<RichTextEditorProps> = ({
     const [{ matches, children }, send] = useMachine(() =>
         editorMachine.withContext({ locked: readonly, onTextChange, onBlur }),
     );
+    useEditorValueUpdates(editor, initialValue);
```
SamuelAlev, line 75: Is it updating the xstate's context? Seems like the `editor` object is not touched here

triggertoo: `editor` and `value` are not part of the xstate context (they are sent with the events)

src/components/RichTextEditor/hooks/useEditorValueUpdates.ts
triggertoo: https://github.com/Frontify/arcade/pull/435

PR #479, opened 2021-12-13
feat: add flyout footer with custom buttons to flyout component

juergscherrer:


src/components/Flyout/Flyout.spec.tsx
```diff
@@ -23,8 +29,23 @@ const Component: FC<Pick<FlyoutProps, "onClick" | "onClose" | "badges">> = ({ on
...
+                <FlyoutFooter
+                    primaryButton={
+                        <Button key="add" style={ButtonStyle.Primary} onClick={cy.stub().as("onClickStub")}>
+                            Add
+                        </Button>
+                    }
+                    secondaryButton={
+                        <Button key="cancel" style={ButtonStyle.Secondary} onClick={cy.stub().as("onCloseStub")}>
+                            Cancel
+                        </Button>
```
jonasbuechel, line 45: Naming primary/secondary is misleading because it's already in use for buttonStyle. 
I suggest to use [TS tuples](https://www.typescriptlang.org/docs/handbook/basic-types.html#tuple) like we do it in [tooltip](https://github.com/Frontify/arcade/blob/fb75b2a37944e3e9ea1455c2df546340bc5c4cfd/src/components/Tooltip/Tooltip.tsx#L39).

jonasbuechel (changes requested):

thisguyscode:
@juergscherrer Seems to be a padding issue when looking at the storybook build

juergscherrer:
@thisguyscode Which padding issue do you mean? Could you please create a screenshot? Thx

thisguyscode:
Oh it's not padding, it's related to the scroll overflow. I asked Samuel in
Slack because it's effecting multiple components.

jonasbuechel (approved):

PR #489, opened 2021-12-15
fix(accordion): prevent navigation from child components

SamCreasey:


src/components/Accordion/Accordion.tsx
```diff
@@ -126,18 +151,33 @@ export const Accordion: FC<AccordionProps> = (props) => {
...
-        accordionProps: { onMouseDown, ...accordionProps },
+        accordionProps: { onMouseDown, onKeyDown, ...accordionProps },
         // @react-aria enable by default typeahead which result in an event fired up on keypress and select the section
         // eslint-disable-next-line @typescript-eslint/ban-ts-comment
         //@ts-ignore
     } = useAccordion({ ...ariaProps, disallowTypeAhead: true }, state, ref);
 
+    const accordionItemCheck = (event: KeyboardEvent<HTMLDivElement>) => {
+        const validItemIds = [ACCORDION_ID, ACCORDION_ITEM_ID];
+        const id = (event.target as HTMLButtonElement).dataset?.testId;
```
triggertoo, line 165: `data-test-id` are only available in development/testing environment and removed in the production build, so we shouldn't rely on it.

triggertoo: you could probably check if the actual item is focused in the selection manager from the item state

SamCreasey: Fixed :)

src/components/Accordion/Accordion.tsx
```diff
@@ -10,10 +10,35 @@ import { Node } from "@react-types/shared";
...
+const ACCORDION_CONTROL_KEYS = [
+    "ArrowDown",
+    "ArrowUp",
+    "ArrowRight",
+    "ArrowLeft",
+    "Home",
+    "End",
+    "PageDown",
+    "PageUp",
+    "a",
```
SamuelAlev, line 38: For what purpose?

src/components/Accordion/Accordion.tsx
```diff
@@ -126,18 +151,33 @@ export const Accordion: FC<AccordionProps> = (props) => {
     const {
         // @react-aria prevents default action for onMouseDown as implemented here: https://github.com/adobe/react-spectrum/blob/e14523fedd93ac1a4ede355aed70988af572ae74/packages/%40react-aria/selection/src/useSelectableCollection.ts#L370
         // This makes it impossible to edit or select text in input fields inside the accordion
+        // onKeydown automatically navigates away from child elements, including dropdown menus etc.
+        // Focus should only be passed if the accordian item itself is focused
```
SamuelAlev, line 155:
```suggestion
        // Focus should only be passed if the accordion item itself is focused
```

src/components/Accordion/Accordion.tsx
```diff
@@ -126,18 +151,33 @@ export const Accordion: FC<AccordionProps> = (props) => {
...
         // eslint-disable-next-line @typescript-eslint/no-unused-vars
-        accordionProps: { onMouseDown, ...accordionProps },
+        accordionProps: { onMouseDown, onKeyDown, ...accordionProps },
         // @react-aria enable by default typeahead which result in an event fired up on keypress and select the section
         // eslint-disable-next-line @typescript-eslint/ban-ts-comment
         //@ts-ignore
     } = useAccordion({ ...ariaProps, disallowTypeAhead: true }, state, ref);
 
+    const accordionItemCheck = (event: KeyboardEvent<HTMLDivElement>) => {
```
SamuelAlev, line 163:
```suggestion
    const accordionKeyDown = (event: KeyboardEvent<HTMLDivElement>) => {
```

src/components/Accordion/Accordion.tsx
SamCreasey: useAccordion uses a keyboard listener which prevents default of these keys, to keep this consistent with the hook you have to manually call it. The main reason is because pressing some of these keys will activate the window scroller, like the up and down arrow key which is not the expected behaviour from the useAccordion hook

SamuelAlev: Ok, but why the letter `a`? I can understand all the others ones

SamCreasey: Ah that was used for their selectAll function, but I will remove that it shouldn't change anything

src/components/Accordion/Accordion.tsx
SamCreasey: fixed

src/components/Accordion/Accordion.tsx
SamCreasey: fixed

src/components/Accordion/Accordion.tsx
```diff
@@ -126,18 +150,33 @@ export const Accordion: FC<AccordionProps> = (props) => {
...
         // eslint-disable-next-line @typescript-eslint/ban-ts-comment
         //@ts-ignore
     } = useAccordion({ ...ariaProps, disallowTypeAhead: true }, state, ref);
 
+    const accordionKeyDown = (event: KeyboardEvent<HTMLDivElement>) => {
+        const key = (event.target as HTMLButtonElement).dataset.key;
+        const isFocused = state.selectionManager.focusedKey === key;
+        if (key && isFocused) {
+            onKeyDown && onKeyDown(event);
+        } else if (ACCORDION_CONTROL_KEYS.includes(event.key)) {
```
triggertoo, line 167: could you try to use https://react-spectrum.adobe.com/react-aria/useFocusWithin.html instead, so you don't need to deal with control key checks?

SamCreasey: fixed as discussed :)

triggertoo (approved):

PR #490, opened 2021-12-16
fix: revert useEditorValueUpdates in RichTextEditor and add clear property

NadiaPosch:


src/components/RichTextEditor/RichTextEditor.tsx
```diff
@@ -28,6 +28,7 @@ export type RichTextEditorProps = {
     onTextChange?: (value: string) => void;
     onBlur?: (value: string) => void;
     readonly?: boolean;
+    clear?: () => void;
```
triggertoo, line 31: i guess it's enough to just use it as a boolean

src/components/RichTextEditor/RichTextEditor.tsx
```diff
@@ -72,9 +74,12 @@ export const RichTextEditor: FC<RichTextEditorProps> = ({
     const [{ matches, children }, send] = useMachine(() =>
         editorMachine.withContext({ locked: readonly, onTextChange, onBlur }),
     );
-    useEditorValueUpdates(editor, initialValue);
 
     useEffect(() => {
+        if (clear !== undefined) {
```
triggertoo, line 79: you can put it into its own useEffect, since it should be executed every time the property changes:
```
useEffect(() => {
    clear && clearEditor(editor);
, [clear]}
```

(with clear as boolean)

NadiaPosch: Yes, makes sense 👍  I wasn't sure if you may want to execute something in the block settings before clearing

PR #495, opened 2021-12-17
style: set overflow to auto instead of scroll

SamuelAlev:


juergscherrer (approved):
Looks good now. Theoretically, it should work without this class because `overflow: auto` is the default case in CSS right?

PR #496, opened 2021-12-17
feat: add `hugWidth` property to the button component which displays it in full width.

juergscherrer:


src/components/Button/Button.tsx
```diff
@@ -84,6 +84,7 @@ export type ButtonProps = {
     icon?: ReactElement;
     children?: string;
     onClick?: (event?: MouseEvent<HTMLButtonElement>) => void;
+    hugWidth?: boolean;
```
SamuelAlev, line 87: We use `hugContent` in the custom blocks, I would sync with designers first

src/components/Button/Button.tsx
```diff
@@ -95,6 +96,7 @@ export const Button: FC<ButtonProps> = ({
     icon,
     children,
     onClick,
+    hugWidth = false,
```
SamuelAlev, line 99: It should be `true` by default as we don't want button to be full width by default

src/components/Button/Button.tsx
```diff
@@ -124,6 +126,7 @@ export const Button: FC<ButtonProps> = ({
                           ]
                         : [isFocusVisible && FOCUS_STYLE, styles[getButtonTheme()][style]],
                 ),
+                hugWidth && "tw-w-full",
```
SamuelAlev, line 129:
```suggestion
                !hugWidth && "tw-w-full",
```
"Hug" is the current behavior (inline & fit to content)

juergscherrer: Ah sorry, then I misunderstood. Thank you for clarifying.

SamuelAlev (approved):

PR #497, opened 2021-12-17
feat(button): added scaleIcon prop

SamCreasey:


CelienB (approved):

SamuelAlev:
Could you show that change in a story and add it as an `args`?

PR #500, opened 2021-12-20
fix(icons): update caret icons

SamCreasey:


SamuelAlev (approved):

thisguyscode:
@SamCreasey @SamuelAlev Are you sure about this one? The viewbox on all icons should be 24x24. Now the up/down carets are a different size than the right/left carets.

SamuelAlev:
It is still correct (1:1 ratio), Sam just changed the svg not the wrapper (wrapper still have 16x16, 24x24, ...).
Capture d’écran 2021-12-21 à 13 00 21

thisguyscode:
The size of the icon is still correct, sure. But the path has changed in relation to it's container. Compare it to the right/left arrows and you'll see what I mean.

The padding in Figma is intended, in order to maintain similar visual weight between icons.

thisguyscode:
@SamuelAlev The tree is a good example of the issue: https://deploy-preview-500--arcade-components.netlify.app/?path=/story/components-tree--tree

SamuelAlev:
>Compare it to the right/left arrows and you'll see what I mean.

It still need to be fixed (as most of the icons)

PR #505, opened 2021-12-21
fix(orderable-list): prevent drag crossover

SamCreasey:
This PR addresses the issue item 10 left on the checklist block ticket concerning accessibility, as well as a partial fix to item 3: https://app.clickup.com/t/ndbnm8


src/components/OrderableList/FocusController.tsx
```diff
@@ -20,24 +20,29 @@ export const FocusController: FC<FocusControllerProps> = ({ children, width = Fo
 
         switch (event.code) {
             case "Space":
+                event.stopPropagation();
                 if (!isChildsEvent) {
+                    event.preventDefault();
                     childRef?.current?.focus();
                 }
                 break;
             case "Enter":
```
NadiaPosch, line 29: You could combine this with the `"Space"` case if the actions are the same

SamCreasey: Updated: the Enter case required additional else logic :)

NadiaPosch (approved):

SamuelAlev (approved):

PR #509, opened 2021-12-21
fix: Color Picker Flyout

Kenny806:


src/components/ColorPicker/BrandColorPicker.tsx
```diff
@@ -56,6 +56,7 @@ export const BrandColorPicker: FC<Props> = ({ palettes: defaultPalettes = [], cu
                         decorator={<IconSearch />}
                         placeholder="Search"
                         onChange={(value) => setQuery(value)}
+                        clearable
```
Kenny806, line 59: This fixes `Search input field should have a X when it is filled to quickly remove search term`

Kenny806:
This PR fixed two separate problems, so it was split into two PRs which were merged individually

PR #526, opened 2022-01-04
fix: allow React node as a children

SamuelAlev:


juergscherrer (approved):
Looks good and is exactly what I need. Thx @SamuelAlev

PR #531, opened 2022-01-07
fix(slider): slider focus ring inside overlays

SamCreasey:
This fixes the issue that causes a "focus visible" ring to appear when the slider label is pressed for the second time inside an overlay.
Clickup link: https://app.clickup.com/t/1p0hg4a

### DIFF
--- a/src/components/Slider/Slider.tsx
+++ b/src/components/Slider/Slider.tsx
@@ -1,6 +1,4 @@
 /* (c) Copyright Frontify Ltd., all rights reserved. */
-/* eslint-disable jsx-a11y/click-events-have-key-events */
-/* eslint-disable jsx-a11y/no-noninteractive-element-interactions */
 
 import { IconProps } from "@foundation/Icon/IconProps";
 import { useMemoizedId } from "@hooks/useMemoizedId";
--- a/src/components/Slider/Slider.tsx
+++ b/src/components/Slider/Slider.tsx
@@ -12,7 +12,7 @@ import { RadioGroupState, useRadioGroupState } from "@react-stately/radio";
 import { FOCUS_STYLE } from "@utilities/focusStyle";
 import { merge } from "@utilities/merge";
 import { AnimateSharedLayout, motion } from "framer-motion";
-import React, { FC, MouseEvent, ReactElement, useMemo, useRef } from "react";
+import React, { FC, ReactElement, useMemo, useRef } from "react";
 
 export type IconItem = {
     id: string;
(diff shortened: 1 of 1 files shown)


src/components/Slider/Slider.tsx
```diff
@@ -56,6 +58,11 @@ const SliderItem = (props: SliderItemProps) => {
     );
     const { isFocusVisible, focusProps } = useFocusRing();
 
+    const handleLabelClick = (event: MouseEvent<HTMLLabelElement>) => {
+        event.preventDefault();
```
hochreutenerl, line 62: From my point of view, I don't think the solution for this issue is to prevent the event, but rather check that `isFocusVisible` has the correct value.

SamCreasey: Label has been converted into a div so that default event does not need to be prevented

src/components/Slider/Slider.tsx
```diff
@@ -1,8 +1,11 @@
 /* (c) Copyright Frontify Ltd., all rights reserved. */
+/* eslint-disable jsx-a11y/click-events-have-key-events */
+/* eslint-disable jsx-a11y/no-noninteractive-element-interactions */
```
hochreutenerl, line 3: Do these two lines need to be there? I don't get any errors when running `npm run lint`

src/components/Slider/Slider.tsx
```diff
@@ -76,8 +87,11 @@ const SliderItem = (props: SliderItemProps) => {
                     aria-hidden="true"
                 />
             )}
-            <label
-                htmlFor={isActive ? id : undefined}
+            <div
```
hochreutenerl, line 90: Sadly that's also neither the perfect way, but if we don't find anything better, probably it will probably cause fewer problems than the event prevention might.
If I see it right it's not possible to have a `<label>` with the react-aria behavior of the div, is it?

src/components/Slider/Slider.tsx
SamCreasey: Ah yes, after reinstalling node modules the errors disappeared, will remove this

src/components/Slider/Slider.tsx
SamCreasey: Yes that's exactly right. Since it is the default behavior of the label itself that is not playing along nicely with react-aria. Also since the div is not a focusable element but the input is, screen-readers read the aria-label out and ignore the div,  so there's no issue there for now

hochreutenerl (approved):
Works nice. Thanks for taking the extra mile with this.

Let's also subscribe to https://github.com/adobe/react-spectrum/issues/2380 so we get an update if anything would change.

PR #538, opened 2022-01-12
style: hide overflow long filename

SamuelAlev:


src/components/AssetInput/AssetInput.tsx
```diff
@@ -147,44 +133,43 @@ const SelectedAsset: FC<AssetProps> = ({ asset, size, actions }) => {
...
             >
                 <AssetThumbnail asset={asset} size={size} isActive={isOpen || isFocusVisible} />
-                <span
+                <div
                     className={merge([
-                        "tw-flex tw-flex-1",
-                        size === AssetInputSize.Large
-                            ? "tw-h-14 tw-w-full tw-border-t tw-border-black-100 tw-border-opacity-10"
-                            : "tw-h-full",
+                        "tw-min-w-0 tw-flex tw-flex-auto tw-border-black-100 tw-border-opacity-25",
```
SamCreasey, line 156:
```suggestion
                        "tw-min-w-0 tw-flex tw-flex-auto tw-border-black-100 tw-border-opacity-25 tw-max-w-full tw-self-stretch",
```

Needs this max-w-full otherwise it breaks on vertical stacking mode. Needs self-stretch for border to stretch the whole width in vertical too

SamCreasey (changes requested):

src/components/AssetInput/AssetInput.tsx
SamuelAlev: What do you mean by vertical stacking mode?

SamCreasey: AH, I mean when the size is switched to the large

SamCreasey (approved):
👍

PR #541, opened 2022-01-17
fix(flyout): Prevent closing on over-scroll

SamCreasey:
Clickup Ticket: https://app.clickup.com/t/22bbehk

This fix prevents the color picker flyout from triggering a close when component is over-scrolled by preventing the scroll propagation


src/components/Flyout/Flyout.tsx
```diff
@@ -164,13 +167,16 @@ export const Flyout: FC<FlyoutProps> = ({
...
     });
+
     const { buttonProps } = useButton({ onPress: () => toggle() }, triggerRef);
     useEffect(() => {
         const revert = watchModals();
 
         return () => revert();
     }, []);
 
+    useContainScroll(overlayRef, { isDisabled: !isOpen });
```
NadiaPosch, line 178: Is there a reason you use an object instead of just a boolean for `isDisabled`?

SamCreasey: No particular reason other than I try to be consistent with the react-aria hooks, where supplying properties to the hooks is always done as an object for clarity, but it's not essential

src/components/Flyout/useContainScroll.ts
```diff
@@ -0,0 +1,38 @@
...
+            const deltaY = event.deltaY;
+            const up = deltaY < 0;
+            const scrollTop = overlay.scrollTop;
+            const scrollHeight = overlay.scrollHeight;
+            const height = overlay.clientHeight;
+
+            if (!up && deltaY > scrollHeight - height - scrollTop) {
+                // Scrolling down, but this will take us past the bottom.
+                overlay.scrollTop = scrollHeight;
+                return prevent();
```
NadiaPosch, line 21: To me it would make more sense to call the prevent method and then return false:
```suggestion
                prevent();
                return false;
```

src/components/Flyout/useContainScroll.ts
```diff
@@ -0,0 +1,38 @@
+import { MutableRefObject, useEffect } from "react";
+
+export const useContainScroll = (overlayRef: MutableRefObject<HTMLDivElement | null>, { isDisabled = false }) => {
+    useEffect(() => {
+        const scrollHandler = (event: WheelEvent) => {
+            const prevent = function () {
```
NadiaPosch, line 6:
```suggestion
            const prevent = () => {
```

src/components/Flyout/useContainScroll.ts
```diff
@@ -0,0 +1,38 @@
...
+                event.stopPropagation();
+                event.preventDefault();
+                return false;
+            };
+            const overlay = overlayRef.current as HTMLDivElement;
+            const deltaY = event.deltaY;
+            const up = deltaY < 0;
+            const scrollTop = overlay.scrollTop;
```
NadiaPosch, line 14: Not sure if this could lead to an error if `overlay` is null
```suggestion
            const scrollTop = overlay?.scrollTop;
```

src/components/Flyout/useContainScroll.ts
```diff
@@ -0,0 +1,38 @@
...
+                overlay.scrollTop = scrollHeight;
+                return prevent();
+            } else if (up && -deltaY > scrollTop) {
+                // Scrolling up, but this will take us past the top.
+                overlay.scrollTop = 0;
+                return prevent();
+            }
+        };
+        const eventProps = { passive: false, capture: true };
+        if (!isDisabled && overlayRef.current !== null) {
```
NadiaPosch, line 29: You could replace `overlayRef.current` with `overlay`
```suggestion
        if (!isDisabled && overlay !== null) {
```

src/components/Flyout/Flyout.tsx
NadiaPosch: Fair point, also you immediately know what the argument is used for 👍

src/components/Flyout/useContainScroll.ts
SamCreasey: fixed with if condition

src/components/Flyout/useContainScroll.ts
SamCreasey: fixed

src/components/Flyout/useContainScroll.ts
SamCreasey: fixed

src/components/Flyout/useContainScroll.ts
```diff
@@ -0,0 +1,46 @@
...
+                event.stopPropagation();
+                event.preventDefault();
+            };
+
+            if (overlay !== null) {
+                const deltaY = event.deltaY;
+                const up = deltaY < 0;
+                const scrollTop = overlay.scrollTop;
+                const scrollHeight = overlay.scrollHeight;
+                const height = overlay.clientHeight;
```
NadiaPosch, line 18: You could destruct like so:
```suggestion
                const { scrollTop, scrollHeight, clientHeight } = overlay;
```

src/components/Flyout/useContainScroll.ts
```diff
@@ -0,0 +1,46 @@
...
+                    overlay.scrollTop = 0;
+                    prevent();
+                    return false;
+                }
+            }
+        };
+
+        const eventProps = { passive: false, capture: true };
+
+        if (!isDisabled && overlay !== null) {
```
NadiaPosch, line 36: You could move the eventProps to the top of the method and move this inside the `if` block from above so you don't have to do the overlay check again

src/components/Flyout/useContainScroll.ts
SamCreasey: done!

src/components/Flyout/useContainScroll.ts
SamCreasey: The other `if` block is inside of the event handler function, so I can't really do that here. An alternative would be to instantiate the `scrollEvent` variable outside of an if, and then assign it the function from within the `if` statement, so that the return function has a reference to it and i don't need to run the overlay check again. 
```
let scrollEvent;
if(overlay !== null){
  scrollEvent = (event) => ...
  window.addEventListener('wheel', scrollEvent)
}
```
I'm not sure if this is any better than what is already there, but wdyt?

NadiaPosch: You're right, I missed that fact! I'm happy with the code, nice work 💪🏽

NadiaPosch (approved):

PR #542, opened 2022-01-17
Simplify color data structure

hochreutenerl:
These are just some tryouts with a simplified color data structure. Note that the code is in no way meant to be merged directly as it is incomplete and not tested.


hochreutenerl:
No more need for this

PR #543, opened 2022-01-18
fix(color-picker): change scroll position

SamCreasey:
Clickup Ticket: https://app.clickup.com/t/nq8zf6 (Checklist item 6 & 7)

Fixes:
- Add toggleable prop to flyout component to determine if overlay is a scrollable container, or if the scrolling should be managed inside the child components
- updated useContainScroll hook to prevent bubbling scroll behaviour if any scrollable children are present inside the overlay
- Added wrapping to brand color palattes to prevent overflow scroll appearing horizontally

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6,7 +6,7 @@
     "packages": {
         "": {
             "name": "@frontify/arcade",
-            "version": "0.16.0",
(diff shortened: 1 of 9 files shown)


Walkeezy (approved):
Nice 🤘

PR #550, opened 2022-01-20
fix(link-chooser): design review corrections

SamCreasey:
Fixes for checklist items 4,5,8,9,10,13 on ticket https://app.clickup.com/t/1vht3xx
- refactored dropdown auto-resize hook and added to link chooser
- Fixed padding and text alignment on no-results displays
- chooser opens if whole input area is clicked


src/components/LinkChooser/types.ts
```diff
@@ -130,10 +130,15 @@ export type SearchInputProps = {
     machineService: MachineService;
     validation: Validation;
     onClear?: () => void;
+    wrapperProps?: {
```
SamuelAlev, line 133: is the wrapper needed? In the end it is the component emitting it

SamCreasey: After design review a checklist item was created since clicking inside the wrapper but not on the input would not open the dropdown, so I had to add additional listeners incase the area around the input was clicked

SamuelAlev: Feel a bit weird, can't we align it like that?
```
    validation: Validation;
    onClear?: () => void;
    onClick?: (event: MouseEvent<HTMLDivElement>) => void;
    onMouseDown?: (event: MouseEvent<HTMLDivElement>) => void;
```

SamuelAlev (approved):

PR #551, opened 2022-01-20
fix(link-chooser): create multiple custom links

SamCreasey:
Clickup Ticket: https://app.clickup.com/t/1vht3xx
This fixes the issue that when a custom link is selected, it overwrites the last custom link in local storage so that it always looks as if there is one link. Now each custom link has it's own unique Id and logic has been added to update/replace existing custom link with selected result (if it is custom), or when typing.


src/components/LinkChooser/utils/useManualComboBoxHandlers.ts
```diff
@@ -31,15 +31,18 @@ export const useManualComboBoxEventHandlers = (
             const item = state.collection.getItem(focusedKey);
             const { key } = event;
 
-            if (state.isOpen && SUBMISSION_KEYS.includes(key) && item) {
-                if (item.parentKey !== "search") {
+            if (state.isOpen && SUBMISSION_KEYS.includes(key)) {
+                if (item && item.parentKey !== "search") {
```
SamuelAlev, line 35:
```suggestion
                if (item?.parentKey !== "search") {
```

src/components/LinkChooser/state/actions.ts
```diff
@@ -13,14 +15,48 @@ export const updateQueryFromObject = assign<LinkChooserContext, DoneInvokeEvent<
...
+
+        if (isDisplayingLocalStorage) {
+            return context.searchResults;
+        } else {
+            if (isSelectedCustomLink) {
+                return [...resultsWithoutCustomLinks, selectedResult];
+            } else {
+                return [...resultsWithoutCustomLinks, createCustomLink(context.query)];
+            }
+        }
```
SamuelAlev, line 58:
```suggestion
        }

        if (isSelectedCustomLink) {
            return [...resultsWithoutCustomLinks, selectedResult];
        } else {
            return [...resultsWithoutCustomLinks, createCustomLink(context.query)];
        }
```

src/components/LinkChooser/LinkChooser.tsx
```diff
@@ -161,12 +165,10 @@ export const LinkChooser: FC<LinkChooserProps> = ({
                 selectedResult = createCustomLink(state.inputValue);
             }
         }
-
         send("CLOSE_DROPDOWN", { data: { selectedResult } });
-        if (selectedResult && state.selectedKey !== selectedResult.id) {
-            state.setSelectedKey(selectedResult.id);
+        if (selectedResult && selectedKey !== selectedResult.id) {
```
SamuelAlev, line 169:
```suggestion
        if (selectedKey !== selectedResult?.id) {
```

SamuelAlev (approved):
Just some nit picks

src/components/LinkChooser/utils/useManualComboBoxHandlers.ts
SamCreasey: This is not possible because if item is null then the if statement evaluates to true

src/components/LinkChooser/state/actions.ts
SamCreasey: fixed

PR #552, opened 2022-01-21
feat: Add autocomplete property for TextInput

hochreutenerl:
Required change for this [ClickUp ticket](https://app.clickup.com/t/z3yx18)

### DIFF
--- a/src/components/TextInput/TextInput.spec.tsx
+++ b/src/components/TextInput/TextInput.spec.tsx
@@ -123,4 +123,19 @@ describe("Text Input component", () => {
         mount(<StatefulInput disabled={false} />);
         cy.get(TEXT_INPUT_ID).should("not.have.attr", "disabled");
     });
+
+    it("has the autoComplete turned on", () => {
+        mount(<StatefulInput autocomplete={true} />);
+        cy.get(TEXT_INPUT_ID).should("have.attr", "autoComplete", "on");
+    });
+
+    it("has the autoComplete turned off", () => {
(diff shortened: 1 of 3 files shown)


src/components/TextInput/TextInput.spec.tsx
```diff
@@ -123,4 +123,19 @@ describe("Text Input component", () => {
         mount(<StatefulInput disabled={false} />);
         cy.get(TEXT_INPUT_ID).should("not.have.attr", "disabled");
     });
+
+    it("has the autoComplete turned on", () => {
+        mount(<StatefulInput autocomplete={true} />);
+        cy.get(TEXT_INPUT_ID).should("have.attr", "autoComplete", "on");
```
SamuelAlev, line 129: Good that HTML is case insensitive 😀
It is lowercase `autocomplete`

SamuelAlev (approved):

src/components/TextInput/TextInput.spec.tsx
hochreutenerl: But know you merged it? 🤔

SamuelAlev: Yes, it is not an issue, it could be `aUtOCoMpLeTe` and still be correct

hochreutenerl: Yep, but should we change it back? I think I had it all lowercase before but changed it without thinking too much just because the IDE suggested it. 🙄

PR #560, opened 2022-01-24
feat(tailwind)!: upgrade to tailwind v3

jonasbuechel:


tailwind.config.js
```diff
@@ -5,7 +5,7 @@ const plugin = require("tailwindcss/plugin");
 
 module.exports = {
     presets: [require("@frontify/arcade-tokens/tailwind")],
-    purge: ["./src/**/*.{js,ts,tsx}"],
```
hochreutenerl, line 8: It looks like in `tailwind-config.ts` we are accessing this value. I am not sure if it is still handled correctly there or if it is even still needed.

tailwind.config.js
```diff
@@ -23,7 +23,7 @@ module.exports = {
             },
             colors: {
                 transparent: "transparent",
-                current: "currentColor",
+                current: "currentColor", // Used for tailwind 3 when using a customized color palette -->  https://tailwindcss.com/docs/upgrade-guide#fill-and-stroke-use-color-palette
```
hochreutenerl, line 26: Nit: I don't think this extra comment is necessary.

tailwind.config.js
jonasbuechel: Good catch! I discussed it with Marco and we changed it in tailwind-config.ts as well 👍

triggertoo (approved):

deniaz (approved):

PR #563, opened 2022-01-26
fix(color-picker-flyout): slider animation when inside accordion

Walkeezy:
Regards this clickup ticket: https://app.clickup.com/t/23t26zw

Removed `AnimateSharedLayout ` as it was removed as of Framer Motion 5. Used `motion.li` to correctly detect changes in the layout and animate it instead.

### DIFF
--- a/src/components/Accordion/Accordion.stories.tsx
+++ b/src/components/Accordion/Accordion.stories.tsx
@@ -15,6 +15,12 @@ import { IconSize } from "@foundation/Icon/IconSize";
 import { action } from "@storybook/addon-actions";
 import { Meta, Story } from "@storybook/react";
 import React, { useState } from "react";
+import { Color } from "../../types/colors";
+import {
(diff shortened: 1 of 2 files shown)


src/components/Accordion/Accordion.stories.tsx
```diff
@@ -150,3 +156,26 @@ export const WithAdvancedFormControls: Story<AccordionProps> = () => {
         </AccordionComponent>
     );
 };
+
+export const WithColorPickerFlyout: Story<ColorPickerFlyoutProps> = ({
```
NadiaPosch, line 160: Instead of creating a separate story, I would incorporate this into the existing one

Walkeezy: Changed 👍🏻

NadiaPosch (approved):

PR #564, opened 2022-01-26
Feature: Support Flyout content width

msiric:
Recently encountered an issue while using the `Flyout` component where the content inside it wasn't nearly as wide as its default 400px min-width which ended up looking like this:

Screenshot 2022-01-26 at 12 00 49

This PR fixes the issue by adding an optional prop to fit the `Flyout` component's width to the content inside it (setting min-width to 0).

End result: 

Screenshot 2022-01-26 at 12 00 23


SamCreasey (approved):
All good 😄

PR #565, opened 2022-01-26
Fix: Make areas around icons clickable

msiric:
This fix resolves issues related to the areas around the caret and clear icons not being clickable:

https://user-images.githubusercontent.com/26199969/151188319-471106d8-ff75-4f22-adf1-cd16cb393ecd.mov

Issues are resolved using `position: absolute` while adding enough padding for the input to not get covered by the icons.

End result:

https://user-images.githubusercontent.com/26199969/151188450-245c3d28-1575-4e4c-b23b-7afeabcbc89c.mov

A test is added to the test suite to prevent this from happening in the future.

Admittedly, the test could be improved (it takes the width of the element and subtracts some pixels from it to target the area where the bug occurs), but it works well on all screen widths I used to verify its correctness.
Another test could be added to test the area around the clear icon, but this would require very precise coordinates to target correctly so I haven't done it at this stage.
If I find a nice way of doing it, I'll make sure to include it in the test suite in the future.

### DIFF
--- a/src/components/Trigger/Trigger.tsx
+++ b/src/components/Trigger/Trigger.tsx
@@ -16,11 +16,11 @@ export enum TriggerSize {
 export type TriggerProps = {
     disabled?: boolean;
     isOpen?: boolean;
-    clearable?: boolean;
     onClear?: () => void;
     buttonProps?: HTMLAttributes<HTMLElement>;
     isFocusVisible?: boolean;
     size?: TriggerSize;
+    showClear?: boolean;
 };
 
 export const Trigger: FC<TriggerProps> = ({
@@ -28,14 +28,12 @@ export const Trigger: FC<TriggerProps> = ({
     onClear,
     children,
     disabled = false,
-    clearable = false,
     isOpen = false,
     isFocusVisible = false,
     size = TriggerSize.Small,
+    showClear = false,
 }) => {
     const { focusProps: clearableFocusProps, isFocusVisible: isClearFocusVisible } = useFocusRing();
-
-    const showClear = clearable && !!onClear;
 
     return (
         <div
@@ -54,8 +52,8 @@ export const Trigger: FC<TriggerProps> = ({
             {children}
             <div
                 className={merge([
-                    "tw-flex-none tw-flex tw-items-center",
-                    size === TriggerSize.Large ? "tw-pr-5 tw-gap-1.5" : "tw-pr-3 tw-gap-1",
+                    "tw-flex-none tw-flex tw-items-center tw-absolute",
+                    size === TriggerSize.Large ? "tw-right-5 tw-gap-1.5" : "tw-right-3 tw-gap-1",
                 ])}
             >
                 {showClear && (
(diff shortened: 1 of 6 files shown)


src/components/Trigger/Trigger.tsx
```diff
@@ -54,8 +52,8 @@ export const Trigger: FC<TriggerProps> = ({
             {children}
             <div
                 className={merge([
-                    "tw-flex-none tw-flex tw-items-center",
-                    size === TriggerSize.Large ? "tw-pr-5 tw-gap-1.5" : "tw-pr-3 tw-gap-1",
+                    "tw-flex-none tw-flex tw-items-center tw-absolute",
```
SamCreasey, line 55: This will also affect the spacing on the multi-select and the color picker. Although from tests the color picker padding already does the job of providing enough space to the caret on the right side, the multi-select should also be updated to accommodate this change

src/components/Trigger/Trigger.tsx
```diff
@@ -64,11 +62,11 @@ export const Trigger: FC<TriggerProps> = ({
                         data-test-id="dropdown-clear-button"
                         aria-label="Clear selection"
                         className={merge([
-                            "tw-p-0 tw-outline-none",
+                            "tw-p-0 tw-outline-none tw-absolute tw-right-5",
```
SamCreasey, line 65: is this necessary? the parent container is a flexbox which provides the spacing between both buttons. I would assume that maybe the tw-absolute can be removed from the clear button and still work as intended

src/components/Trigger/Trigger.tsx
msiric: Just tried both the Color Picker and the Multi Select components and they both seem to have the correct spacing on the right of the caret icon.
Could you verify as well?

src/components/Trigger/Trigger.tsx
msiric: Without these styles, the small gap between the clear icon and the caret icon will not be clickable as shown in the "before" video of the PR.
If that's okay with you guys then it can be removed.

src/components/Trigger/Trigger.tsx
SamCreasey: Multi Select has the correct spacing on the right of the icon, however there is a small chance that the multi select options could cover the caret icon if the input is full of options

src/components/Trigger/Trigger.tsx
SamCreasey: Ah you are correct 👍

src/components/Trigger/Trigger.tsx
msiric: Ah gotcha!
Let me see if something can be done about this.

msiric: Please take a look at this:

https://user-images.githubusercontent.com/26199969/151532476-25d2853f-dec6-41d1-9bd4-3c9874df2c52.mov

It seems that it doesn't affect the Multi Select component.
There might be some other edge cases that I'm missing though.

SamCreasey: Screenshot 2022-01-28 at 11 52 01
Hopefully this demonstrates the issue :)

SamCreasey: It's only a small visual glitch, but might as well fix it in this pr

msiric: Thanks, makes sense now.

msiric: Should be fixed now.

SamCreasey: looks great 😄

SamCreasey (approved):

PR #568, opened 2022-01-27
fix: RichTextEditor toolbar position when scrolling

NadiaPosch:
concerns this ticket: https://app.clickup.com/t/1vqjqh3
Position of the toolbar should now also work when scrolling


triggertoo (approved):
💯

PR #569, opened 2022-01-28
feat(Accordion): allow sections to be open by default

CelienB:
This PR allows for Accordion sections being open by default when the `active` header property is set to `true`.
https://app.clickup.com/t/1y7qr5d


src/components/Accordion/Accordion.spec.tsx
```diff
@@ -66,6 +66,22 @@ describe("Accordion Component", () => {
...
+                    <TextInput />
+                </AccordionItem>
+                <AccordionItem header={{ children: "2", active: true }}>
+                    <TextInput />
+                </AccordionItem>
+                <AccordionItem header={{ children: "3" }}>3</AccordionItem>
+            </Accordion>,
+        );
+
+        cy.get(ACCORDION_ITEM_CONTENT_ID).should("exist");
```
SamuelAlev, line 82: Not a good test, you need to verify if the 2nd is correctly open and the rest still closed

src/components/Accordion/Accordion.spec.tsx
```diff
@@ -66,6 +66,36 @@ describe("Accordion Component", () => {
...
+    it("opens select sections initially", () => {
+        mount(
+            <Accordion>
+                <AccordionItem header={{ children: "1", active: true }}>
+                    <TextInput />
+                </AccordionItem>
+                <AccordionItem header={{ children: "2" }}>
+                    <TextInput />
+                </AccordionItem>
+                <AccordionItem header={{ children: "3" }}>3</AccordionItem>
```
SamuelAlev, line 78: You can add multiple tests
- It should have none opened by default
- It should have one opened
- It should have one opened which is not the first one
- It should have multiple opened

src/components/Accordion/Accordion.tsx
```diff
@@ -76,6 +83,7 @@ const AriaAccordionItem: FC<AriaAccordionItemProps> = ({ item, state, header })
                         variants={{
                             open: { height: "auto", overflow: "visible" },
                             collapsed: { height: 0, overflow: "hidden" },
+                            still: {},
```
SamuelAlev, line 86: What is it used for? There is no reference for it in the code

SamuelAlev (changes requested):
No changes needed yet, I'm just "locking" the PR so I can test it

SamuelAlev (approved):

src/components/Accordion/Accordion.tsx
```diff
@@ -36,6 +38,19 @@ const AriaAccordionItem: FC<AriaAccordionItemProps> = ({ item, state, header })
     const { buttonProps, regionProps } = useAccordionItem({ item }, state, triggerRef);
     const isOpen = state.expandedKeys.has(item.key) && item.props.children;
     const { isFocusVisible, focusProps } = useFocusRing();
+    const [isActive, setIsActive] = useState(false);
+    const isActiveByDefault = header.active;
```
NadiaPosch, line 42: Personally, I'd find it more intuitive if you combine this with the `active` prop:
```suggestion
    const [isActive, setIsActive] = useState(header.active);
```

src/components/Accordion/Accordion.tsx
```diff
@@ -36,6 +38,19 @@ const AriaAccordionItem: FC<AriaAccordionItemProps> = ({ item, state, header })
...
+
+    useEffect(() => {
+        if (isActiveByDefault) {
```
NadiaPosch, line 45:
```suggestion
        if (isActive) {
```

src/components/Accordion/Accordion.tsx
```diff
@@ -36,6 +38,19 @@ const AriaAccordionItem: FC<AriaAccordionItemProps> = ({ item, state, header })
...
+            state.toggleKey(item.key);
+        }
+        // We add a timeout to avoid isActive being set to true before
+        // first re-render and play animations on load.
+        setTimeout(() => setIsActive(true), 50);
```
NadiaPosch, line 50:
```suggestion
        // We add a timeout to avoid isActive being set to false before
        // the first re-render, so that the animations won't play on load
        setTimeout(() => setIsActive(false), 50);
```

src/components/Accordion/Accordion.tsx
```diff
@@ -36,6 +38,19 @@ const AriaAccordionItem: FC<AriaAccordionItemProps> = ({ item, state, header })
...
+    }, []);
+
+    const shouldPlayAnimations = !isActive && isActiveByDefault;
```
NadiaPosch, line 53:
```suggestion
```

src/components/Accordion/Accordion.tsx
```diff
@@ -70,7 +85,7 @@ const AriaAccordionItem: FC<AriaAccordionItemProps> = ({ item, state, header })
                 {item.props.children && isOpen && (
                     <motion.div
                         key={item.key}
-                        initial={"collapsed"}
+                        initial={shouldPlayAnimations ? false : "collapsed"}
```
NadiaPosch, line 88:
```suggestion
                        initial={isActive ? false : "collapsed"}
```

src/components/Accordion/Accordion.tsx
```diff
@@ -82,7 +97,7 @@ const AriaAccordionItem: FC<AriaAccordionItemProps> = ({ item, state, header })
                     >
                         <div {...regionProps} className="tw-px-8 tw-pb-6">
                             <motion.div
-                                initial={{ opacity: 0 }}
+                                initial={shouldPlayAnimations ? false : { opacity: 0 }}
```
NadiaPosch, line 100:
```suggestion
                                initial={isActive ? false : { opacity: 0 }}
```

PR #576, opened 2022-01-31
docs: Implement new Sidebar and Folder Structure

deniaz:


.storybook/preview.tsx
```diff
@@ -4,7 +4,7 @@ import { dark, light } from "./frontifyTheme";
 export const parameters = {
     options: {
         storySort: {
-            order: ["Foundation", "Components"],
```
jonasbuechel, line 7: Foundation is now at the end and only contains "Icon". I think we should either put "Icon" into "Components" else or keep "Foundation" at a higher level in the order.

Screenshot 2022-02-02 at 17 06 34

jonasbuechel (changes requested):

.storybook/preview.tsx
jonasbuechel: we decided to do that in a separated PR

jonasbuechel (approved):

PR #581, opened 2022-02-01
fix: decouple link chooser from clarify

NadiaPosch:
Concerns this ticket: https://app.clickup.com/t/23yxnvc

This PR removes some of the Business Logic from the LinkChooser by replacing it with a more flexibel approach to add custom `extraSections`.

Keyboard navigation is only partly working and will be addressed in this follow-up ticket: https://app.clickup.com/t/24g28a0

### DIFF
--- a/src/components/LinkChooser/state/machine.ts
+++ b/src/components/LinkChooser/state/machine.ts
@@ -7,21 +7,21 @@ import {
     clearSelectedResult,
     copyLinkToClipboard,
     emitSelectSearchResult,
+    fetchExtraSectionResults,
     fetchGlobalSearchResults,
-    fetchTemplateSearchResults,
+    fillResultsWithNewRecentQueries,
+    interruptFetching,
     openPreviewContext,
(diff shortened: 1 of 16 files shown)


src/components/LinkChooser/state/machine.ts
```diff
@@ -169,28 +169,23 @@ export const linkChooserMachine = createMachine<LinkChooserContext, DoneInvokeEv
...
                             },
-                        },
-                    },
-                    [DropdownState.Templates]: {
-                        ...initializeSectionState(SectionState.Fetching, "fetchTemplates", fetchTemplateSearchResults),
-                        on: {
-                            GO_TO_DEFAULT: `${DropdownState.Default}.${SectionState.Fetching}`,
                             CLEARING: {
                                 target: DropdownState.Default,
                                 actions: [...clearingActions],
```
triggertoo, line 191: no need to destruct ;)

triggertoo (approved):

PR #583, opened 2022-02-02
feat: add tailwind config to exports

jonasbuechel:


vite.config.ts
```diff
@@ -43,7 +43,7 @@ export default defineConfig({
         sourcemap: true,
         minify: true,
         rollupOptions: {
-            input: [...componentsPath.map((element) => element.path), "src/styles.css"],
+            input: [...componentsPath.map((element) => element.path), "src/styles.css", "tailwind.config.js"],
```
SamuelAlev, line 46: Probably not needed as it is already a `.js` file and use `require`, wdyt?

jonasbuechel: I added it to ensure `tailwind.config.js` is copied to `/dist` is there another way to achieve it?

SamuelAlev: Make sense 👍

deniaz (approved):

vite.config.ts
jonasbuechel: Thanks for double-checking 🙌

PR #584, opened 2022-02-02
feat: add hoverable state to card component

msiric:
Adds an optional `isHoverable` prop to the `Card` component to support the Brand Listening card design.

### DIFF
--- a/src/components/Card/Card.tsx
+++ b/src/components/Card/Card.tsx
@@ -4,17 +4,17 @@ import { merge } from "@utilities/merge";
 import React, { FC, ReactChild } from "react";
 
 export type CardProps = {
-    isHoverable?: boolean;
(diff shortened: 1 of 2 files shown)


src/components/Card/Card.tsx
```diff
@@ -1,14 +1,22 @@
 /* (c) Copyright Frontify Ltd., all rights reserved. */
 
+import { merge } from "@utilities/merge";
 import React, { FC, ReactChild } from "react";
 
 export type CardProps = {
+    isHoverable?: boolean;
```
SamuelAlev, line 7:
```suggestion
    hoverable?: boolean;
```
WDYT?

msiric: Makes sense 👍

PR #594, opened 2022-02-07
fix(color-picker-flyout): wrap palettes and add padding option

SamCreasey:
In order for the overlay to not cover the intercom button the position must be calculated externally and passed in to the useOverlayPosition, so that the bottom margin can be specified for each case and taken into account when determining if the overlay should be displayed above or below the trigger.
The Color Picker Flyout storybook now has the option to pass the vertical margins to see the effect of the calculation.
Color palettes in the color picker now wrap to avoid overflow


src/components/Flyout/helpers/shouldDisplayAbove.ts
```diff
@@ -0,0 +1,37 @@
...
+    triggerRef: MutableRefObject<HTMLDivElement | null>;
+    overlayHeight?: number;
+    margins: {
+        top: number;
+        bottom: number;
+    };
+    offset?: number;
+};
+
+export const shouldDisplayAbove = ({
```
NadiaPosch, line 15: Is there a reason that you pass in an object instead of separate arguments? If not, I'd refactor the method

NadiaPosch (approved):

PR #606, opened 2022-02-09
feat: button rounded property

urietberger:
This change introduces the property to render a button with a fully rounded radius. We need this to reflect the design of the legacy component `IconButton`, which was heavily used as a trigger for settings- and actions-menus, etc. Unless we have migrated all its occurrences to React, we can't easily switch to the new square design. As soon as we can switch to the new design, this property will be removed.

### DIFF
--- a/src/components/Button/Button.spec.tsx
+++ b/src/components/Button/Button.spec.tsx
@@ -3,7 +3,7 @@
 import { mount } from "@cypress/react";
 import IconIcons from "@foundation/Icon/Generated/IconIcons";
 import React from "react";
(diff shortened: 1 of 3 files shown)


urietberger:
@jonasbuechel Here's the PR about the rounding property. Strangely, Storybook doesn't apply the arguments I set in the story. At least not when I run Storybook locally. Is this a known issue?

SamuelAlev (approved):
Hijacking the PR as I'll make a release soon 😀 
Everything looks good 👍

PR #612, opened 2022-02-11
feat(RichTextEditor): rebuild with Plate

Walkeezy:
https://plate.udecode.io/

Todo:
- [x] Link element
- [x] Clear functionality
- [x] ~~RTE does dynamically resize to the Placeholder length~~ -> RTE is now always the width of the parent
- [x] Ensure HTML import works (do we still need our custom logic or is this provided by slate?)
- [x] Tests
- [x] Cleanup HTML structure for lists (currently looks like this: p > ul > li)
- [x] Fix import for lists (doesn't render correctly)


src/components/RichTextEditor/RichTextEditor.tsx
```diff
@@ -1,156 +1,132 @@
...
+import { UnderlineMark } from "./components/marks/underline";
 import { Toolbar } from "./Toolbar";
-import { clearEditor } from "./utils/editor/clear";
-import { parseRawValue } from "./utils/editor/parseRawContent";
+import { clearEditor } from "./utils/clearEditor";
 
 export type RichTextEditorProps = {
     placeholder?: string;
-    value?: string;
+    value?: any;
```
triggertoo, line 43: the idea there was to provide a serialized json object as string, as we provide also a serialized json object in the onChange-/onBlur-event

Walkeezy: Changed it back to serialized json

Walkeezy:
Before this can be merged, we need to adapt the changes from #618.

src/components/RichTextEditor/RichTextEditor.tsx
```diff
@@ -31,126 +17,52 @@ export type RichTextEditorProps = {
...
 export const ON_SAVE_DELAY_IN_MS = 500;
-const isModifyingKey = (key: string) => !["Alt", "Control", "Meta", "Shift"].includes(key);
 
 export const RichTextEditor: FC<RichTextEditorProps> = ({
     value: initialValue,
     placeholder = "",
     readonly = false,
     clear = false,
     onTextChange,
-    onBlur,
```
triggertoo, line 67: I'd keep the onBlur event, so we can persist the content when leaving the editor

src/components/RichTextEditor/RichTextEditor.tsx
```diff
@@ -31,126 +17,52 @@ export type RichTextEditorProps = {
...
-                                DoneInvokeEvent<ToolbarData>
-                            >,
-                        }}
-                    >
-                        <Toolbar />
-                    </ToolbarContext.Provider>
-                )}
-            </Slate>
+                    styles={{ root: { border: "none", background: "#ffffff" } }}
+                    //classNames={{ root: "tw-drop-shadow-md" }}
```
triggertoo, line 61: can this be removed?

triggertoo: or even better we pass in the classNames and remove the styles ;)

src/components/RichTextEditor/RichTextEditor.tsx
```diff
@@ -31,126 +17,52 @@ export type RichTextEditorProps = {
...
+    const editor = usePlateEditorState();
 
     useEffect(() => {
-        if (clear) {
-            const emptyValue = parseRawValue();
-            clearEditor(editor);
-            Transforms.insertNodes(editor, emptyValue);
-            send("RESET_TEXT");
+        if (clear && editor) {
+            // clear selection, history and children of editor
```
NadiaPosch, line 38: This can be removed IMO

src/components/RichTextEditor/components/elements/link.tsx
```diff
@@ -0,0 +1,18 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+import { PlateRenderElementProps } from "@udecode/plate";
+import React from "react";
+
+export const LinkElement = (props: PlateRenderElementProps) => {
+    return (
```
NadiaPosch, line 7: You could make a direct return

src/components/RichTextEditor/components/elements/link.tsx
```diff
@@ -0,0 +1,18 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+import { PlateRenderElementProps } from "@udecode/plate";
+import React from "react";
+
+export const LinkElement = (props: PlateRenderElementProps) => {
```
NadiaPosch, line 6: You could destruct and make a direct return 
```suggestion
export const LinkElement: FC<PlateRenderElementProps> = ({ element, attributes, children }) => (
```

src/components/RichTextEditor/editor-config.ts
```diff
@@ -0,0 +1,72 @@
...
+    createPlugins,
+    createSoftBreakPlugin,
+    createStrikethroughPlugin,
+    createUnderlinePlugin,
+    ELEMENT_LI,
+    ELEMENT_LIC,
+    ELEMENT_LINK,
+    ELEMENT_OL,
+    ELEMENT_UL,
+    LinkElement,
```
NadiaPosch, line 23: I think this is the wrong import

src/components/RichTextEditor/editor-config.ts
```diff
@@ -0,0 +1,72 @@
...
+    MARK_BOLD,
+    MARK_CODE,
+    MARK_ITALIC,
+    MARK_STRIKETHROUGH,
+    MARK_UNDERLINE,
+} from "@udecode/plate";
+import { ListItemElement } from "./components/elements/list-item";
```
NadiaPosch, line 30: You could create an `index.ts` in the components directory where you export all components

src/components/RichTextEditor/Toolbar.tsx
```diff
@@ -3,126 +3,73 @@
...
-                    className="tw-popper-arrow tw-absolute tw-w-2 tw-h-2 tw-pointer-events-none before:tw-absolute before:tw-z-10 before:tw-w-2 before:tw-h-2 before:tw-bg-white before:tw-rotate-45"
-                ></div>
-            </div>
-        </>
+        <div
+            data-test-id="toolbar"
+            className="tw-flex tw-p-0.5 tw-items-center tw-bg-white tw-rounded tw-shadow-mid tw-gap-0.5"
+        >
+            <ButtonGroup testId="text-alignment-buttons">
+                <AlignToolbarButton value="left" icon={<IconTextAlignLeft />} />
```
NadiaPosch, line 54: The icons are bigger than in the current RTE, I think you need to specify the correct size

src/components/RichTextEditor/RichTextEditor.tsx
Walkeezy: Removed classnames, styles are needed to overwrite the default ones.

src/components/RichTextEditor/RichTextEditor.tsx
Walkeezy: Removed

src/components/RichTextEditor/editor-config.ts
Walkeezy: Good catch, thanks!

src/components/RichTextEditor/RichTextEditor.tsx
Walkeezy: Makes sense, added.

src/components/RichTextEditor/components/elements/link.tsx
Walkeezy: Refactored

src/components/RichTextEditor/editor-config.ts
Walkeezy: Done 💯

src/components/RichTextEditor/Toolbar.tsx
Walkeezy: Current RTE is not according to Figma anway, so I'll leave it as it is and we can tackle that, as soon as the design is fixed an we implement the actual design of the RTE.

src/components/RichTextEditor/TextStyleDropdown/ListBox.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { ListBoxSection } from "./ListBoxSection";
+import { Option } from "./Option";
+
+interface ListBoxProps extends AriaListBoxOptions<unknown> {
+    listBoxRef?: RefObject<HTMLUListElement>;
+    state: ListState<unknown>;
+}
+
+export const ListBox: FC<ListBoxProps> = (props) => {
+    const ref = useRef<HTMLUListElement>(null);
```
NadiaPosch, line 14: Not sure if it makes a difference, but I'd add `null` to make the typing more specific (same for all the other useRefs)
```suggestion
    const ref = useRef<HTMLUListElement | null>(null);
```

src/components/RichTextEditor/RichTextEditor.tsx
```diff
@@ -25,17 +25,18 @@ export const RichTextEditor: FC<RichTextEditorProps> = ({
     readonly = false,
     clear = false,
     onTextChange,
+    onBlur,
 }) => {
     const editableProps: EditableProps = {
         placeholder: placeholder,
         readOnly: readonly,
+        onBlur: () => onBlur,
```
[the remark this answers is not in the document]
Walkeezy: Changed

src/components/RichTextEditor/TextStyleDropdown/ListBox.tsx
Walkeezy: Component not used anymore / deleted.

triggertoo (approved):

PR #617, opened 2022-02-14
feat(badge): next badge styles

SamCreasey:
Ticket: https://app.clickup.com/t/256tm5y
- Added disabled, emphasis and size properties
- added new styles for emphasis and disabled states,
- modified existing style classes to match new arcade tokens
- added new tests for new props
- updated storybook with new properties

### DIFF
--- a/src/components/Badge/Badge.tsx
+++ b/src/components/Badge/Badge.tsx
@@ -153,7 +153,7 @@ export const Badge: FC<BadgeProps> = ({
                 <span className="tw-text-center tw-text-xxs tw-font-sans tw-font-normal">{children}</span>
                 {onDismiss && (
                     <button
-                        data-test-id="badge-icon"
+                        data-test-id="badge-dismiss"
                         onClick={(event) => {
                             event.stopPropagation();
                             onDismiss();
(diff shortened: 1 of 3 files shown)


src/components/Badge/Badge.tsx
```diff
@@ -28,55 +28,90 @@ export enum BadgeStyle {
...
+
 export type BadgeProps = PropsWithChildren<{
     style?: BadgeStyle;
     icon?: ReactElement<IconProps>;
     status?: BadgeStatus | string;
     onClick?: () => void;
     onDismiss?: () => void;
+    disabled?: boolean;
+    emphasis?: "Strong" | "None";
+    size?: "s" | "m";
```
Walkeezy, line 101: Would it make sense to create Enums for `emphasis` and `size`?

Walkeezy (approved):

src/components/Badge/Badge.tsx
SamCreasey: That is a good question. I will ask in the arcade team about this because I noticed some components deal exclusively in enums and others don't even touch them 🤔

thisguyscode:
@SamCreasey - Sorry this isn't clear from the Figma component (I'll update it), but the status color is actually a custom color, static color.

1. The user can choose a totally custom color for statuses
2. Even if this was not the case the color should probably not change between themes, which would be the case here.

Basically instead of a `BadgeStatus enum` it should be `StatusColor: Color`.

PR #624, opened 2022-02-16
feat: Import new Fondue icons

thisguyscode:
- Uses `arcade-icons` as source for new icons, without removing the icons that were already generated from old source
- Changes generation script to remove some sizing and such
- Adds generation of dynamic icon, which includes all variants of the icon shape
- Refactored the icon generation in general

Essentially there should be no breaking changes, and we can slowly migrate to the new Icons

### DIFF
--- a/src/components/RichTextEditor/RichTextEditor.spec.tsx
+++ b/src/components/RichTextEditor/RichTextEditor.spec.tsx
@@ -92,9 +92,9 @@ describe('RichTextEditor Component', () => {
 
         insertTextAndOpenToolbar();
         cy.get(TOOLBAR).should('be.visible');
-        cy.get(TOOLBAR_GROUP_0).find('> span').should('have.length', 1);
-        cy.get(TOOLBAR_GROUP_1).find('> span').should('have.length', 2);
-        cy.get(TOOLBAR_GROUP_2).find('> span').should('have.length', 1);
+        cy.get(TOOLBAR_GROUP_0).find('[data-testid=ToolbarButton]').should('have.length', 1);
+        cy.get(TOOLBAR_GROUP_1).find('[data-testid=ToolbarButton]').should('have.length', 2);
+        cy.get(TOOLBAR_GROUP_2).find('[data-testid=ToolbarButton]').should('have.length', 1);
         cy.get(TOOLBAR_GROUP_3).should('not.exist');
     });
 


jonasbuechel (changes requested):
It works great! 🥳  I added some suggestions regarding usage of TypeScript.

Main issue: the bundle size will be increased by >5.5 MB! As tree-shaking is not working yet in Arcade we need to wait to merge until we fixed it.
Screenshot 2022-02-22 at 17 11 39

thisguyscode:
@jonasbuechel Thanks for the typescript tips, that's great help. Looking forward to solving tree-shaking :D

thisguyscode:
@jonasbuechel What's the best way to apply these suggestions? Via the Github UI or should I do the manual work locally and push the changes?

jonasbuechel:
> 

You can directly commit it via GitHub UI. Just ensure to doublecheck my code as well and have a test run locally (get the changes via pull)

imoutaharik:
@jonasbuechel This Pr is ready but deployment is failing.
FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory

scripts/generateReactIcons.ts
```diff
@@ -2,52 +2,107 @@
...
 import { writeFile, readFile } from "fs/promises";
+import { camelCase, toUpper } from "lodash";
 // @ts-ignore
 import { transform } from "@svgr/core";
 import { Entry } from "fast-glob/out/types";
 import { IconTemplate } from "../src/foundation/Icon/IconTemplate";
+import { IconComponent, IconTemplateDynamic } from "../src/foundation/Icon/IconTemplateDynamic";
 
-export const GENERATED_ICONS_INDEX_PATH = "src/foundation/Icon/Generated/index.ts";
+const ICON_SOURCE_DIRECTORY = `node_modules/@frontify/arcade-icons/icons/`;
```
jonasbuechel, line 12:
```suggestion
const ICON_SOURCE_DIRECTORY = `node_modules/@frontify/fondue-icons/icons/`;
```

jonasbuechel (changes requested):

src/foundation/Icon/Generated/IconArrowBidirectional.tsx
```diff
@@ -0,0 +1,25 @@
+import React, { memo } from "react";
+import { GeneratedIconProps } from "@foundation/Icon/IconProps";
+import { IconSize } from "@foundation/Icon/IconSize";
+
+import IconArrowBidirectional12 from './IconArrowBidirectional12';;
```
jonasbuechel, line 5: There seems to be an error in the template we use for generation which produces double semicolons.
```suggestion
import IconArrowBidirectional12 from './IconArrowBidirectional12';
```

jonasbuechel (changes requested):

src/foundation/Icon/Generated/IconArrowBidirectional.tsx
imoutaharik: Fixed!

imoutaharik (approved):

PR #631, opened 2022-02-23
feat(Modal): create Modal component

CelienB:
This (draft) PR adds a Modal component to Arcade.
Ticket: [Marketplace - Component: Modal](https://app.clickup.com/t/1zftb6j)


src/components/Modal/types.ts
```diff
@@ -0,0 +1,62 @@
...
+import { AriaDialogProps } from "@react-types/dialog";
+import { PatternDesign, PatternTheme } from "@foundation/Pattern";
+import { ScrollWrapperDirection } from "@components/ScrollWrapper/types";
+import { ButtonHTMLAttributes, HTMLAttributes, ReactElement, RefObject } from "react";
+
+export type ModalButton = {
+    label: string;
+    action: () => void;
+    ref?: RefObject<HTMLButtonElement>;
+    ariaButtonProps?: ButtonHTMLAttributes<HTMLButtonElement>;
```
SamCreasey, line 13: not using so can be removed

src/components/Modal/types.ts
```diff
@@ -0,0 +1,62 @@
...
+
+import { OverlayProps } from "@react-aria/overlays";
+import { AriaDialogProps } from "@react-types/dialog";
+import { PatternDesign, PatternTheme } from "@foundation/Pattern";
+import { ScrollWrapperDirection } from "@components/ScrollWrapper/types";
+import { ButtonHTMLAttributes, HTMLAttributes, ReactElement, RefObject } from "react";
+
+export type ModalButton = {
+    label: string;
+    action: () => void;
```
SamCreasey, line 11: change to onClick

src/components/Modal/Modal.stories.tsx
```diff
@@ -0,0 +1,218 @@
...
+                        <ModalFooter
+                            buttons={[
+                                { label: "Okay", action: () => action("click") },
+                                {
+                                    label: "Cancel",
+                                    action: () => state.close(),
+                                },
+                            ]}
+                        />
+                    }
```
SamCreasey, line 131: @SamuelAlev do you think for the header, footer and body we should just pass props instead of full components and automatically wrap in the ModalHeader, ModalFooter and ModalBody components respectively?

SamCreasey, line 131: @SamuelAlev do you think for the header, footer and body we should just pass props instead of full components and automatically wrap in the ModalHeader, ModalFooter and ModalBody components respectively?

src/components/ScrollWrapper/ScrollWrapper.tsx
```diff
@@ -0,0 +1,41 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+import React, { FC, useRef } from "react";
+import { useScrollWrapper } from "./hooks/useScrollWrapper";
+import { ScrollWrapperDirection, scrollWrapperDirections, ScrollWrapperProps } from "./types";
+
+export const ScrollWrapper: FC<ScrollWrapperProps> = ({ direction, children }) => {
```
SamCreasey, line 7: Add default direction here rather than below

SamCreasey, line 7: Add default direction here rather than below

src/components/ScrollWrapper/hooks/useScrollWrapper.ts
```diff
@@ -0,0 +1,37 @@
...
+    const onScroll = (event: UIEvent) => {
+        const target = event.target as HTMLElement;
+        setScrollDimensions({
+            top: target.scrollTop,
+            height: target.scrollHeight,
+            left: target.scrollLeft,
+            width: target.scrollWidth,
+        });
+    };
+
```
SamCreasey, line 17: add useEffect to set the scroll wrapper values once the element has loaded, incase area is too small to scroll and having the height set to infinity will cause the bottom shadow to permanently appear

SamCreasey, line 17: add useEffect to set the scroll wrapper values once the element has loaded, incase area is too small to scroll and having the height set to infinity will cause the bottom shadow to permanently appear

src/foundation/Pattern/PatternScaleOrigin.tsx
```diff
@@ -0,0 +1,3 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+export type PatternScaleOrigin = Array<"top" | "left" | "bottom" | "right">;
```
SamuelAlev, line 3:
```suggestion
export type PatternScaleOrigin = ("top" | "left" | "bottom" | "right")[];
```

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,99 @@
...
+    return (
+        <motion.div
+            variants={UNDERLAY_VARIANTS}
+            initial="initial"
+            animate="show"
+            exit="exit"
+            style={{ background: "rgba(0, 0, 0, .5)" }}
+            onPointerDown={onPointerDown}
+            data-is-underlay={true}
+            className="tw-fixed tw-top-0 tw-left-0 tw-bottom-0 tw-right-0 tw-z-50 tw-flex tw-justify-center tw-items-center tw-p-3"
```
SamuelAlev, line 61:
```suggestion
            className="tw-fixed tw-inset-0 tw-z-50 tw-flex tw-justify-center tw-items-center tw-p-3"
```

src/components/Modal/Modal.stories.tsx
```diff
@@ -0,0 +1,210 @@
...
+            options: ["None", ...Object.values(ModalHeaderVariant)],
+            mapping: { None: null, ...Object.values(ModalHeaderVariant) },
+            defaultValue: ModalHeaderVariant.Default,
+            control: { type: "select" },
+        },
+    },
+    decorators: [
+        (Story) => (
+            <OverlayProvider>
+                <OverlayContainer>
```
SamuelAlev, line 88: Would be cool to not have to wrap our whole application in those Provider and Container

deniaz: Hey @SamuelAlev 

@jonasbuechel asked me to have a look at the PR. From what I understand, unfortunately we won't get around that.

> **The application is contained in an OverlayProvider**, which is used to hide the content from screen readers with aria-hidden while a modal is open. In addition, each modal **must be contained in an OverlayContainer**, which uses a React Portal to render the modal at the end of the document body.
>
> Note: `useModal` only hides content within parent OverlayProvider components. [...]

https://react-spectrum.adobe.com/react-aria/useDialog.html

SamuelAlev: We can not wrap our app where we use Terrific with some React components, it would make `React -> Terrific -> React`.
In full React page (marketplace, brand listening, ...) it should be ok however.

deniaz: If that's an issue I only see three ways going forward:

- Find a library that does it differently and redo this PR
- Implement a solution without any third-party dependencies
- Find a way for the Spectrum components to work the way you need in Clarify

I'll approve the PR from my side, but it's probably your decision on how you want to go forward with it. I'm also unsure about the importance of a11y in Clarify during the Terrific → React transition phase.

deniaz (approved):

src/components/Modal/Modal.stories.tsx
jonasbuechel: We can use the modal component without OverlayProvider as well, but we will lose the accessibility optimizations this way. I'll give it a try within Terrific. As the whole Terrific app is in bad shape accessibility-wise, I think we can live with this compromise for legacy views.

jonasbuechel: It works within a react component in clarify 👍 
Screenshot 2022-03-31 at 11 16 06

SamuelAlev (approved):

PR #632, opened 2022-02-25
Fix/flyout component border without footer

CelienB:


CelienB:
@urietberger I've re-created a Story of your exact use case in Storybook. The prominent border issue doesn't really seem to show if you take a look at what's rendered in the Story.
Could you try this out and maybe help me getting it to show?

SamuelAlev:
Stale

PR #633, opened 2022-02-25
feat(Typography)!: change font weights

thisguyscode:


src/typography/Code/Code.tsx
```diff
@@ -1,7 +1,7 @@
 import { merge } from "@utilities/merge";
 import React, { FC, PropsWithChildren } from "react";
 
-type CodeHeight = "medium" | "strong";
+type CodeHeight = "default" | "strong" | /* DEPRECATE */ "medium";
```
jonasbuechel, line 4: When using TypeScript this will fail on most setups anyway. I think we can change it instead of deprecate it and make it a breaking change as we are in an early stage anyways 🤔

jonasbuechel (changes requested):
Let's break it instead of deprecate it. We're anyways creating multiple breaking releases the next months 🤔  When we describe it properly in the changelogs I think this will be fine.

src/typography/Code/Code.tsx
thisguyscode: @jonasbuechel That's up to you, but I'd really like to bundle breaking changes together into bigger releases when we can avoid it.

jonasbuechel: Well, bundling lot of breaking changes means usually it's a bigger hurdle to bring it to Clarify. I think we should aim for smaller releases in general.

src/typography/Text/Text.tsx
```diff
@@ -5,7 +5,7 @@ import React, { FC, PropsWithChildren } from "react";
 import { decorationMap, displayMap, overflowMap, whitespaceMap, wordBreakMap } from "../shared/records";
 import { SharedTypographyProps } from "../shared/types";
 
-type TextWeight = "regular" | "strong" | "x-strong";
+type TextWeight = "default" | "strong" | "x-strong" | /* DEPRECATE */ "regular";
```
jonasbuechel, line 8:
```suggestion
type TextWeight = "default" | "strong" | "x-strong";
```

src/typography/Text/Text.tsx
```diff
@@ -19,9 +19,11 @@ export type TextProps = PropsWithChildren<
 >;
 
 const weightMap: Record<TextWeight, string> = {
+    default: "tw-font-regular",
+    strong: "tw-font-medium",
+    "x-strong": "tw-font-bold",
+    /* DEPRECATE */
     regular: "tw-font-regular",
```
jonasbuechel, line 26:
```suggestion
```

src/typography/Heading/Heading.tsx
```diff
@@ -5,7 +5,7 @@ import React, { FC, PropsWithChildren } from "react";
 import { decorationMap, displayMap, overflowMap, whitespaceMap, wordBreakMap } from "../shared/records";
 import { SharedTypographyProps } from "../shared/types";
 
-type HeadingWeight = "medium" | "strong";
+type HeadingWeight = "default" | "strong" | /* DEPRECATE */ "medium";
```
jonasbuechel, line 8:
```suggestion
type HeadingWeight = "default" | "strong";
```

src/typography/Heading/Heading.tsx
```diff
@@ -19,8 +19,10 @@ export type HeadingProps = PropsWithChildren<
 >;
 
 const weightMap: Record<HeadingWeight, string> = {
-    medium: "tw-font-medium",
+    default: "tw-font-medium",
     strong: "tw-font-bold",
+    /* DEPRECATE */
+    medium: "tw-font-medium",
```
jonasbuechel, line 25:
```suggestion
```

src/typography/Code/Code.tsx
jonasbuechel, line 4:
```suggestion
type CodeHeight = "default" | "strong";
```

src/typography/Code/Code.tsx
```diff
@@ -15,8 +15,10 @@ export type CodeProps = PropsWithChildren<{
 }>;
 
 const weightMap: Record<CodeHeight, string> = {
-    medium: "tw-font-medium",
+    default: "tw-font-regular",
     strong: "tw-font-bold",
+    /* DEPRECATE */
+    medium: "tw-font-medium",
```
jonasbuechel, line 21:
```suggestion
```

jonasbuechel (changes requested):
Let's remove the deprecations

jonasbuechel (approved):

PR #640, opened 2022-03-01
feat(textInput): Add control to enable or disable spellcheck

imoutaharik:
Ticket: #275n4v7

### DIFF
--- a/src/components/TextInput/TextInput.tsx
+++ b/src/components/TextInput/TextInput.tsx
@@ -123,6 +123,8 @@ export const TextInput: FC<TextInputProps> = ({
         return type;
     };
 
+    const spellcheckProp = typeof spellcheck === "boolean" ? { spellCheck: spellcheck } : null;
+
     return (
         <div
             {...focusProps}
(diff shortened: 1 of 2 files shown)


jonasbuechel:
@imoutaharik I just changed the title to "feat:" because it adds functionality. I also linked the GitHub Issue, so it will close once it's merged 👍

src/components/TextInput/TextInput.tsx
```diff
@@ -168,6 +170,7 @@ export const TextInput: FC<TextInputProps> = ({
                 autoComplete={autocomplete ? "on" : "off"}
                 size={size}
                 data-test-id="text-input"
+                spellCheck={spellcheck ?? true}
```
jonasbuechel, line 173: [The default](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/spellcheck) seems to be unset so I think we should not enforce `true nor `false` as a default. What do you think?

jonasbuechel (changes requested):

src/components/TextInput/TextInput.tsx
imoutaharik: Also, I read that you shouldn't have `<input spellcheck></input>`  if it's undefined so I set the attribute only if we have true or false passed in.

jonasbuechel (approved):

PR #647, opened 2022-03-03
fix(Breadcrumbs): Move Hook Calls from Loops to Child Components

### LINKED ISSUE #645: Breadcrumbs: Rendered fewer hooks than expected. This may be caused by an accidental early return statement.
When the `items` prop of the Breadcrumb component changes its value, React crashes with the following error:

`Rendered fewer hooks than expected. This may be caused by an accidental early return statement.`

This is most likely because `useRef` is used inside the rendering which will result in different amount/order of hooks. 

**Reproduction**

The reproduction shows:

- Breadcrumbs with initial state
- useEffect changes the state → error

https://codesandbox.io/s/breadcrumbs-tabindex-frontify-arcade-2-3-0-forked-c0vklv?file=/src/App.js

deniaz:
Fixes #645

### DIFF
--- a/src/components/Breadcrumbs/Breadcrumbs.spec.tsx
+++ b/src/components/Breadcrumbs/Breadcrumbs.spec.tsx
@@ -1,7 +1,8 @@
 /* (c) Copyright Frontify Ltd., all rights reserved. */
 
 import { mount } from "@cypress/react";
-import React from "react";
+import React, { FC, useState } from "react";
+import { BreadcrumbsProps } from ".";
 import { Breadcrumbs } from "./Breadcrumbs";
 
 beforeEach("Getting the seperator", () => {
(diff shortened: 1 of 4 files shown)


src/components/Breadcrumbs/Breadcrumbs.spec.tsx
```diff
@@ -15,6 +16,24 @@ const BREADCRUMB_ITEMS = [
...
+                onClick={() => {
+                    setItems((items) => [...items, { label: "Some fourth label", link: "/some-fourth-link" }]);
+                }}
+            >
+                Add Block
+            </button>
+        </div>
+    );
+};
+
```
jonasbuechel, line 36: Cool to see a regression test here 🚀

jonasbuechel (approved):

PR #649, opened 2022-03-03
feat: migrate OrderableList to use react-dnd

Kenny806:


src/components/OrderableListV2/CollectionItem.tsx
```diff
@@ -0,0 +1,39 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+import { useFocusRing } from "@react-aria/focus";
+import { useId } from "@react-aria/utils";
+import { FOCUS_STYLE } from "@utilities/focusStyle";
+import React from "react";
```
GoranRibic, line 6: no need for React import

src/components/OrderableListV2/DropZone.tsx
```diff
@@ -0,0 +1,60 @@
...
+            canDrop: monitor.canDrop(),
+        }),
+    });
+
+    const isActive = isOver && canDrop;
+
+    return (
+        <div
+            role="row"
+            aria-hidden={isActive ? "false" : "true"}
```
GoranRibic, line 49: instead of ternary operator use isActive.toString()

src/components/OrderableListV2/OrderableListV2.spec.tsx
```diff
@@ -0,0 +1,217 @@
...
+                .trigger("dragenter", { ...getCoordinatesByIndex(dropIndex, "After"), dataTransfer })
+                .trigger("dragover", { ...getCoordinatesByIndex(dropIndex, "After"), dataTransfer });
+            cy.get(INSERTION_INDICATOR).should("exist").and("have.length", 1);
+            cy.get(LIST_ID).trigger("drop", { dataTransfer, force: true });
+            cy.wrap($el).trigger("dragend", { dataTransfer, force: true });
+            cy.wrap($el).should("have.text", ItemDragState.Idle);
+        });
+    });
+
+    it("Has keyboard accessibility", () => {
```
GoranRibic, line 135: Does it make sense to split this into multiple tests?

src/components/OrderableListV2/OrderableListV2.stories.tsx
```diff
@@ -0,0 +1,164 @@
...
+                dragStoryStyles[componentDragState],
+                isFocusVisible && "tw-bg-violet-20",
+            ])}
+        >
+            <div className="tw-text-xs tw-text-black-60">Position: Foo</div>
+            <hr className="tw-mt-2 tw-mb-2 tw-border-black-20 tw-bg-black-20" />
+            <div>{textContent}</div>
+            <hr className="tw-mt-3 tw-mb-2 tw-border-black-20 tw-bg-black-20" />
+            <div className="tw-flex tw-justify-between tw-text-s">
+                <span className="tw-font-medium">{isFocusVisible && "Im in keyboard focus"}</span>
```
GoranRibic, line 131: isFocusVisible should be beside span element not inside it, or i'm missing something?

src/components/OrderableListV2/DropZone.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+type DropZoneData<T> = {
+    targetItem: OrderableListItem<T>;
+    position: DropZonePosition;
+};
+
+export type DropZoneProps<T> = {
+    data: DropZoneData<T>;
+    onDrop?: OnDropCallback<T>;
+    children?: JSX.Element;
```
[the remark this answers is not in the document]
Kenny806: Because in Tree, we need to wrap the element in the dropzone to enable drag&dropping stuff on/into an element. For dropzones before/after an item, the dropzone produces a standalone element

src/components/OrderableListV2/DropZone.tsx
```diff
@@ -0,0 +1,60 @@
+import React from "react";
```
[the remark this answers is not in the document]
Kenny806: Snímek obrazovky 2022-03-04 v 14 59 06 I think it is needed

src/components/OrderableListV2/OrderableListV2.spec.tsx
Kenny806: Maybe. But this will be deleted for now, as we don't have keyboard control in the current implementation. There will be a separate ticket for that.

src/components/OrderableListV2/OrderableListV2.stories.tsx
Kenny806: Probably right (not sure as this is copy-pasted from the original OrderableList). Anyway, I moved it.

src/components/OrderableListV2/DropZone.tsx
Kenny806: It's actually even simpler, as the `aria-hidden` attribute accepts also boolean. Done :+1:

Kenny806: Oh, and I actually need the opposite value, so it's `aria-hidden={!isActive}`

GoranRibic (approved):

PR #655, opened 2022-03-07
feat(TextInput): add copyable prop

SamCreasey:
Add a new property to the TextInput component to allow for copying the value of the input. 
Adds a new "useCopy" hook

### DIFF
--- a/src/components/TextInput/TextInput.stories.tsx
+++ b/src/components/TextInput/TextInput.stories.tsx
@@ -19,7 +19,7 @@ export default {
         type: TextInputType.Text,
         validation: Validation.Default,
         spellcheck: true,
-        copyable: true,
+        copyable: false,
     },
     argTypes: {
         validation: {
--- a/src/components/TextInput/TextInput.stories.tsx
+++ b/src/components/TextInput/TextInput.stories.tsx
@@ -19,6 +19,7 @@ export default {
         type: TextInputType.Text,
         validation: Validation.Default,
         spellcheck: true,
+        copyable: true,
     },
     argTypes: {
         validation: {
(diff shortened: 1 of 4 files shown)


src/components/TextInput/TextInput.stories.tsx
```diff
@@ -19,6 +19,7 @@ export default {
         type: TextInputType.Text,
         validation: Validation.Default,
         spellcheck: true,
+        copyable: true,
```
hochreutenerl, line 22: nit: I would suggest false as a default

src/components/TextInput/TextInput.tsx
```diff
@@ -89,6 +94,7 @@ export const TextInput: FC<TextInputProps> = ({
     autocomplete = false,
     dotted = false,
     value = "",
+    copyable,
```
hochreutenerl, line 97: Probably it makes sense to define false as a default value

src/components/TextInput/TextInput.tsx
```diff
@@ -219,6 +229,32 @@ export const TextInput: FC<TextInputProps> = ({
...
+            {copyable && (
+                <button
+                    className={merge([
+                        "tw-flex tw-items-center tw-justify-center tw-transition-colors tw-rounded",
+                        disabled ? "tw-cursor-default tw-text-black-40" : "tw-text-black-60 hover:tw-text-black-100",
+                        copyButtonIsFocusVisible && FOCUS_STYLE,
+                    ])}
+                    onClick={() => copy(value)}
+                    data-test-id="copy-icon"
+                    title="Copy input value"
```
hochreutenerl, line 241: Just for confirmation: Is this the final title?

SamCreasey: I changed it to "copy input text", but if you have any other ideas I would appreciate it :)

src/components/TextInput/TextInput.tsx
SamCreasey: done

src/components/TextInput/TextInput.stories.tsx
SamCreasey: done

hochreutenerl (approved):
Nice work 💪

PR #658, opened 2022-03-08
fix(datepicker): Add datepicker to the bundle

jonasbuechel:
How to test: 
Run `npm build`, `dist/src/components/datepicker/`should now be available


urietberger (approved):
Let's go! 🚀

PR #660, opened 2022-03-08
feat(Textarea)!: add value prop and remove children

leko91:
Removing the `children` from the `<Textarea>` component and adding `value` prop instead, as discussed here: 
https://frontify.slack.com/archives/C02PSJTPA3D/p1646391887846429

### DIFF
--- a/src/components/Textarea/Textarea.spec.tsx
+++ b/src/components/Textarea/Textarea.spec.tsx
@@ -12,7 +12,7 @@ const INPUT_TEXT = "some text";
 
 describe("Textarea component", () => {
     it("renders", () => {
-        mount(<Textarea>{DEFAULT_TEXT}</Textarea>);
+        mount(<Textarea></Textarea>);
         cy.get("textarea").as("textarea");
         cy.get("@textarea").should("not.have.attr", "value");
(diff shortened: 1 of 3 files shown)


jonasbuechel (changes requested):
Currently, if you add a value via the storybook controls  (or URL see prefilled version [here](https://deploy-preview-660--arcade-components.netlify.app/?path=/story/components-textarea--textarea&args=validation:!undefined;value:my+value)) we're not able to write into the textarea anymore. I think this might be because we're using controlled value and didn't add a state to it.
It's working for TextInput and as I see we connect value with a state [there](https://github.com/Frontify/arcade/blob/16fee7c91060fd26ee0842cd6292155534ad317e/src/components/TextInput/TextInput.stories.tsx#L49). 

More about the topic
https://reactjs.org/docs/forms.html#the-textarea-tag
https://react-spectrum.adobe.com/react-spectrum/TextArea.html
https://reactjs.org/docs/forms.html#controlled-components
https://reactjs.org/docs/uncontrolled-components.html

jonasbuechel:
Changed the PR's title because this change is breaking as we change the props

jonasbuechel (approved):

PR #661, opened 2022-03-09
feat(RichTextEditor): add support for multiple editors

Walkeezy:
This PR introduces a `id` prop to the RichTextEditor, so that multiple editors can be rendered.


src/components/RichTextEditor/RichTextEditor.tsx
```diff
@@ -10,6 +10,7 @@ import { TextStyleType } from "./utils/getTextStyles";
 import { EMPTY_VALUE, parseRawValue } from "./utils/parseRawValue";
 
 export type RichTextEditorProps = {
+    id?: string;
```
SamuelAlev, line 13: We can use the Id generator as a fallback

src/components/RichTextEditor/RichTextEditor.spec.tsx
```diff
@@ -157,27 +158,35 @@ describe("RichTextEditor Component", () => {
...
+        cy.get(TEXTSTYLE_OPTION).eq(5).click();
+        cy.get("[contenteditable=true]").should("include.html", textStyleClassnames[TextStyles.ELEMENT_CUSTOM2]);
+    });
 
-        cy.get(TEXTSTYLE_DROPDOWN_TRIGGER).click({ force: true });
-        cy.get(TEXTSTYLE_OPTION).eq(3).click();
-        cy.get("[contenteditable=true]").should("include.html", "<h4");
+    it("renders multiple editors", () => {
+        mount(
+            <>
```
SamCreasey, line 183: It might be worth testing that there is no breaking when using multiple editors, for example testing it by typing it into both and ensuring that the values remain different.

src/components/RichTextEditor/RichTextEditor.tsx
Walkeezy: Added

src/components/RichTextEditor/RichTextEditor.spec.tsx
Walkeezy: Added

src/components/RichTextEditor/RichTextEditor.tsx
```diff
@@ -3,12 +3,14 @@
...
 import React, { FC, useCallback, useEffect, useState } from "react";
 import { EditableProps } from "slate-react/dist/components/editable";
-import { debounce } from "../..";
+import { debounce, generateRandomId } from "../..";
 import { Toolbar } from "./Toolbar";
 import { getEditorConfig } from "./utils/getEditorConfig";
 import { TextStyleType } from "./utils/getTextStyles";
 import { EMPTY_VALUE, parseRawValue } from "./utils/parseRawValue";
 
+const defaultEditorId = generateRandomId();
```
SamuelAlev, line 12: This is a sideeffect (not treeshakable), could you move this inside the component and use `useMemoizedId`?

Walkeezy: Changed to `useMemoizedId`

src/components/RichTextEditor/RichTextEditor.tsx
```diff
@@ -52,26 +57,27 @@ const RichTextEditorComponent: FC<RichTextEditorProps> = ({
...
         <div data-test-id="rich-text-editor" className="tw-relative tw-w-full">
             <Plate
-                initialValue={parseRawValue(initialValue)}
-                onChange={onChange}
+                id={editorId}
+                editor={editor}
+                initialValue={parseRawValue(editor, initialValue)}
+                onChange={(value) => {
+                    onChange(value);
+                    localValue.current = value;
```
SamCreasey, line 74: might be worth moving this line into the useCallback so that the function is not re-rendered each time

SamCreasey (approved):

SamuelAlev (approved):

PR #662, opened 2022-03-09
fix: export validation file

SamCreasey:
This fixes build errors in clarify with the latest version of arcade due to Validation not being exported


hochreutenerl (approved):
Thanks for fixing 👍

PR #666, opened 2022-03-10
feat(icon): add buildings icon

thisguyscode:


thisguyscode:
@SamCreasey and @SamuelAlev - Added you as reviewers as you're the ones using this icon anyway, so kill two birds with one stone (notify you and get it reviewed ;) )

thisguyscode:
Hmmm, I just noticed that various Caret icons got generated too. I know they're used in the DatePicker which is already merged, so not sure what's going on there. Do you?

SamuelAlev:
They just never ran `npm run build` but the CI does it so it is still bundled

thisguyscode:
> They never ran `npm run build` but the CI does it so it is still bundled

@SamuelAlev  So it's okay or should I also not run build?

SamuelAlev (approved):

PR #673, opened 2022-03-16
fix(Dropdown): Prevent keyboard interactions when disabled

deniaz:


SamCreasey:
It's still possible to use some keyboard interactions. For example, even though there is no focus ring you can still tab onto the component and change the selection with the left and right arrow keys. Ideally Tabbing onto the component would be completely disabled

PR #674, opened 2022-03-16
feat(accordion): extract header component

SamCreasey:
- Non-breaking update to the Accordion Component to separate the header from the AccordionItem itself and allow the user to pass in a "headerComponent" prop for custom styling.
- AccordionHeader is based on the FieldsetHeader, but allows us to deprecate easier when we need to.
- Refactor the Accordion Header Caret to allow users to easily import into own custom headers.
- Added a "divider" prop to the Accordion and the AccordionItem
- Added a "border" prop to the Accordion
- Added Collapsible Wrap Utility Component With Spec File & Storybook and refactored Accordion to accommodate for the new component


src/components/CollapsibleWrap/CollapsibleWrap.stories.tsx
```diff
@@ -0,0 +1,60 @@
...
+import { Heading } from "@typography/Heading";
+import { Card } from "@components/Card";
+
+// eslint-disable-next-line import/no-default-export
+export default {
+    title: "Components/Collapsible Wrap",
+    component: CollapsibleWrapComponent,
+} as Meta<CollapsibleWrapProps>;
+
+export const CollapsibleWrap: Story<CollapsibleWrapProps> = () => {
```
jonasbuechel, line 18: Please add the [storybook controls](https://storybook.js.org/docs/react/essentials/controls) so we have props documentation for free and a tool to play around :)

src/components/Accordion/Accordion.stories.tsx
```diff
@@ -110,6 +113,206 @@ export const WithDifferentAccordionItems: Story<AccordionProps> = () => {
...
+        {children}
+        <AccordionHeaderIcon isOpen={isOpen} type={FieldsetHeaderType.Accordion} size={AccordionHeaderIconSize.Small} />
+    </div>
+);
+
+const Code: FC = ({ children }) => (
+    <code className="tw-bg-black-5 tw-rounded tw-px-2 tw-text-box-negative-strong tw-text-s">{children}</code>
+);
+
+const PropsTable: FC<{ rows: [ReactNode, ReactNode, ReactNode][] }> = ({ rows }) => (
```
jonasbuechel, line 132: Please add the [storybook controls](https://storybook.js.org/docs/react/essentials/controls) so we have props documentation for free and a tool to play around :)

src/components/CollapsibleWrap/CollapsibleWrap.spec.tsx
```diff
@@ -0,0 +1,49 @@
...
+            </CollapsibleWrapComponent>
+            <Button onClick={() => setIsOpen((open) => !open)}>Click To Toggle Open</Button>
+        </>
+    );
+};
+
+describe("CollapsibleWrap", () => {
+    it("renders without crashing", () => {
+        mount(<CollapsibleWrap />);
+        cy.get("div").should("exist");
```
jonasbuechel, line 31: Please use a test-id

src/components/Accordion/Accordion.tsx
```diff
@@ -75,63 +57,43 @@ const AriaAccordionItem: FC<AriaAccordionItemProps> = ({ item, state, header, pa
...
-                        variants={{
-                            open: { height: "auto", overflow: "visible" },
-                            collapsed: { height: 0, overflow: "hidden" },
-                        }}
-                        transition={{ type: "tween" }}
-                        data-test-id="accordion-item-content"
+                        initial={{ opacity: 0 }}
+                        animate={{ opacity: 1 }}
+                        exit={{ opacity: 0 }}
+                        transition={{ duration: 0.3 }}
```
jonasbuechel, line 70: motion.div is part of collapsibleWrap, is it needed here again?

src/components/CollapsibleWrap/CollapsibleWrap.tsx
```diff
@@ -0,0 +1,29 @@
...
+        {isOpen && children && (
+            <motion.div
+                initial={"collapsed"}
+                animate={"open"}
+                exit={"collapsed"}
+                variants={{
+                    open: { height: "auto", overflow: "hidden" },
+                    collapsed: { height: 0, overflow: "hidden" },
+                }}
+                transition={{ type: "tween" }}
```
jonasbuechel, line 22: Screenshot 2022-03-23 at 11 17 28
Setting overflow hidden prevents dropdown from going out of an accordion item.

jonasbuechel (changes requested):

src/components/CollapsibleWrap/CollapsibleWrap.tsx
SamuelAlev: Intended behavior, the dropdown should be always on top, so the fix should be there (not only an issue in accordion)

src/components/Accordion/Accordion.tsx
SamCreasey: That is a good question. I have removed this and instead added the opacity to the collapsible wrap with the option to turn it off if required, as well as update storybook with this option

src/components/CollapsibleWrap/CollapsibleWrap.stories.tsx
SamCreasey: Added :)

src/components/CollapsibleWrap/CollapsibleWrap.tsx
SamCreasey: @jonasbuechel I could revert this, however a problem can occur in that while the accordion is expanding there will be a visual glitch where overflowing content (which is all content at the beginning since the height is 0) will briefly be displayed over the Accordion below until the Accordion reaches its full size

src/components/CollapsibleWrap/CollapsibleWrap.spec.tsx
SamCreasey: Added

src/components/Accordion/Accordion.stories.tsx
```diff
@@ -110,6 +113,206 @@ export const WithDifferentAccordionItems: Story<AccordionProps> = () => {
...
+    <table className="tw-table-fixed tw-border tw-border-black-10 tw-my-4">
+        <thead className="tw-bg-black-5">
+            <tr className="tw-p-3 tw-border-b tw-border-b-black-10">
```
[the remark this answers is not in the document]
SamCreasey: 👍

src/components/CollapsibleWrap/CollapsibleWrap.stories.tsx
jonasbuechel: Nice 🙌

src/components/Accordion/Accordion.stories.tsx
jonasbuechel: Nice, yes I get it, it's also not really able to handle components via props.

src/components/Accordion/AccordionHeader.stories.tsx
```diff
@@ -0,0 +1,42 @@
...
+                    IconIcons: "Icon",
+                    IconDefault: "Brand",
+                    IconCallout: "Callout",
+                },
+            },
+        },
+    },
+    args: {
+        isOpen: false,
+        disabled: false,
```
jonasbuechel, line 37:
```suggestion
        disabled: false,
        bold:true,
```

jonasbuechel (approved):
Nice improvements, I have one little contribution to the docs ;-)

PR #677, opened 2022-03-18
fix(tabs): Improve styling and add scroll into active tab

imoutaharik:


imoutaharik:
@LidijaStoilova I made some updates and the accessibility should be better now. :)

deniaz (approved):

PR #681, opened 2022-03-28
feat(icons): add Plus-circle icon

imoutaharik:
@thisguyscode I'm adding the icon you sent me but I noticed there is another Plus icon that looks the same but has a different path. 
There are three instances of the same icon with different svgs.
Screenshot 2022-03-28 at 15 28 55
Screenshot 2022-03-28 at 15 29 13

And the Plus icon also look blurry in Clarify.


thisguyscode:
And FYI the new 'Plus' doesn't have a circle. That must be another legacy thing we have to clean up later. 

New versions are Plus (plain addition symbol) and Plus Circle (addition symbol in a circle)

thisguyscode:
@imoutaharik 

Screenshot 2022-03-28 at 15 58 22

The 3 icons look identical to me (even if the paths are not identical), so I imagine the blurriness is in vertical alignment of the button's content.

thisguyscode:
@imoutaharik is this ready for approval? The icons look okay to me but I think you mentioned something about using the old one right?

jonasbuechel (approved):

PR #682, opened 2022-03-28
feat(dropdown): Add Alignment options

imoutaharik:
Adding alignment Left, Middle, Right.
+ Fixing Animation glitch

### DIFF
--- a/src/components/Dropdown/Dropdown.tsx
+++ b/src/components/Dropdown/Dropdown.tsx
@@ -30,6 +30,11 @@ export enum DropdownAlignment {
     End = "End",
 }
 
+export enum DropdownPosition {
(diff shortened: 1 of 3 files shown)


src/components/Dropdown/Dropdown.tsx
```diff
@@ -37,6 +54,8 @@ export type DropdownProps = {
     decorator?: ReactElement;
     autoResize?: boolean;
     validation?: Validation;
+    alignment?: DropdownAlignment;
+    position?: DropdownPosition;
```
jonasbuechel, line 58: @thisguyscode do you like to add this to the Figma component as well?

jonasbuechel (approved):

PR #683, opened 2022-03-29
fix(Flyout): Fix border-radius with fixed Header and Footer.

imoutaharik:
clickup task: #29nx6nc


src/components/Flyout/Overlay.tsx
```diff
@@ -50,7 +50,11 @@ const OverlayComponent: ForwardRefRenderFunction<HTMLDivElement, OverlayProps> =
                 {fixedHeader}
                 <div
                     ref={scrollRef}
-                    className="tw-flex tw-overflow-y-auto tw-overflow-x-hidden tw-flex-col tw-divide-y tw-divide tw-divide-black-10 tw-rounded tw-bg-white tw-text-black dark:tw-text-white dark:tw-bg-black-95"
+                    className={merge([
+                        "tw-flex tw-overflow-y-auto tw-overflow-x-hidden tw-flex-col tw-divide-y tw-divide tw-divide-black-10 tw-bg-white tw-text-black dark:tw-text-white dark:tw-bg-black-95",
+                        !fixedHeader && "tw-rounded-t",
+                        !fixedFooter && "tw-rounded-b",
```
jonasbuechel, line 56: We currently can't retest it at Arcade only. Can we add that case as a storybook example?

jonasbuechel (changes requested):

src/components/Flyout/Overlay.tsx
imoutaharik: I added a story, you can remove the header and footer in the controls but the border radius is very small. It's kind of hard to see but it's there.

jonasbuechel (approved):

PR #689, opened 2022-03-31
fix(rte): export empty value

SamCreasey:
Export the specified value of the empty rich text editor for use within clarify to implement the suggestion made here https://github.com/Frontify/clarify/pull/8165#discussion_r832395314


hochreutenerl (approved):
Looks good  👍

PR #693, opened 2022-04-01
fix(RichTextEditor): test for multiple editors

Walkeezy:
This PR changes the test for multiple editors to render the editor directly with the different values, instead of typing into it. Previously the test failed randomly and no other solution could be found.

Ticket: https://app.clickup.com/t/29vhrb8


SamCreasey (approved):
No failing test now 🥳

PR #694, opened 2022-04-01
fix(modal): remove fixed max height

SamCreasey:
This pr changes the modal max height to be dependent on the screen height, rather than what it was originally (fixed max height of 600px)


Walkeezy (approved):

thisguyscode:
@Walkeezy @SamCreasey - To me this looks broken. Am I missing something?

PR #712, opened 2022-04-07
fix: dropdown is hidden in Flyout overflow

imoutaharik:


src/components/Flyout/Flyout.stories.tsx
```diff
@@ -135,15 +137,35 @@ const WithButtonFlyoutTemplate: Story<FlyoutProps> = (args) => {
...
+                                { id: "2", title: "Item 2" },
+                                { id: "3", title: "Item 3" },
+                                { id: "4", title: "Item 4" },
+                                { id: "5", title: "Item 5" },
+                            ],
+                        },
+                    ]}
+                />
+                <p className="tw-pt-3">Field 2</p>
+                <DatePicker />
```
jonasbuechel: @imoutaharik I just added onChange state to have a working date picker here

src/components/Dropdown/Dropdown.tsx
```diff
@@ -141,12 +133,14 @@ export const Dropdown: FC<DropdownProps> = ({
...
     const showClear = !!activeItem && !!onClear;
 
-    const getDropdownBottomPosition = (dropDownSize: string) => {
-        return dropDownSize === DropdownSize.Small ? "tw-mb-2 tw-bottom-[34px]" : "tw-mb-2 tw-bottom-[60px]";
-    };
+    const leftPosition =
+        triggerRef.current && overlayRef.current
+            ? triggerRef.current?.getBoundingClientRect().right - overlayRef.current?.getBoundingClientRect().width
+            : "auto";
+    const bottomPosition = triggerRef.current ? triggerRef.current?.getBoundingClientRect().top - 8 : "auto";
```
jonasbuechel, line 140: Spacing between trigger and flyout is different for top and bottom position.
Please make sure to align with [the figma design](https://www.figma.com/file/pevuvey1joCsopTRr9f4OE/Fondue-React-Components?node-id=13839%3A254266)
Screenshot 2022-04-11 at 10 37 26
Screenshot 2022-04-11 at 10 37 19

src/components/Dropdown/Dropdown.tsx
```diff
@@ -141,12 +133,14 @@ export const Dropdown: FC<DropdownProps> = ({
...
 
     return (
-        <div className="tw-relative tw-w-full tw-font-sans tw-text-s">
+        <div className="tw-w-full tw-font-sans tw-text-s">
```
jonasbuechel, line 143: The validation loader circle is off since we changed positioning here.
Screenshot 2022-04-11 at 10 28 15

jonasbuechel (changes requested):
The overflow problem is solved 🥳 Let's now fix the misaligned loader and fine-tune the flyout position for the top position 👌

selinahasenboehler:
Hey @jonasbuechel and @imoutaharik 
Any progress on this one soon? (would solve some issues we have in our External Data Feature)

src/components/Dropdown/Dropdown.tsx
```diff
@@ -141,12 +133,14 @@ export const Dropdown: FC<DropdownProps> = ({
...
     const showClear = !!activeItem && !!onClear;
 
-    const getDropdownBottomPosition = (dropDownSize: string) => {
-        return dropDownSize === DropdownSize.Small ? "tw-mb-2 tw-bottom-[34px]" : "tw-mb-2 tw-bottom-[60px]";
-    };
+    const leftPosition =
+        triggerRef.current && overlayRef.current
+            ? triggerRef.current?.getBoundingClientRect().right - overlayRef.current?.getBoundingClientRect().width
+            : "auto";
+    const topPosition = triggerRef.current ? triggerRef.current?.getBoundingClientRect().top - 3 : "auto";
```
jonasbuechel, line 140: This does not work if we have Large dropdown triggers or if the dropdown is in a scrollable container. I think there's no way around using popper for that job.

src/components/Dropdown/Dropdown.tsx
```diff
@@ -211,9 +210,17 @@ export const Dropdown: FC<DropdownProps> = ({
...
-                </span>
+                <div className="tw-relative">
+                    <span
+                        style={{
+                            top: size === DropdownSize.Small ? "-47px" : "-71px",
+                            right: "-11px",
+                        }}
+                        className="tw-absolute tw-bg-white tw-rounded-full tw-p-[2px] tw-border tw-border-black-10"
+                    >
+                        <LoadingCircle size={LoadingCircleSize.ExtraSmall} />
```
jonasbuechel, line 221: The loading circle is now back at the expected position 👍

jonasbuechel (changes requested):

src/components/Dropdown/Dropdown.tsx
imoutaharik: Updated!

src/components/Dropdown/Dropdown.tsx
```diff
@@ -18,6 +18,8 @@ import React, { FC, ReactElement, useEffect, useRef } from "react";
 import { DEFAULT_DROPDOWN_MAX_HEIGHT, useDropdownAutoHeight } from "./useDropdownAutoHeight";
 import { Validation } from "@utilities/validation";
 import { LoadingCircle, LoadingCircleSize } from "@components/LoadingCircle";
+import { createPortal } from "react-dom";
+import { usePopper } from "react-popper";
```
jonasbuechel, line 22: By adding VariationPlacement we get access to the placement types.

```suggestion
import { usePopper } from "react-popper";
import { VariationPlacement } from "@popperjs/core";
```

src/components/Dropdown/Dropdown.tsx
```diff
@@ -132,15 +134,25 @@ export const Dropdown: FC<DropdownProps> = ({
...
-            ? triggerRef.current?.getBoundingClientRect().right - overlayRef.current?.getBoundingClientRect().width
-            : "auto";
-    const topPosition = triggerRef.current ? triggerRef.current?.getBoundingClientRect().top - 3 : "auto";
+    const dropdownRef = useRef<HTMLDivElement | null>(null);
+    const placement = `${position}-${alignment}`;
+
+    const popperInstance = usePopper(triggerRef?.current, dropdownRef.current, {
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        // @ts-ignore
+        placement: placement.toLowerCase(),
```
jonasbuechel, line 143: By using a map we could bring in the types for placement, what do you think?

```suggestion
    const placementMap: Record<string, VariationPlacement> = {
        ["Top-Start"]: "top-start",
        ["Top-End"]: "top-end",
        ["Bottom-Start"]: "bottom-start",
        ["Bottom-End"]: "bottom-end",
    };

    const popperInstance = usePopper(triggerRef?.current, dropdownRef.current, {
        placement: placementMap[`${position}-${alignment}`],
```

src/components/Dropdown/Dropdown.tsx
```diff
@@ -132,15 +134,25 @@ export const Dropdown: FC<DropdownProps> = ({
...
+        modifiers: [
```
jonasbuechel, line 144: Currently, it popper flips the flyout automatically. Should we support that when defining it by property as well? Or would it be easier to add auto placements?

jonasbuechel (changes requested):
Awesome, it works like a charm now 👍  I've one suggestion for types and a point to discuss about placements.

src/components/Dropdown/Dropdown.tsx
jonasbuechel: @thisguyscode what do you think?

imoutaharik: Right! I see we could use placement auto as the dropdown is supposed to only go up or down. 
`placement: 'auto',
  modifiers: [
    {
      name: 'flip',
      options: {
        allowedAutoPlacements: ['top-left', 'top-right', 'bottom-left', 'bottom-right'], 
      },
    },
  ],`

Then there might be no need to have alignment and position props.

jonasbuechel (approved):

PR #715, opened 2022-04-08
fix(ColorPicker): remove overflow hidden from picker

thisguyscode:
- Allows the 'handles' to appear even if they are right at the edge or corner of the color picker


Walkeezy (approved):
👍🏻

PR #716, opened 2022-04-08
feat(Modal): add width property

thisguyscode:
Perhaps not the best approach, but as the API (`width`) is unlikely to change, it might make sense to release this so @findmind can use it and fix the internals later (as there are other Modal internals that need fixing anyway).


jonasbuechel:
@thisguyscode I adjusted the story to support undefined and simplified it a bit 👌

imoutaharik (approved):

PR #726, opened 2022-04-13
BREAKING CHANGE: Refactor Tooltip component

imoutaharik:
*Simplify Tooltip component usage for developers
- Removed TooltipArrow
- Removed React-aria
- Popper moved inside the component
- Added positioning (position and alignment), flip, and arrow
- Added arrow styling
- Added timeout on hover
- Accessibility


src/components/Tooltip/Tooltip.tsx
```diff
@@ -56,113 +56,221 @@ const paddingsBottom = {
...
+            triggerRefElement.current.addEventListener("mouseleave", () => setTimeout(() => setIsOpen(false), 2000));
+            triggerRefElement.current.addEventListener("focus", () => setIsOpen(true));
+        }
+
+        return () => {
+            if (triggerRefElement) {
+                triggerRefElement.current.removeEventListener("click", () => setIsOpen(!isOpen));
+                triggerRefElement.current.removeEventListener("mouseover", () => setIsOpen(true));
+                triggerRefElement.current.removeEventListener("mouseleave", () =>
+                    setTimeout(() => setIsOpen(false), 2000),
```
lenardpalko, line 175: Maybe we could add a config for the timeout value as well, that could have a default value if not set explicitly.

src/components/Tooltip/Tooltip.tsx
```diff
@@ -56,113 +56,221 @@ const paddingsBottom = {
...
+    });
+
+    const arrowStyling = setArrowClasses();
+    const [isOpen, setIsOpen] = useState(false);
+
+    useEffect(() => {
+        if (triggerRefElement) {
+            triggerRefElement.current.addEventListener("click", () => setIsOpen(!isOpen));
+            triggerRefElement.current.addEventListener("mouseover", () => setIsOpen(true));
+            triggerRefElement.current.addEventListener("mouseleave", () => setTimeout(() => setIsOpen(false), 2000));
```
lenardpalko, line 166: We either need to clear the timeout if the mouse is on the tooltip, either add a new variable e.g. `mouseOverTooltip` that would be true while the user is over the tooltip and which conditions the display of the tooltip as well. So the condition to display the tooltip would be then: `isOpen && mouseOverTooltip`. Otherwise the tooltip closes even if the user is on over the tooltip with the mouse.

src/components/Tooltip/Tooltip.tsx
```diff
@@ -56,113 +56,221 @@ const paddingsBottom = {
...
-            <div
-                ref={tooltipContainerRef}
-                className="tw-popper-container tw-inline-block tw-max-w-[200px] tw-bg-black-100 dark:tw-bg-white tw-rounded-md tw-shadow-mid tw-text-white dark:tw-text-black-100 tw-z-20"
-                style={style}
-                data-test-id="tooltip"
-                {...popperAttributes}
-                {...tooltipProps}
-            >
-                {brightHeader && <BrightHeader headerStyle={brightHeader} />}
+        <div>
```
lenardpalko, line 183: Maybe we could directly use ReactPortal here to avoid any positioning issues with other elements on the page.

Kenny806: Agreed. Both me and Lenny currently have to use Portal to correctly position the Tooltip, would be great if this was built into the component itself.

imoutaharik: @lenardpalko @Kenny806 I just updated according to all comments. I agree it makes sense to use Portal within the component. I also fixed the timeout issue.

imoutaharik:
@SamuelAlev I added you as a reviewer as I made a lot of changes in the tooltip component that may affect your team. 
@thisguyscode I also added you so you're aware of all the changes, I actually got rid of the tooltipArrow component that was passed as a child and used the logic that was thought for the popover.

thisguyscode:
@imoutaharik A few comments about the hover delay:

- There seem to be multiple properties in Storybook to control this, and I can't immediately see the effect of either.
- The tooltip should never be destroyed when the mouse is over it, regardless of hover delay
- In Storybook the delay is really high. I believe the default behavior should be almost 0 delay. Allowing just enough time to mouseover for the interactive tooltips.

thisguyscode:
@imoutaharik General Storybook:

- Could you add a story with multiple tooltips please? Would be cool to test how they behave when switching between
- I don't think you always need to add all properties to all stories. This way it's hard to find the relevant ones for the particular story.
- The `hoverDelay` / `hoverdelay` duplicate properties mentioned above.

src/components/Tooltip/Tooltip.tsx
Kenny806: @imoutaharik From my side it works fine now. I won't approve the PR since @thisguyscode still has some comments, but the way Tooltip is used is good for me and I have no further comments there.

lenardpalko:
@imoutaharik for my use case also seems to be all good. It simplifies a lot on how the tooltip should be used. Thanks for the great work 👍

src/components/Tooltip/Tooltip.stories.tsx
```diff
@@ -43,72 +43,128 @@ export default {
...
+export const MultipleTooltipsComponent: Story<TooltipProps> = (args: TooltipProps) => {
+    const firstTriggerRef = useRef<HTMLButtonElement>(null);
+    const secondTriggerRef = useRef<HTMLButtonElement>(null);
+    return (
+        <div className="tw-w-screen tw-h-screen tw-flex tw-justify-center tw-items-center">
+            <p className="tw-mr-2">Label</p>
+            <button ref={firstTriggerRef} className="tw-mr-1">
+                <IconAttentionFilled size={IconSize.Size16} />
+            </button>
+            <Tooltip {...args} triggerRefElement={firstTriggerRef} withArrow={true} />
```
SamCreasey, line 93: WithArrow Storybook control is overwritten by this prop and it feels a bit weird that it toggles one and not the other

SamCreasey:
If you hover over the tooltip itself it disappears, which is not great for when you have a link inside the tooltip

https://user-images.githubusercontent.com/93908356/166215714-b5b0d6f2-af92-4f38-84aa-7b0e8ea236f6.mov

SamCreasey:
> If you hover over the tooltip itself it disappears, which is not great for when you have a link inside the tooltip
> 
>  Screen.Recording.2022-05-02.at.11.44.54.mov

Also it is not possible to tab onto a link inside the tooltip

SamCreasey:
Screenshot 2022-05-02 at 11 48 44
I think the focus style of the element when using keyboard navigation might need to be updated to match the exisiting focus styles of other components

src/components/Tooltip/Tooltip.tsx
```diff
@@ -56,113 +58,236 @@ const paddingsBottom = {
...
+                name: "flip",
+                enabled: flip,
+            },
+        ],
+    });
+
+    const arrowStyling = setArrowClasses();
+    const [isOpen, setIsOpen] = useState(false);
+
+    let timeout: ReturnType<typeof setTimeout> | null = null;
```
SamCreasey, line 165: I wonder if timeout should be a ref object so that it persists between renders 🤔

src/components/Tooltip/Tooltip.tsx
```diff
@@ -4,31 +4,31 @@ import { Button, ButtonSize, ButtonStyle } from "@components/Button/Button";
...
 
 export type TooltipButton = {
     label: string;
     action: () => void;
 };
 
-type PopperAttributes = { [key: string]: string };
-
 export type TooltipProps = {
+    triggerRefElement?: MutableRefObject<any>;
```
SamCreasey, line 31: Might consider changing any to HTMLElement, wdyt?

src/components/Tooltip/Tooltip.tsx
```diff
@@ -56,113 +58,236 @@ const paddingsBottom = {
...
+            default:
+                return "before:tw-border-b-0 before:tw-border-r-0 tw-top-[-6px] before:tw-bg-black-100 before:dark:tw-bg-white";
+        }
+    };
+
+    const tooltipOffset = withArrow ? 10 : 5;
+    const popperInstance = usePopper(triggerRefElement?.current, tooltipContainerRef.current, {
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        // @ts-ignore
+        placement: placement,
```
SamCreasey, line 141: This can be removed by casting the type of the placement to Placement

imoutaharik: Fixed using variation Placement

src/components/Tooltip/Tooltip.tsx
```diff
@@ -56,113 +59,268 @@ const paddingsBottom = {
...
+    ["top-End"]: "top-end",
+    ["bottom-Start"]: "bottom-start",
+    ["bottom-End"]: "bottom-end",
+    ["left-Start"]: "left-start",
+    ["left-End"]: "left-end",
+    ["right-Start"]: "right-start",
+    ["right-End"]: "right-end",
+};
+
+export const Tooltip: FC<TooltipProps> = ({
```
SamuelAlev, line 86: [`FC` is not allowed anymore](https://frontify.slack.com/archives/C01JE9QENCF/p1650544457958619)

src/components/Tooltip/Tooltip.tsx
```diff
@@ -56,113 +59,268 @@ const paddingsBottom = {
...
+    withArrow,
+    flip = true,
+    triggerRefElement,
+    hoverDelay = 200,
+}) => {
     const linkRef = useRef<HTMLAnchorElement | null>(null);
     const { linkProps } = useLink({}, linkRef);
-
     const { isFocusVisible, focusProps } = useFocusRing();
+    const hasLargePaddingTop = linkUrl || brightHeader || buttons || heading || headingIcon;
```
SamuelAlev, line 106: `useMemo` so it doesn't get recomputed for each render in case any other props changes

src/components/Tooltip/Tooltip.tsx
```diff
@@ -56,113 +59,268 @@ const paddingsBottom = {
...
+                ]);
+            case currentPlacement.toString().includes(TooltipPosition.Left):
+                return merge([
+                    "before:tw-border-b-0 before:tw-border-l-0 tw-right-[-6px]",
+                    brightHeader && alignment === TooltipAlignment.Start
+                        ? brightHeaderArrowBackgroundColors[brightHeader]
+                        : "before:tw-bg-black-100 before:dark:tw-bg-white",
+                ]);
+            default:
+                return "before:tw-border-b-0 before:tw-border-r-0 tw-top-[-6px] before:tw-bg-black-100 before:dark:tw-bg-white";
```
SamuelAlev, line 139: Can we make it as a `const` outside the component?
If not, we can at least wrap it in a `useCallback`

src/components/Tooltip/Tooltip.tsx
```diff
@@ -56,113 +59,268 @@ const paddingsBottom = {
...
+    const handleShowTooltipOnHover = () => {
+        if (timeoutRef.current) {
+            clearTimeout(timeoutRef.current);
+        }
+        setIsOpen(true);
+    };
+
+    const handleHideTooltipOnHover = () => {
+        timeoutRef.current = setTimeout(() => setIsOpen(false), hoverDelay);
+    };
```
SamuelAlev, line 179: `useCallback`

src/components/Tooltip/Tooltip.tsx
```diff
@@ -56,113 +59,268 @@ const paddingsBottom = {
...
+
+    const checkIfHovered = (e: Event) => {
```
SamuelAlev, line 181:
```suggestion
    const checkIfHovered = (event: Event) => {
```

src/components/Tooltip/Tooltip.tsx
```diff
@@ -56,113 +59,268 @@ const paddingsBottom = {
...
+        if (hoveredElement.includes(tooltipContainerRef?.current)) {
+            handleShowTooltipOnHover();
+        }
+    };
+
+    const hasInteractiveElements = !!(buttons?.length || linkUrl?.length);
+    const firstButtonRef = useRef<HTMLButtonElement | null>(null);
+    const lastButtonRef = useRef<HTMLButtonElement | null>(null);
+
+    useEffect(() => {
```
SamuelAlev, line 198: Better to add the events as props of `TooltipIcon` and use the `on...` notation in the component instead of vanilla events and ref

SamuelAlev (changes requested):

src/components/Tooltip/Tooltip.tsx
imoutaharik: True! And I voted to keep it 😁 . I removed it. Thanks for all the feedback. I took into account all your comments and I just simplified everything again.

src/components/Tooltip/Tooltip.tsx
```diff
@@ -56,113 +59,263 @@ const paddingsBottom = {
...
-                            rel="noopener noreferrer"
-                            className={merge([
-                                "tw-text-xs tw-text-black-40 dark:tw-text-black-80 tw-underline tw-mt-1",
-                                isFocusVisible && FOCUS_STYLE,
-                            ])}
+                {triggerElement &&
+                    cloneElement(triggerElement, {
+                        ref: triggerRefElement,
+                    })}
+            </div>
```
SamCreasey, line 213: Wrapping the trigger element in a div brings about the same issues that we had with the FlyoutTrigger, where you might have a component that directly relies on it's parent for styling and cannot anymore. I would maybe suggest a render function.
`const triggerProps = {onMouseOver, onMouseLeave, onFocus, onBlur}`
```
<>
  {triggerElement(triggerProps, triggerRef)}
</>
```
wdyt?

src/components/Tooltip/Tooltip.tsx
```diff
@@ -56,113 +59,263 @@ const paddingsBottom = {
...
+const placementMap: Record<string, VariationPlacement> = {
+    ["top-Start"]: "top-start",
+    ["top-End"]: "top-end",
+    ["bottom-Start"]: "bottom-start",
+    ["bottom-End"]: "bottom-end",
+    ["left-Start"]: "left-start",
+    ["left-End"]: "left-end",
+    ["right-Start"]: "right-start",
+    ["right-End"]: "right-end",
+};
```
SamuelAlev, line 84:
```suggestion
const placementMap: Record<`${TooltipPosition}-${TooltipAlignment}`, VariationPlacement> = {
    ["top-Start"]: "top-start",
    ["top-End"]: "top-end",
    ["bottom-Start"]: "bottom-start",
    ["bottom-End"]: "bottom-end",
    ["left-Start"]: "left-start",
    ["left-End"]: "left-end",
    ["right-Start"]: "right-start",
    ["right-End"]: "right-end",
};
```
And seems we have different casing

src/components/Tooltip/Tooltip.tsx
```diff
@@ -72,36 +72,40 @@ export enum TooltipAlignment {
...
-    ["top-Start"]: "top-start",
-    ["top-End"]: "top-end",
-    ["bottom-Start"]: "bottom-start",
-    ["bottom-End"]: "bottom-end",
-    ["left-Start"]: "left-start",
-    ["left-End"]: "left-end",
-    ["right-Start"]: "right-start",
-    ["right-End"]: "right-end",
+const placementMap: Record<`${TooltipPosition}-${TooltipAlignment}`, Placement> = {
+    ["Top-Start"]: "top-start",
```
SamuelAlev, line 76: Can be written "[`${TooltipPosition.Top}-${TooltipAlignment.Start}`]"

src/components/Tooltip/Tooltip.tsx
```diff
@@ -56,113 +59,267 @@ const paddingsBottom = {
...
+    ["Right-End"]: "right-end",
+    ["Top-Middle"]: "top",
+    ["Right-Middle"]: "right",
+    ["Bottom-Middle"]: "bottom",
+    ["Left-Middle"]: "left",
+};
+
+const getArrowClasses = (currentPlacement: string, brightHeader: BrightHeaderStyle | undefined, alignment: string) => {
+    switch (true) {
+        case currentPlacement.toString().includes("top"):
```
SamuelAlev, line 92: Isn't it 
```suggestion
        case currentPlacement.toString().includes(TooltipPosition.Top):
```

imoutaharik: It's TooltipPosition.Top.toLowerCase(), Do we want it like this?

src/components/Tooltip/Tooltip.tsx
imoutaharik: Typescript didn't like that at all.

src/components/Tooltip/Tooltip.tsx
SamuelAlev: I read the code a bit more, now I see why 👍

src/components/Tooltip/Tooltip.tsx
imoutaharik: Will create a task for it.

PR #736, opened 2022-04-21
feat: Multiselect Tree component

selimovicz:


selimovicz:
> @selimovicz - Modifier key seems incorrect to me. Should be CTRL/CMD not SHIFT.
> 
> SHIFT usually allows for range selection in lists.

Yes, you're right. I talked with Simone and we both agreed that CMD/CTRL is more suitable in this situation.

Kenny806 (approved):

PR #739, opened 2022-04-25
chore: update storybook-builder-vite

Walkeezy:
Issue is fixed: https://github.com/storybookjs/builder-vite/pull/346
Flyout in storybook should work as intended with this version of the package.


SamuelAlev (approved):
LGTM 😄

PR #746, opened 2022-04-29
feat: Flyout min height

mikeyrayvon:
This adds a `minHeight` style prop for the Flyout content via a `contentMinHeight` component prop
[Loom demo](https://www.loom.com/share/df05651b51e9412fa2a846de835cbaeb)

### DIFF
--- a/src/components/Flyout/Flyout.tsx
+++ b/src/components/Flyout/Flyout.tsx
@@ -46,6 +46,7 @@ export type FlyoutProps = PropsWithChildren<{
     onOpenChange: (isOpen: boolean) => void;
     fixedHeader?: ReactNode;
     fixedFooter?: ReactNode;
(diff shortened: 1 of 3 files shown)


src/components/Flyout/Flyout.tsx
```diff
@@ -67,6 +68,7 @@ export const Flyout: FC<FlyoutProps> = ({
     fitContent = false,
     fixedHeader,
     fixedFooter,
+    contentMinHeight = "0px",
```
imoutaharik, line 71:
```suggestion
    contentMinHeight,
```

imoutaharik (commented):
Looks good to me! Thanks a lot for the support 😄 I added a suggestion to set it as number instead of string.

imoutaharik (approved):

PR #752, opened 2022-05-02
fix(AssetInput): flyout width is calculated incorrectly

fulopdaniel:
On component mount the width is sometimes 0  when the parent container has `display:none` applied. (e.g. in Tabs, if the the parent tab container is not active).

Fix: applied a resize observer to keep the width calculation always up-to-date.

### DIFF
--- a/src/components/AssetInput/SingleAsset/SelectedAsset.tsx
+++ b/src/components/AssetInput/SingleAsset/SelectedAsset.tsx
@@ -39,13 +39,19 @@ export const SelectedAsset: FC<Required<SelectedAssetProps>> = ({ asset, size, a
 
     const [flyoutWidth, setFlyoutWidth] = useState(0);
 
-    const calcFlyoutWidth = () => {
-        setFlyoutWidth(buttonRef.current?.getBoundingClientRect().width || 0);
-    };
-
     useEffect(() => {
-        calcFlyoutWidth();
-        window.addEventListener("resize", calcFlyoutWidth, false);
+        const calculateFlyoutWidth = () => {
+            const calculatedWidth = buttonRef.current?.getBoundingClientRect().width ?? 0;
+            setTimeout(() => setFlyoutWidth(calculatedWidth), 0);
+        };
+        const resizeObserver = new ResizeObserver(calculateFlyoutWidth);
+        if (buttonRef.current) {
+            resizeObserver.observe(buttonRef.current);
+        }
+
+        return () => {
+            resizeObserver.disconnect();
+        };
     }, []);
 
     return (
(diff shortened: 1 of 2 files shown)


src/components/AssetInput/SingleAsset/SelectedAsset.tsx
```diff
@@ -39,13 +39,17 @@ export const SelectedAsset: FC<Required<SelectedAssetProps>> = ({ asset, size, a
...
-    const calcFlyoutWidth = () => {
-        setFlyoutWidth(buttonRef.current?.getBoundingClientRect().width || 0);
-    };
-
     useEffect(() => {
-        calcFlyoutWidth();
-        window.addEventListener("resize", calcFlyoutWidth, false);
+        const calcFlyoutWidth = () => {
+            setFlyoutWidth(buttonRef.current?.getBoundingClientRect().width || 0);
+        };
```
SamuelAlev, line 45:
```suggestion
        const calculateFlyoutWidth = () => {
            setFlyoutWidth(buttonRef.current?.getBoundingClientRect().width ?? 0);
        };
```
We have a preference for `??` over `||` in the codebase.
I would also rename the method to a more verbose one (we anyway minify everything)

src/components/AssetInput/SingleAsset/SelectedAsset.tsx
```diff
@@ -39,13 +39,17 @@ export const SelectedAsset: FC<Required<SelectedAssetProps>> = ({ asset, size, a
...
 
-    const calcFlyoutWidth = () => {
-        setFlyoutWidth(buttonRef.current?.getBoundingClientRect().width || 0);
-    };
-
     useEffect(() => {
-        calcFlyoutWidth();
-        window.addEventListener("resize", calcFlyoutWidth, false);
+        const calcFlyoutWidth = () => {
+            setFlyoutWidth(buttonRef.current?.getBoundingClientRect().width || 0);
```
SamCreasey, line 44: Might require the setFlyoutWidth to be wrapped in a setTimeout, see the useScrollWrapper hook for an example and explanation. If you try to run a cypress test in clarify that contains the flyout it may or may not fail without it, so worth checking out.

fulopdaniel: Thanks for the heads-up, for now the tests seem to run fine, but to make sure and avoid errors in the future I added this timeout anyway.

SamuelAlev:
Could you add some tests to avoid any regression?

fulopdaniel:
> Could you add some tests to avoid any regression?


Absolutely

SamuelAlev (approved):

PR #761, opened 2022-05-04
fix(ScrollWrapper): added side spacing

SamCreasey:
Screenshot 2022-05-04 at 11 00 15

Adds some side spacing to the Scroll Wrapper to prevent focus styles and sliders being clipped by the overflow-hidden class


imoutaharik:
@SamCreasey I think we can even add a bit of padding vertically as the action button is also cropped when focused.
Screenshot 2022-05-04 at 11 12 29

PR #766, opened 2022-05-04
feat(Tree): add badge icon to tree node item

leko91:
As discussed somewhat in the [Slack thread](https://frontify.slack.com/archives/C02PSJTPA3D/p1650372981299969).

I did the styling as it is in the [design for the Navigation Manager](https://www.figma.com/file/f37w3IPvu5wBpo0GakMNm0/Navigation?node-id=985%3A467784).

### DIFF
--- a/src/components/Tree/Node.tsx
+++ b/src/components/Tree/Node.tsx
@@ -140,7 +140,12 @@ export const Node = ({
                             <span className="tw-flex tw-items-center" data-test-id="node-link-name">
                                 {name}
                                 {badge && (
-                                    <div className="tw-flex tw-justify-center tw-items-center tw-w-8 tw-h-5 tw-ml-2 tw-bg-box-neutral tw-rounded-full">
+                                    <div
+                                        className={merge([
+                                            "tw-flex tw-justify-center tw-items-center tw-w-8 tw-h-5 tw-ml-2 tw-bg-box-neutral tw-rounded-full",
+                                            selected && "tw-bg-transparent",
+                                        ])}
+                                    >
                                         <span>{badge}</span>
                                     </div>
                                 )}
--- a/src/components/Tree/Node.tsx
+++ b/src/components/Tree/Node.tsx
@@ -55,9 +55,9 @@ export const Node = ({
     onDrop,
     treeName,
 }: NodeProps): ReactElement<NodeProps> => {
-    const { id, value, name, label, icon, nodes, actions } = node;
+    const { id, value, name, label, icon, nodes, actions, badge } = node;
     const [{ opacity }, drag] = useDrag({
-        item: { id, value, name, label, icon, nodes, actions },
+        item: { id, value, name, label, icon, nodes, actions, badge },
         collect: (monitor) => ({
             opacity: monitor.isDragging() ? 0.4 : 1,
         }),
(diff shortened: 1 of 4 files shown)


src/components/Tree/Tree.tsx
```diff
@@ -15,6 +15,7 @@ export interface TreeFlatListItem {
     label?: string;
     value?: string;
     actions?: React.ReactNode[];
+    badge?: ReactElement<IconProps>;
```
julianiff, line 18: Would we be able to restrict a bit more what we put into the Badge? Because its called Badge, i think it would make sense to restrict to Icons/Badge, and also what you would need to fulfill your requirements.

src/components/Tree/Node.tsx
```diff
@@ -137,7 +137,19 @@ export const Node = ({
...
+                            <span className="tw-flex tw-items-center" data-test-id="node-link-name">
+                                {name}
+                                {badge && (
+                                    <div
+                                        className={merge([
+                                            "tw-flex tw-justify-center tw-items-center tw-w-8 tw-h-5 tw-ml-2 tw-bg-box-neutral tw-rounded-full",
+                                            selected && "tw-bg-transparent",
+                                        ])}
+                                    >
+                                        <span>{badge}</span>
```
julianiff, line 149:
```suggestion
                                       {badge}
```

Do we need a span here? if not we could leave it out.

src/components/Tree/Node.tsx
```diff
@@ -137,7 +137,19 @@ export const Node = ({
...
+                                    </div>
+                                )}
```
julianiff, line 151: What do you think, would it make sense if we refactor this new addition to the Node into a separate function, smth like const insertBadge/addBadge that suggests that a Badge gets included.

src/components/Tree/utils/mocks.tsx
```diff
@@ -141,6 +141,7 @@ export const mockNodesFlat: DraggableItem<TreeFlatListItem>[] = [
         icon: <IconFolder size={IconSize.Size16} />,
         value: "https://weare.frontify.com/document/923#/test",
         sort: null,
+        badge: <IconSmileysPeople size={IconSize.Size16} />,
```
julianiff, line 144: Looove the smily :D 
We also have the Badge Component, maybe we could include also the Badge in another Node to also show this showcase.

julianiff (commented):
Looks great thank you for the PR 👍. I added some small comments from my side.

leko91:
@julianiff
I've updated the PR based on your comments. Also, I've added a bit more styling since Badge and our Icon differentiate a bit. Maybe that's something that the design team could fix, so we always use only the badge element, if they agree we could remove the Icon use case.

src/components/Tree/Node.tsx
```diff
@@ -85,6 +85,22 @@ export const Node = ({
         setShowNodes(!showNodes);
     };
 
+    const insertBadge = () => {
+        console.log(badge);
```
julianiff, line 89:
```suggestion
```

julianiff (approved):
Looks great. There is just a console.log to remove.

julianiff:
Ah if possible, could we add also a test that checks if the badge is rendered correctly? That could avoid later regression bugs if someone would change the badge.

leko91:
> Ah if possible, could we add also a test that checks if the badge is rendered correctly? That could avoid later regression bugs if someone would change the badge.

@julianiff 
Added the tests now 😄

julianiff (approved):

thisguyscode:
@julianiff I thought we would use a proper Badge component. Is there a limitation to the Badge that made that difficult/impossible?

PR #768, opened 2022-05-05
feat(Tree): create editable Tree node item

leko91:
This was discussed in this [Slack thread](https://frontify.slack.com/archives/C02PSJTPA3D/p1650997178045199?thread_ts=1650372981.299969&cid=C02PSJTPA3D).

@julianiff
The PR is not complete (test failing, logs, etc.), just want to sync with you first on the implementation, fix any big issues if any, and then complete everything so we skip double work.

### DIFF
--- a/src/components/Tree/Tree.tsx
+++ b/src/components/Tree/Tree.tsx
@@ -18,6 +18,8 @@ export interface TreeFlatListItem {
     actions?: React.ReactNode[];
     badge?: ReactElement<IconProps> | ReactElement<BadgeProps>;
     parentId: NullableString;
+    editable?: boolean;
+    onEditableSave?: (value: string) => void;
 }
 
 export type TreeProps = {
--- a/src/components/Tree/components/EditableNodeItem.tsx
+++ b/src/components/Tree/components/EditableNodeItem.tsx
@@ -9,7 +9,7 @@ export const EditableNodeItem = ({ name, onEditableSave }: EditableNodeItem) => 
     const [inputValue, setInputValue] = useState(name);
     const [showInput, setShowInput] = useState<boolean>(false);
 
-    const onKeyDown = (event: KeyboardEvent<HTMLInputElement>) => {
+    const handleKeyDown = (event: KeyboardEvent<HTMLInputElement>) => {
         if (event.key === "Enter") {
             setShowInput(false);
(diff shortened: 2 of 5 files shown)


GoranRibic (approved):

julianiff:
@leko91 Sorry for the late reply I missed the email. Looks great!

src/components/Tree/Tree.tsx
```diff
@@ -16,6 +16,8 @@ export interface TreeFlatListItem {
     value?: string;
     actions?: React.ReactNode[];
     parentId: NullableString;
+    editable?: boolean;
+    onEditableSave?: (value: string) => void;
 }
```
julianiff, line 21: We were discussing about the properties. Do you think it would be enough to just have the Callback property instead of both?

src/components/Tree/components/EditableNodeItem.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+export const EditableNodeItem = ({ name, onEditableSave }: EditableNodeItem) => {
+    const [inputValue, setInputValue] = useState(name);
+    const [showInput, setShowInput] = useState<boolean>(false);
+
+    const handleKeyDown = (event: KeyboardEvent<HTMLInputElement>) => {
+        if (event.key === "Enter") {
+            setShowInput(false);
+
+            onEditableSave && onEditableSave((event.target as HTMLInputElement).value);
```
julianiff, line 16: The interface defined the property not as optional, i think you can remove the check with the &&

julianiff (approved):
Thx @leko91 for the nice prototype.

src/components/Tree/Tree.tsx
leko91: It would be, I am also not a fan of having both properties. However, consistency wise, I think editable is a good prop to have.

Guess it's your call in the end 😄

julianiff: Then we leave for now if you see that it helps consistency

leko91:
@julianiff 
Hey Julian, I've updated the code a bit, added a close on blur, focus on opening the input, and written tests. Can you check it out again, please? 😃

julianiff (approved):

PR #778, opened 2022-05-10
style: remove restriction of height

SamuelAlev:
Remove extra height restriction which is not correct
Before:
Screenshot 2022-05-10 at 15 21 17
After:
Screenshot 2022-05-10 at 15 21 24

Add remove the double build from CI pipeline 🙈


SamuelAlev:
@imoutaharik Fixed

imoutaharik (approved):

PR #784, opened 2022-05-11
chore: inline peerDependencies

SamuelAlev:
Inline `peerDependencies`, `npm` starting version 7 support auto install of `peerDependencies` however other package manager may not, example `pnpm` that we use in [Frontify/guideline-blocks](https://github.com/Frontify/guideline-blocks) requires to manual installation of `peerDependencies`.

The way to fix it is to manually add the dependencies which is not optimal: https://github.com/Frontify/guideline-blocks/blob/main/.pnpmfile.cjs


package.json
```diff
@@ -139,11 +140,18 @@
...
         "react-datepicker": "^4.7.0",
         "react-dnd": "^15.1.1",
         "react-dnd-html5-backend": "^15.1.2",
+        "react-is": "^18.1.0",
         "react-popper": "^2.2.5",
+        "slate": "^0.78.0",
+        "slate-history": "^0.66.0",
+        "slate-hyperscript": "^0.77.0",
+        "slate-react": "^0.79.0",
+        "styled-components": "^5.3.5",
```
SamuelAlev, line 154: Storybook dependency

package.json
```diff
@@ -139,11 +140,18 @@
...
         "react-colorful": "^5.5.1",
         "react-datepicker": "^4.7.0",
         "react-dnd": "^15.1.1",
         "react-dnd-html5-backend": "^15.1.2",
+        "react-is": "^18.1.0",
         "react-popper": "^2.2.5",
+        "slate": "^0.78.0",
+        "slate-history": "^0.66.0",
+        "slate-hyperscript": "^0.77.0",
+        "slate-react": "^0.79.0",
```
SamuelAlev, line 153: plate dependency

package.json
```diff
@@ -139,11 +140,18 @@
         "@xstate/react": "^1.6.3",
         "date-fns": "^2.28.0",
         "framer-motion": "^6.2.9",
+        "immer": "^9.0.12",
```
SamuelAlev, line 143: xstate dependency

package.json
SamuelAlev, line 153: plate dependency

triggertoo (approved):
👍

PR #786, opened 2022-05-12
feat: add Modal.Body padding prop. move Modal padding into parts

mikeyrayvon:
- moves the `tw-p-14` padding class from the `<Modal>` and reproduces that padding with classes in the Modal Body, Header, and Footer.
- adds a `horizontalPadding` class to `Modal.Body` which defaults to `true`, to remove the horizontal padding class from the `Modal.Body` wrapper `div` when set to `false`
- adds a Story for demoing the new prop


thisguyscode (approved):
Looks good to me, thanks @mikeyrayvon!

PR #789, opened 2022-05-12
feat: Add formId prop and form attribute to Button

mikeyrayvon:
This adds an optional `formId` prop to `ButtonProps`, which is used in the `form` attribute on the `<Button>`. This associates the child `<button>` with a `<form id={formId} />` that it falls outside of, allowing a button to trigger `onSubmit` for a form even if it is outside the form in the DOM, such as when in a `<Flyout>` `fixedFooter`.

[via MDN:](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#attr-form)
>form 
>The `<form>` element to associate the button with (its form owner). The value of this attribute must be the id of a `<form>` in the same document. (If this attribute is not set, the `<button>` is associated with its ancestor `<form>` element, if any.)
> 
> This attribute lets you associate `<button>` elements to `<form>`s anywhere in the document, not just inside a `<form>`. It can also override an ancestor `<form>` element.

### DIFF
--- a/src/components/Button/Button.tsx
+++ b/src/components/Button/Button.tsx
@@ -120,6 +120,7 @@ export type ButtonProps = {
     onClick?: (event?: MouseEvent<HTMLButtonElement>) => void;
     hugWidth?: boolean;
     "aria-label"?: string;
+    formId?: string;
 };
 
 const ButtonComponent: ForwardRefRenderFunction<HTMLButtonElement | null, ButtonProps> = (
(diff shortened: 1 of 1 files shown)


mikeyrayvon:
This will be useful for `<Modal>`s and `<Flyout>`s with a form in the body and a submit button in the footer

julianiff (approved):
Great idea!

PR #792, opened 2022-05-13
fix: prevent tooltip appearing in modal

SamCreasey:
Prevents tooltip from automatically appearing when modal is opened, and also adds a tab-index to scrolling content so that it can be navigated via keyboard

### DIFF
--- a/src/components/Modal/Modal.tsx
+++ b/src/components/Modal/Modal.tsx
@@ -71,8 +71,7 @@ const ModalComponent: FC<ModalProps> = memo((props) => {
             data-is-underlay={true}
             className={`tw-fixed tw-top-0 tw-left-0 tw-bottom-0 tw-right-0 tw-flex tw-justify-center tw-items-center tw-p-4`}
         >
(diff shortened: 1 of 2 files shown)


imoutaharik:
@SamCreasey It looks good to me! But I'm just not sure about the tab-index and focus ring on the scrollable content. 
Should it be really focusable? WDYT @LidijaStoilova ?

LidijaStoilova:
Hi @SamCreasey 
I believe the modal element itself should not be focusable, as this might be confusing for screen reader users. The focus we have on the other elements should be enough. What I really like about it is, that the focus gets 'trapped' in the modal and doesn't jump to the background. Very well done!

SamCreasey:
@LidijaStoilova @imoutaharik This follows accessibility guidelines specified here https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/region_role#scrolling_content_areas_with_overflow_text. If there is no focusable content inside of the scrolling div then users cannot ever scroll the content via keyboard, so the scrolling div itself needs to be tabbable to cover for this case.

imoutaharik (approved):

PR #793, opened 2022-05-13
fix: allow `mailto` and `tel` links

SamuelAlev:
To fix https://frontify.slack.com/archives/C031PENPQE6/p1652368375709749


Walkeezy (approved):
👌

PR #801, opened 2022-05-16
Feat: modal compact padding option

mikeyrayvon:
- Adds a `padding` prop to the `<Modal>` to set either `Default` for `p-14` classes which result in 56px padding, or `Compact` for `p-6` classes which result in 24px padding. A `padding` object is distributed to Modal parts with a `layoutContext`
- Adds a storybook control for the `padding` prop and fixes the control for the `horizontalPadding` prop

### DIFF
--- a/src/components/Modal/Modal.stories.tsx
+++ b/src/components/Modal/Modal.stories.tsx
@@ -1,6 +1,6 @@
 /* (c) Copyright Frontify Ltd., all rights reserved. */
 
-import { Modal, MODAL_PADDING } from "./Modal";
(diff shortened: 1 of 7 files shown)


thisguyscode (approved):
@mikeyrayvon looks good to me, nice one!

As discussed there are some adjustments to the Header and Footer components to fully support the publisher's table view but I imagine that's best in another branch right?

mikeyrayvon:
Yeah let's do those adjustments to the Header and Footer in a separate branch 💯

PR #802, opened 2022-05-16
feat: Add Flyout positioning and offset options

imoutaharik:


thisguyscode:
@imoutaharik this little PR is going to make some people very happy, well done! Just one small comment from me.

imoutaharik:
@thisguyscode Thanks, but where is the comment? I can't see it. ;) suspense.

src/components/Flyout/Flyout.tsx
```diff
@@ -27,6 +27,13 @@ import { Overlay } from "./Overlay";
 export const FLYOUT_DIVIDER_COLOR = "#eaebeb";
 export const FLYOUT_DIVIDER_HEIGHT = "10px";
 
+export enum FlyoutPlacement {
+    Top = "top left",
+    Bottom = "bottom left",
+    Right = "right",
+    Left = "left",
```
jonasbuechel, line 34: Is aligning left for Top and Bottom the thing we like to enforce? Do we have a need for right-aligned as well?

src/components/Flyout/hooks/useOverlayPositionWithBottomMargin.ts
```diff
@@ -14,25 +15,31 @@ type UseOverlayPositionWithBottomMarginProps = {
...
     overlayRef,
     scrollRef,
     isOpen,
+    placement,
+    offset,
 }: UseOverlayPositionWithBottomMarginProps) => {
     const overlayHeight = getTotalOverlayHeight(overlayRef, scrollRef);
 
     const isFlipped = shouldDisplayAbove(triggerRef, overlayHeight, FLYOUT_OVERLAY_OFFSET, INTERCOM_BUTTON_HEIGHT);
+    const verticalPosition = isFlipped ? FlyoutPlacement.Top : FlyoutPlacement.Bottom;
```
jonasbuechel, line 33: If `placement = FlyoutPlacement.Top` and `isFlipped = false` it evaluates to Bottom.
The expected behavior would be displaying on top.

Screenshot 2022-05-23 at 11 29 01

jonasbuechel (changes requested):
I added a story to easier see placements and found an issue when using "top left"

src/components/Flyout/Flyout.tsx
imoutaharik: Top left and bottom left were originally enforced but I added the right position for more positioning flexibility.

src/components/Flyout/hooks/useOverlayPositionWithBottomMargin.ts
imoutaharik: I fixed it but the flip logic is handled manually as not all the flyout height is within the scroll area. It's calculated for top and bottom but we don't have flip logic done yet for right and left.

PR #803, opened 2022-05-16
fix: Fix focusRing being cut off in Modal

imoutaharik:
ClickUp ticket: #2atweze


imoutaharik:
> Screenshot 2022-05-24 at 16 19 33
> 
> Adding padding here causes content to be displayed outside of the scroll wrapper

@SamCreasey Thank you for spotting this, I fixed the issue. Can you please review it one more time? :)

imoutaharik:
> @imoutaharik I'm not sure if this is the best way to go regarding the scrollWrapper, I tried something similar before but there are some edge cases where it does not work. For example here when you scroll to the left it gets incredibly glitchy if you add/remove padding based on the shadow https://user-images.githubusercontent.com/93908356/170949762-9c9cb1e0-338b-4f15-97c7-df28bd322fdf.mov

@SamCreasey Talking with Jamie about it this morning, we decided to remove the left and right shadow to tackle the issue.

SamCreasey:
> @SamCreasey Talking with Jamie about it this morning, we decided to remove the left and right shadow to tackle the issue.

Screenshot 2022-06-03 at 13 34 46

@imoutaharik That makes sense to me. Two more things. We should remove the horizontal scroll wrapper story and consider what happens with this input. Since there is padding on the scroll area the right side of the input looks good here, but if the content is small enough to not require a scroll-bar on the right hand side then the input is not aligned with the footer buttons, which probably needs a designers opinion on too

imoutaharik:
> > @SamCreasey Talking with Jamie about it this morning, we decided to remove the left and right shadow to tackle the issue.
> 
> Screenshot 2022-06-03 at 13 34 46
> 
> @imoutaharik That makes sense to me. Two more things. We should remove the horizontal scroll wrapper story and consider what happens with this input. Since there is padding on the scroll area the right side of the input looks good here, but if the content is small enough to not require a scroll-bar on the right hand side then the input is not aligned with the footer buttons, which probably needs a designers opinion on too

Good catch! I fixed it. 😄

SamCreasey:
Screenshot 2022-06-14 at 11 05 01
You can see that the shadow clips the scrollbar half-way through here now, maybe this needs to be addressed also? 🤔

imoutaharik:
> Screenshot 2022-06-14 at 11 05 01
> 
> You can see that the shadow clips the scrollbar half-way through here now, maybe this needs to be addressed also? 🤔

@SamCreasey I think this should be good now. 😅

julianiff (approved):

SamCreasey (approved):

PR #819, opened 2022-05-25
fix: Adjust Tree component styles to match the designs

marguslant0:
Adjusting `Tree` component styles from this:
image

to this:
image

based on the notes and examples left by @thisguyscode on the Figma file here: https://www.figma.com/file/M0dX8EsqEgnHxNStt6Lgud/Fondue-React-Components---UPDATED?node-id=12226%3A195530 

Changes:
- Alignment of decorator/label regardless of whether there are child nodes
- Offset of child nodes
- Hover/Selection styles
- (Horizontal padding as a property - not implemented, as I didn't have a use case for it yet)


src/components/Tree/Node.tsx
```diff
@@ -141,27 +135,33 @@ export const Node = ({
...
                         onClick={onNodeClick}
                     >
-                        <div className="tw-flex tw-space-x-2 tw-items-center">
-                            <span data-test-id="toggle" onClick={toggleNodesVisibility}>
+                        <div className="tw-flex tw-space-x-1 tw-items-center">
+                            <span data-test-id="toggle" className="tw-w-3" onClick={toggleNodesVisibility}>
                                 {nodes &&
                                     (showNodes ? (
-                                        <IconCaretDown size={IconSize.Size16} />
+                                        <IconCaretDown size={IconSize.Size12} />
```
imoutaharik, line 162: Hey @marguslant0, Looking at the design, I don't think we should use the CaretDown Icon here. It should be 9px high and not 12 and completely filled. It would be better to make it custom, don't you think?

marguslant0: Hey @imoutaharik, thanks for the suggestion! Do you know if we have such custom triangle already in use somewhere in Arcade?

src/components/Tree/Node.tsx
```diff
@@ -141,27 +135,33 @@ export const Node = ({
             >
                 <div
                     className={merge([
-                        ' tw-flex tw-py-1 tw-px-2 tw-rounded tw-no-underline tw-leading-6',
+                        ' tw-flex tw-py-2 tw-px-2.5 tw-no-underline tw-leading-5',
```
imoutaharik, line 138:
```suggestion
                        'tw-flex tw-py-2 tw-px-2.5 tw-no-underline tw-leading-5 tw-text-text'
```

src/components/Tree/Node.tsx
```diff
@@ -141,27 +135,33 @@ export const Node = ({
...
                         strong && 'tw-font-bold',
-                        value && !selected && 'hover:tw-bg-black-5',
-                        selected ? 'tw-bg-violet-60 tw-text-white' : 'tw-text-black',
+                        value && !selected && 'hover:tw-bg-black-10 hover:tw-text-box-neutral-inverse-hover',
```
imoutaharik, line 140:
```suggestion
                        value && !selected && 'hover:tw-bg-box-neutral-hover hover:tw-text-box-neutral-inverse-hover',
```

src/components/Tree/Node.tsx
```diff
@@ -141,27 +135,33 @@ export const Node = ({
...
+                        selected
```
imoutaharik, line 141:
```suggestion
                        selected &&
                            'tw-bg-box-selected-strong tw-text-box-selected-strong-inverse hover:tw-bg-box-selected-strong-hover hover:tw-text-box-selected-strong-inverse-hover',
```

src/components/Tree/Node.tsx
```diff
@@ -141,27 +135,33 @@ export const Node = ({
             >
                 <div
                     className={merge([
```
imoutaharik, line 137: Some styling tokens weren't the correct ones according to design. I made some suggestions.

src/components/Tree/Node.tsx
```diff
@@ -141,27 +135,33 @@ export const Node = ({
...
-                        ' tw-flex tw-py-1 tw-px-2 tw-rounded tw-no-underline tw-leading-6',
+                        ' tw-flex tw-py-2 tw-px-2.5 tw-no-underline tw-leading-5',
                         strong && 'tw-font-bold',
-                        value && !selected && 'hover:tw-bg-black-5',
-                        selected ? 'tw-bg-violet-60 tw-text-white' : 'tw-text-black',
+                        value && !selected && 'hover:tw-bg-black-10 hover:tw-text-box-neutral-inverse-hover',
+                        selected
+                            ? 'tw-bg-violet-60 tw-text-white hover:tw-bg-box-selected-strong-hover hover:tw-text-box-selected-strong-inverse-hover'
+                            : 'tw-text-black',
```
imoutaharik, line 143: I think that the ternary condition overwrites the default text styling and sets it to black when the item is not selected.

imoutaharik (changes requested):

src/components/Tree/Node.tsx
imoutaharik: No, apparently all collapsable components in Arcade are using the Caret Icon. It seems only this one has a slightly different design.

marguslant0: Perhaps it's new the new intended design for all carets? If so, maybe it makes sense to keep it as is for now, for consistency?

src/components/Tree/Node.tsx
marguslant0: Good catch, updated! Thanks!

src/components/Tree/Node.tsx
imoutaharik: I would totally get @thisguyscode's input on this. 😉 Let's not decide on the design ourselves.

imoutaharik (approved):

src/components/Tree/Node.tsx
imoutaharik: We'll discuss it next week. I'll approve and merge this version for the urgency.

PR #832, opened 2022-06-01
fix(richtext): support old link format

rawi96:


src/components/RichTextEditor/plugins/linkChooserPlugin/ui/LinkChooserFlyout.tsx
```diff
@@ -37,7 +37,6 @@ export const LinkChooserFlyout = ({
         <Flyout
             isOpen={isFlyoutOpen}
             onOpenChange={setIsFlyoutOpen}
-            contentMinHeight={170}
```
triggertoo, line 40: revert?

src/components/RichTextEditor/plugins/linkChooserPlugin/transforms/getAndUpsertLink.ts
```diff
@@ -9,17 +9,29 @@ export const getAndUpsertLink = async <T = {}>(
     editor: PlateEditor<T>,
     getChosenLink?: LinkChooserPlugin['getChosenLink'],
 ) => {
-    const type = ELEMENT_LINK_CHOOSER;
+    const types = [ELEMENT_LINK, ELEMENT_LINK_CHOOSER];
```
triggertoo, line 12: I just wondered if it wouldn't be better if we stick to `ELEMENT_LINK` instead of introducing a new element type and guaranteeing compatibility to the previous version

PR #835, opened 2022-06-06
fix(RichTextEditor): responsive design & show correct selected dropdown trigger

celine-s:
Includes:
* responsive toolbar design (depending on RTE size)
* heading dropdown trigger shows correct label (heading1, mixed, etc.)

This resolves following two tickets: 
* [Design changes](https://app.clickup.com/t/2bmxehp)
* [RichTextEditor: show correct selected heading in the dropdownTrigger](https://app.clickup.com/t/2bgyakn)


src/components/RichTextEditor/RichTextEditor.tsx
```diff
@@ -46,6 +48,21 @@ export const RichTextEditor: FC<RichTextEditorProps> = ({
         onBlur: () => onBlur && onBlur(JSON.stringify(localValue.current)),
     };
 
+    useMemo(() => {
```
triggertoo, line 51: you could use a `useCallback` instead, since the dependency is mutable: see https://reactjs.org/docs/hooks-faq.html#how-can-i-measure-a-dom-node

src/components/RichTextEditor/TextStyleDropdown/DropdownTrigger.tsx
```diff
@@ -1,45 +1,86 @@
...
+
+const listTitle: Record<ListType, string> = {
+    [ListType.UL]: "Bullet List",
+    [ListType.OL]: "List",
+    [ListType.CHECKLIST_ITEM]: "Checklist",
+};
+
+type AvailableTextStyles = ListType & TextStyles;
+
+const availableTextStyleTitles: Record<AvailableTextStyles, string> = { ...listTitle, ...textStyleTitles };
```
triggertoo, line 36: you could rename this to `TEXT_STYLE_TITLES` to keep it shorter and uppercase for a file constant

src/components/RichTextEditor/TextStyleDropdown/DropdownTrigger.tsx
```diff
@@ -1,45 +1,86 @@
...
+
+const getSelectedTextStyles: (editor: PlateEditor) => AvailableTextStyles[] = (editor) => {
```
triggertoo, line 38:
```ts
const getSelectedTextStyles: (editor: PlateEditor) => AvailableTextStyles[] = (editor) => {
    if (!editor.selection) {
        return [];
    }
        
    return Array.from(
        getNodes(editor, {
            unhang: true,
            at: editor.selection,
            match: (node) => isElement(node) && isType(editor, node, Object.values({ ...TextStyles, ...ListType })),
        }),
    ).reduce<AvailableTextStyles[]>((types, current) => {
        const type = current[0].type as AvailableTextStyles;
        if (!types.includes(type)) {
            types.push(type);
        }
         return types;
    }, []);
};
```

triggertoo: could you create a custom hook from this and get the editor via plate hook, so this can be reused?

src/components/RichTextEditor/utils/toolbarCalc.ts
```diff
@@ -0,0 +1,50 @@
...
+            buttonGroupWidthsPerRow[rowCount].push({ actions, buttonGroupWidth, index });
+        } else {
+            buttonGroupWidthsPerRow.push([{ actions, buttonGroupWidth, index }]);
+        }
+
+        return buttonGroupWidthsPerRow.filter((element) => element.length > 0);
+    }, []);
+
+export const calculateToolbarWidth = (toolbarButtonGroups: ButtonGroupWidths[]) =>
+    toolbarButtonGroups?.length &&
```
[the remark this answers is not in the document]
triggertoo: this returns now `number|false`, this should always return a number

src/components/RichTextEditor/utils/toolbarCalc.ts
```diff
@@ -0,0 +1,50 @@
+import { ButtonGroupWidths } from '../types';
+import { EditorActions } from './actions';
+
+export const getButtonGroupWidthsPerRow = (toolbarWidth: number, buttonGroupsWidths: ButtonGroupWidths) =>
+    buttonGroupsWidths.reduce<ButtonGroupWidths[]>((buttonGroupWidthsPerRow, { actions, buttonGroupWidth, index }) => {
+        const rowCount = buttonGroupWidthsPerRow?.length === 0 ? 0 : buttonGroupWidthsPerRow?.length - 1;
```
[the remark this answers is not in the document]
celine-s: nice, thanks 👌😍

PR #842, opened 2022-06-13
feat(Tree): remove motion from actions

thisguyscode:


src/components/Tree/Node.tsx
```diff
@@ -190,19 +190,17 @@ export const Node = ({
...
-                                {actions.map((action) => action)}
-                            </motion.div>
-                        )}
-                    </AnimatePresence>
+
+                    {actions && (
+                        <div
+                            className={merge([
+                                'tw-flex tw-items-center',
+                                isHovered || selected ? 'tw-visible tw-space-x-1.5' : 'tw-invisible tw-w-0',
```
imoutaharik, line 198: What would be the purpose of having the visibility class? As the width is set on hover, I think it exactly the same as having `actions && (isHovered || selected) &&` on line 194. And that ternary condition might not be necessary. Don't you think?

src/components/Tree/Node.tsx
```diff
@@ -190,19 +190,17 @@ export const Node = ({
                             </span>
```
imoutaharik, line 190: There might be some import from framer-motion that need to be removed.

imoutaharik (changes requested):

src/components/Tree/Node.tsx
thisguyscode: It's an accesibility condition. Hover is not available to all users, it's only the visibility that we want to set.

It also ensures that the height of the row doesn't change when showing/hiding the button.

Either visibility or overflow: hidden would probably be fine but I went with visibility as it's more semantically clear.

imoutaharik: I see! Got it, thanks for the explanation. Then that would be just removing the import and I can approve. 😄

imoutaharik (approved):

imoutaharik:
> @imoutaharik do you know why this one is failing? Screenshot 2022-06-13 at 15 00 29

It was an issue with the node.js version but it's now fixed. On the other end, tests are taking ages. 🤔

PR #846, opened 2022-06-20
Functionality to add open Accordion panels

julianiff:
This PR adds functionality to the accordion to detect when new Panels are added to the !end! of the Accordion.


imoutaharik (approved):
We can unblock Michael like this but I think we need to rethink and refactor that component in the future.

julianiff:
Thank you for your review, ill forward the PR to Michael to see If this would fulfill his needs

PR #847, opened 2022-06-21
feat(RichText): Pass Custom DesignTokens into RichText

rawi96:


src/components/RichTextEditor/utils/getTextStyles.ts
```diff
@@ -1,14 +1,13 @@
 /* (c) Copyright Frontify Ltd., all rights reserved. */
```
celine-s, line 1: Maybe rename it to something like `textStyles` as getTextStyle function is removed

src/components/RichTextEditor/RichTextEditor.tsx
```diff
@@ -48,6 +48,28 @@ export const RichTextEditor: FC<RichTextEditorProps> = ({
         onBlur: () => onBlur && onBlur(JSON.stringify(localValue.current)),
     };
 
+    if (!designTokens.heading1) {
+        designTokens.heading1 = defaultDesignTokens.heading1;
+    }
+    if (!designTokens.heading2) {
```
celine-s, line 54: else if?

celine-s (commented):
I just added some small comments. Overall it looks really good and works smoothly👌 🥳

src/components/RichTextEditor/RichTextEditor.tsx
rawi96: There could be more than one style missing. With the else if we would lose it.

celine-s (approved):
nice! 🚀

PR #856, opened 2022-06-28
feat(RichTextEditor): add design tokens to context api

rawi96:
https://app.clickup.com/t/2cdd1dd

### DIFF
--- a/src/components/RichTextEditor/RichTextEditor.spec.tsx
+++ b/src/components/RichTextEditor/RichTextEditor.spec.tsx
@@ -4,6 +4,7 @@ import { mount } from '@cypress/react';
 import { ELEMENT_LINK, ELEMENT_PARAGRAPH } from '@udecode/plate';
 import React, { FC, useState } from 'react';
 import { ON_SAVE_DELAY_IN_MS, RichTextEditor, RichTextEditorProps } from './RichTextEditor';
+import { DesignTokens } from './types';
 import { EditorActions } from './utils/actions';
 
 const RICH_TEXT_EDITOR = '[data-test-id=rich-text-editor]';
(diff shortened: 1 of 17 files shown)


src/components/RichTextEditor/RichTextEditor.spec.tsx
```diff
@@ -64,6 +66,31 @@ const RichTextWithLink: FC<{ text: string; link: string }> = ({ text, link }) =>
...
 
+const RichTextWithChangeDesignTokensButton: FC = () => {
+    const [designTokens, setDesignTokens] = useState<DesignTokens>({
+        custom1: {
+            fontSize: '42px',
+        },
+    });
+    return (
+        <div>
+            <button
```
celine-s, line 77: smart way of testing this 👌 🙂

src/components/RichTextEditor/RichTextEditor.tsx
```diff
@@ -87,6 +66,30 @@ export const RichTextEditor: FC<RichTextEditorProps> = ({
         return observer;
     }, []);
 
+    useEffect(() => {
+        if (!designTokens.heading1) {
```
celine-s, line 70: You could do a loop over the existing `textstyles`. This way it would also be it's also dynamic when a new textstyle gets introduced.

```
for (const element of Object.values(TextStyles).filter(
            (textStyle) => textStyle !== TextStyles.ELEMENT_PARAGRAPH,
        )) {
            if (!designTokens[element]) {
                designTokens[element] = defaultDesignTokens[element];
            }
        }
        ```

celine-s (approved):
just added a small comment. Code looks good and it works smoothly also on beta block. Noiice! 🤩

PR #860, opened 2022-06-29
fix(Tree): Badge on editable tree node item

leko91:
There was a use case that I missed when developing the badge and editable feature, they didn't work together.

This is basically just adding a prop to the `EditableNodeItem` component so it also includes the badge (or icon) if it exists.

### DIFF
--- a/src/components/Tree/components/EditableNodeItem.tsx
+++ b/src/components/Tree/components/EditableNodeItem.tsx
@@ -4,21 +4,19 @@ import React, {
     FocusEventHandler,
     KeyboardEvent,
     KeyboardEventHandler,
-    ReactElement,
+    ReactNode,
     useRef,
     useState,
 } from 'react';
 
-import { BadgeProps, IconProps } from '../../..';
-
 export interface EditableNodeItem {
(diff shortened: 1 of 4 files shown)


src/components/Tree/components/EditableNodeItem.tsx
```diff
@@ -62,8 +66,13 @@ export const EditableNodeItem = ({ name, targetItemId, onEditableSave }: Editabl
...
                 </div>
             ) : (
-                <div data-test-id="node-link-name" onDoubleClick={handleDoubleClick}>
+                <div
+                    data-test-id="node-link-name"
+                    className="tw-flex tw-items-center"
+                    onDoubleClick={handleDoubleClick}
+                >
                     {name}
+                    {badge}
```
julianiff, line 75: Do you know if you need more than 1 badge component at a time? Otherwise, it might be then simpler to just add name and badge through the children prop. What do you think?

julianiff (approved):
Looks cool, now the badge is also there when editing

imoutaharik (approved):

PR #861, opened 2022-06-29
fix(Tree): Item multiselect bug

leko91:
There is currently a minor bug when selecting multiple items in the tree.

Since the code was only checking if the `command` or `control` key were pressed, there was some unexpected behavior when doing something like `command+tab` or `command+f` on the page. You can reproduce it on the [fondue-components](https://fondue-components.frontify.com/?path=/story/components-tree--tree), press `command+f` for example, and the multi-select will always be on. This PR fixes it.

### DIFF
--- a/src/components/Tree/Tree.tsx
+++ b/src/components/Tree/Tree.tsx
@@ -99,6 +99,8 @@ export const Tree: FC<TreeProps> = ({
     const downKeyHandler = (event: KeyboardEvent) => {
         if (event.key === 'Meta' || event.ctrlKey) {
             setMultiSelectMode(true);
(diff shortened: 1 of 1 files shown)


julianiff (approved):
Should the multiselectMode also exit when you click a 2nd time ctrl?

leko91:
> Should the multiselectMode also exit when you click a 2nd time ctrl?

@julianiff Technically, it should only work when holding ctrl/cmd (that's why we're removing the event listener onKeyUp). This only prevents the multselect mode from activating when doing something like ctrl+f or cmd+tab for example on the page.

PR #864, opened 2022-07-01
fix: fix issue with multiple DndProviders

Kenny806:


src/utilities/dnd/DndWrapper.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+import React, { PropsWithChildren, memo, useEffect, useState } from 'react';
+import { DndProvider } from 'react-dnd';
+import { HTML5Backend } from 'react-dnd-html5-backend';
+
+interface DndWrapperProps {
+    id: string;
+}
+
+export const DndWrapper = memo(({ id, children }: PropsWithChildren<DndWrapperProps>) => {
```
GoranRibic, line 11: I would avoid using PropsWithChildren as children should be mandatory in this case.

src/utilities/dnd/DndWrapper.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+    useEffect(() => {
+        setContext(document.getElementById(id));
+    }, [id]);
+
+    return context ? (
+        <DndProvider backend={HTML5Backend} options={{ rootElement: context }}>
+            {children}
+        </DndProvider>
+    ) : null;
```
GoranRibic, line 22: Better to return same time so return <></>.

src/components/Tree/Tree.tsx
```diff
@@ -118,20 +116,22 @@ export const Tree: FC<TreeProps> = ({
...
-                    nodes: treeNodes,
-                    treeName,
-                    activeIds,
-                    onClick: onNodeClick,
-                    onDrop: handleDrop,
-                    onEditableSave: handleEditableSave,
-                })}
-            </ul>
-        </DndProvider>
+        <div id={treeName}>
```
GoranRibic, line 119: Why is treeName used as id? Kind of misleading. Should it be treeId or just id?

Kenny806: Renamed it

GoranRibic (approved):

PR #874, opened 2022-07-13
feat(Textarea): added autosize property

SamCreasey:
Added a new feature to the Textarea component to be able to autosize when necessary. This component now has 3 new properties: autosize, minRows and maxRows. When setting the autosize to true (false by default) the textarea will automatically grow to max the size of the text until it hits the maxRow boundary. Since sizing is done computationally the resize handle for the textarea is removed for autosizing textareas.

### DIFF
--- a/src/components/Textarea/Textarea.spec.tsx
+++ b/src/components/Textarea/Textarea.spec.tsx
@@ -9,46 +9,47 @@ const PLACEHOLDER = 'This is the placeholder';
 const DECORATOR_TEXT = 'And also a decorator';
 const DECORATOR = <span>{DECORATOR_TEXT}</span>;
 const INPUT_TEXT = 'some text';
+const ROW_HEIGHT = 16;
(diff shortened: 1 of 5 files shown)


src/components/Textarea/Textarea.spec.tsx
```diff
@@ -9,46 +9,47 @@ const PLACEHOLDER = 'This is the placeholder';
 const DECORATOR_TEXT = 'And also a decorator';
 const DECORATOR = <span>{DECORATOR_TEXT}</span>;
 const INPUT_TEXT = 'some text';
+const ROW_HEIGHT = 16;
+const TEXTAREA_ID = 'textarea.tw-border';
```
imoutaharik, line 13: I think it would be better to add a data-test-id like for example `data-test-id="textarea-component"` or something like that instead of using the tag with the tailwind class. Don't you think?

SamCreasey: Fixed :)

PR #875, opened 2022-07-13
feat(RichText): add editable link text to flyout

rawi96:


src/components/RichTextEditor/plugins/linkChooserPlugin/transforms/getAndUpsertLink.ts
```diff
@@ -23,15 +23,34 @@ export const getAndUpsertLink = async <T = {}>(
...
                     link: linkNode[0].url,
                     icon: 'LINK',
                 },
                 openInNewTab: false,
             };
         } else {
-            prevChosenLink = linkNode[0].chosenLink as ChosenLink;
+            prevChosenLink = {
+                searchResult: {
+                    id: linkNode[0].chosenLink.searchResult.id,
```
triggertoo, line 35: could you check if linkNode[0] is defined and destruct to searchResult?

src/components/RichTextEditor/plugins/linkChooserPlugin/transforms/getAndUpsertLink.ts
```diff
@@ -23,15 +23,34 @@ export const getAndUpsertLink = async <T = {}>(
...
+                    title: linkNode[0].children[0].text,
+                    link: linkNode[0].chosenLink.searchResult.link,
+                    icon: linkNode[0].chosenLink.searchResult.icon,
+                },
+                openInNewTab: linkNode[0].chosenLink.openInNewTab,
+            };
         }
+    } else {
```
triggertoo, line 43: to avoid if else nesting, you could use early returns instead which improves the readabilty

triggertoo (approved):

PR #876, opened 2022-07-14
feat(theming): add theme-switcher feature

julianiff:
- Theme Switcher for Storybook
- Removed legacy values in tailwind.config
- To be used with https://github.com/Frontify/fondue-tokens/pull/84 as token inputs (needs to be merged first)


julianiff:
@imoutaharik @thisguyscode i added a test with the button text color to see if the themeing also works on a component level. And seems good so far. Its the button, positive solid if you want to check it out. (best with the side-by-side view)

thisguyscode:
@julianiff Thanks for all this! I just got around to setting my local by up, as we don't have the netlify previews at the moment. Will give a look ASAP

thisguyscode:
@julianiff this is incredible!!! I'm so happy to see all the lovely dark stuff.

Obviously, we don't expect all the dark modes to work, but I also noticed a regression in the light theme button, probably you're aware of it.

julianiff:
stale, opened in new PR

PR #881, opened 2022-07-15
chore: update cypress v10

celine-s:


cypress.config.ts
```diff
@@ -0,0 +1,20 @@
...
+    retries: {
+        runMode: 2,
+    },
+    component: {
+        setupNodeEvents() {
+            return require('./cypress/plugins/index.ts');
+        },
+        viewportHeight: 1280,
+        viewportWidth: 800,
+        specPattern: ['./src//**/*.cy.tsx', './**/*.spec.ts', './src//**/*.spec.tsx'],
```
julianiff, line 14: /src// the 2 // are intented?

tsconfig.json
```diff
@@ -24,6 +25,6 @@
         },
         "types": ["cypress", "cypress-real-events", "vite/client"]
     },
-    "include": ["src/types", "src/**/*"],
+    "include": ["src/types", "src/**/*", "cypress/support/component.ts"],
```
julianiff, line 28: why does component.ts need to be included in the output build?

julianiff (approved):

cypress.config.ts
celine-s: oh, no - thanks. Adapted in [ced93cc](https://github.com/Frontify/fondue/pull/881/commits/ced93cc13033c3e3660257880031d56da99aa50d)

tsconfig.json
celine-s: in the component the typing for the `mount` command is defined. That the test file know of this `cy.mount` it has to be included there. Would it be better/clearer if I would make a file `cypress/support/types.ts`?

julianiff: Nono i was just curious to understand the include :) Thx for clarifying :)

PR #889, opened 2022-07-22
chore: update fondue-tokens to v3.1.0-next.1

julianiff:
Update to use the new tokens in fondue


julianiff:
Dont want to use next version

PR #890, opened 2022-07-25
chore: various improvements

SamuelAlev:
- Update various deps
- Remove `pnpm-lock` as we use npm in this repository
- Add `key` attribute to RTE to allow change of initial value (cc @rawi96)
- Commit generated icons which haven't been pushed
- Clean output of build and add name on each files instead of having `index.es*.js`
- Add missing temp polyfill for styled-component


vite.config.ts
```diff
@@ -35,12 +35,15 @@ export default defineConfig({
         alias,
     },
     // needs to be defined here, such that it is not undefined in the tests.
-    define: { 'process.env.REACT_APP_SC_ATTR': process.env.REACT_APP_SC_ATTR },
+    define: {
+        'process.env.REACT_APP_SC_ATTR': JSON.stringify(process.env.REACT_APP_SC_ATTR),
```
celine-s, line 39: ahh, nice :)

celine-s (approved):

src/components/Tree/utils/getReorderedNodes.ts
```diff
@@ -14,7 +14,7 @@ export const getReorderedNodes = (
 
     return sourceItem
         ? moveItems<TreeFlatListItem>(
-              sameLevelNodes.at(targetItemIndex),
+              sameLevelNodes[targetItemIndex < 0 ? sameLevelNodes.length + targetItemIndex : targetItemIndex],
```
celine-s, line 17: maybe change to (sonarCloud issue):
`const index = targetItemIndex < 0 ? sameLevelNodes.length + targetItemIndex : targetItemIndex`
`sameLevelNodes[index]`

PR #891, opened 2022-07-26
Storybook: add eslint plugin & fix lint problems

julianiff:
Hey, i added the eslint checking to vite this time. Now we get updates while working in storybook. I put everything on warning to not break the build and will contact someone who works on the richtext, as there are some more errors that need to be looked at.


.storybook/main.ts
```diff
@@ -1,21 +1,20 @@
...
     },
     stories: ['../src/**/*.stories.tsx'],
     addons: ['storybook-dark-mode', '@storybook/addon-links', '@storybook/addon-essentials', '@storybook/addon-a11y'],
     staticDirs: ['assets'],
-    features: {
-        storyStoreV7: true,
-        buildStoriesJson: true,
-    },
+    // features: {
+    //     storyStoreV7: true,
```
julianiff, line 16: i removed storyStore again, cause doesnt work super great

src/typography/shared/Shared.stories.tsx
```diff
@@ -1,5 +1,6 @@
 /* (c) Copyright Frontify Ltd., all rights reserved. */
 
+export default {};
```
SamuelAlev, line 3: File can be renamed to `shared.ts` no?
Then the `export default` would no longer be required

julianiff: thx for the input, changing it

.storybook/main.ts
```diff
@@ -1,21 +1,20 @@
...
+    // },
```
julianiff, line 17:
```suggestion
```

.storybook/main.ts
julianiff: you are right

imoutaharik (approved):

PR #901, opened 2022-08-01
feat(MultiSelect): add smaller variant

fulopdaniel:
Add a smaller version to keep MultiSelect in line with other form components.

Before:
image
image

After:
image
image

### DIFF
--- a/src/components/MultiSelect/MultiSelect.stories.tsx
+++ b/src/components/MultiSelect/MultiSelect.stories.tsx
@@ -3,13 +3,14 @@
 import { Meta, Story } from '@storybook/react';
 import React, { useState } from 'react';
 import { FormControl } from '@components/FormControl';
-import { MultiSelect as MultiSelectComponent, MultiSelectProps, MultiSelectType } from './MultiSelect';
+import { MultiSelect as MultiSelectComponent, MultiSelectProps, MultiSelectSize, MultiSelectType } from './MultiSelect';
 
 export default {
     title: 'Components/Multi Select',
     component: MultiSelectComponent,
     args: {
         type: MultiSelectType.Default,
+        size: MultiSelectSize.Medium,
         activeItemKeys: ['Short tag', 'Tag 74'],
         items: [
             {
@@ -35,6 +36,10 @@ export default {
     argTypes: {
         type: {
             options: Object.keys(MultiSelectType),
+            control: { type: 'select' },
+        },
+        size: {
+            options: Object.keys(MultiSelectSize),
             control: { type: 'select' },
         },
     },
--- a/src/components/MultiSelect/MultiSelect.tsx
+++ b/src/components/MultiSelect/MultiSelect.tsx
@@ -1,18 +1,25 @@
 /* (c) Copyright Frontify Ltd., all rights reserved. */
 
 import { Checklist, ChecklistDirection } from '@components/Checklist/Checklist';
-import { Tag, TagType } from '@components/Tag/Tag';
+import { Tag, TagSize, TagType } from '@components/Tag/Tag';
 import { Trigger } from '@components/Trigger/Trigger';
 import { useButton } from '@react-aria/button';
 import { FocusScope, useFocusRing } from '@react-aria/focus';
 import { useOverlay } from '@react-aria/overlays';
 import { mergeProps } from '@react-aria/utils';
+import { merge } from '@utilities/merge';
 import { AnimatePresence, motion } from 'framer-motion';
 import React, { FC, useRef, useState } from 'react';
+import { getSizeClasses } from './helpers';
 
 export enum MultiSelectType {
     Default = 'Default',
     Summarized = 'Summarized',
+}
+
+export enum MultiSelectSize {
+    Small = 'Small',
+    Medium = 'Medium',
 }
 
 export type MultiSelectItem = {
@@ -28,6 +35,7 @@ export type MultiSelectProps = {
     ariaLabel?: string;
     placeholder?: string;
     type?: MultiSelectType;
+    size?: MultiSelectSize;
 };
 
 export const MultiSelect: FC<MultiSelectProps> = ({
@@ -38,6 +46,7 @@ export const MultiSelect: FC<MultiSelectProps> = ({
     disabled = false,
     placeholder,
     type = MultiSelectType.Default,
+    size = MultiSelectSize.Medium,
 }) => {
     const [open, setOpen] = useState(false);
     const overlayRef = useRef<HTMLDivElement | null>(null);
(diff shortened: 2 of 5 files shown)


src/components/MultiSelect/MultiSelect.stories.tsx
```diff
@@ -10,6 +10,7 @@ export default {
     component: MultiSelectComponent,
     args: {
         type: MultiSelectType.Default,
+        size: 'm',
```
imoutaharik, line 13: I think it would be better for readability to actually use the full word 'small', 'medium', etc... instead of just one letter. (even if it's totally understandable 😁 )

src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -15,6 +16,8 @@ export enum MultiSelectType {
     Summarized = 'Summarized',
 }
 
+export type MultiSelectSize = 's' | 'm';
```
imoutaharik, line 19: There has been a whole discussion on enum vs union types, you can check it here: https://weare.frontify.com/document/42#/frontend-react/typescript/the-enum-vs-union-types-discussion
All 'size' types in all components use enum so I think it's better to be consistent and change it to enum.

src/components/MultiSelect/MultiSelect.stories.tsx
```diff
@@ -37,6 +38,10 @@ export default {
             options: Object.keys(MultiSelectType),
             control: { type: 'select' },
         },
+        size: {
+            options: ['s', 'm'],
```
imoutaharik, line 42: See comment below on union type. It would be better to use the enum type like this: `Object.values(MultiSelectSize),`

src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -38,6 +42,7 @@ export const MultiSelect: FC<MultiSelectProps> = ({
     disabled = false,
     placeholder,
     type = MultiSelectType.Default,
+    size = 'm',
```
imoutaharik, line 45: same here with the recommended changes -> `MultiSelectType.Medium`

src/components/Tag/Tag.tsx
```diff
@@ -16,6 +16,8 @@ export enum TagType {
     PreviouslySelected = 'PreviouslySelected',
 }
 
+export type TagSize = 's' | 'm';
```
imoutaharik, line 19: Same as Multi select for the type.

src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -94,14 +99,18 @@ export const MultiSelect: FC<MultiSelectProps> = ({
                 <div
                     {...mergeProps(buttonProps, focusProps)}
                     ref={triggerRef}
-                    className="tw-flex-1 tw-flex tw-flex-wrap tw-gap-1 tw-pl-[19px] tw-py-[11px] tw-min-h-[34px] tw-outline-none tw-pr-7"
+                    className={merge([
+                        'tw-flex-1 tw-flex tw-flex-wrap tw-gap-1 tw-pl-[19px] tw-min-h-[34px] tw-outline-none tw-pr-7',
```
imoutaharik, line 103: Those custom values feel strange for the padding. Is there a link to the figma file?

src/components/Tag/Tag.tsx
```diff
@@ -33,15 +35,17 @@ type TagPropsSelected = {
...
 
 type TagPropsUnselected = {
     type: TagType.Suggested | TagType.PreviouslySelected;
     label: string;
     onClick?: null;
+    size?: TagSize;
 };
 
-export const Tag: FC<TagProps> = ({ type, label, onClick }) => {
+export const Tag: FC<TagProps> = ({ type, label, onClick, size = 'm' }) => {
```
imoutaharik, line 48: same here -> `size = Tagsize.Medium`

imoutaharik (changes requested):

src/components/MultiSelect/MultiSelect.tsx
fulopdaniel: I was more or less trying to get values from this one:
https://www.figma.com/file/WfEbqZI0nnPWHN43MCtvnm/Fondue---Inputs?node-id=83%3A22094

This is more like a middle ground, we need a multi-select input that is in line with other inputs (in terms of height).

src/components/Tag/Tag.tsx
fulopdaniel: Got it, fixed for all occurences! :)

src/components/MultiSelect/MultiSelect.tsx
imoutaharik: I see but could it be adjusted for the other inputs? Fondue Inputs have a size of 36px and I would rather have some consistency in the design system to have similar sizing. I would rather get similar values from the Text Input Component: https://fondue-components.frontify.com/?path=/story/components-text-input--text
But that's a design decision. 🙈 It would be better to have Designer feedback on this one.

On the other hand, specifying the height for the Small variant creates an issue when we select many options.
The tags overlap, and the multi-select should expand so I would use min-height instead.
Screenshot 2022-08-03 at 11 22 53

Also, the minimum height of the default size of MultiSelect as shown in Figma is 38px: https://www.figma.com/file/M0dX8EsqEgnHxNStt6Lgud/Fondue-Components---WIP?node-id=12222%3A190439

I would use something like this:

```
className={merge([
    'tw-flex-1 tw-flex tw-flex-wrap tw-gap-1 tw-pl-[19px] tw-outline-none tw-pr-7',
     size === MultiSelectSize.Small ? 'tw-min-h-9 tw-items-center' : 'tw-min-h-[38px] tw-py-[11px]',
])}
```
What do you think?

src/components/Tag/Tag.tsx
imoutaharik: Amazing! Thank you so much 🙌🏼 

That would be great if you could remove the FC, same discussion here: https://weare.frontify.com/document/42#/frontend-react/typescript/stop-using-react-fc

We should take the habit to remove it in all components we update if we see it still using it. So we can get this all cleaned up overtime 😄

fulopdaniel: Sure, makes sense! Done!

src/components/MultiSelect/MultiSelect.tsx
fulopdaniel: That is a very good point, good catch. 36px with borders is 34px without, but nevertheless I changed the approach to use paddings in all cases to avoid scenarios like your screenshot. This way it is more consistent, what do you think? (I created a helpers file for this similar to what you have for Badges)

imoutaharik: Yes, makes sense to me! Let's keep it to 34px so it is 36px with the border. Even if the Figma says 38px, I would rather have it exactly like the Text Input component. But I'll ask Jamie when he's back anyway.

src/components/MultiSelect/helpers.ts
```diff
@@ -0,0 +1,15 @@
+import { MultiSelectSize } from './MultiSelect';
+
+export const getSizeClasses = (size: MultiSelectSize, hasActiveKeys: boolean) => {
```
imoutaharik, line 3: I don't think we should return an empty string and having Medium size as the default anyway, I think that part won't even be reached.

We could simply have it like this:
`export const getSizeClasses = (size: MultiSelectSize, hasActiveKeys: boolean) => {
    if (size === MultiSelectSize.Small) {
        if (hasActiveKeys) {
            return 'tw-py-1';
        } else {
            return 'tw-py-2';
        }
    }
    return 'tw-py-[11px]';
};`

Don't you think?

fulopdaniel: Sure, we can do that 😇

imoutaharik (approved):

PR #920, opened 2022-08-16
feat(table): add sticky header

joseazevedo88:
Table header now sticks to the top of the container

image

### DIFF
--- a/src/components/Table/Table.tsx
+++ b/src/components/Table/Table.tsx
@@ -164,7 +164,7 @@ export const Table = ({
                                     <TableColumnHeader key={column.key} column={column} state={state} />
                                 ),
                             )}
-                            <td />
+                            <th />
                         </TableHeaderRow>
                     ))}
                 </thead>
--- a/src/components/Table/Table.tsx
+++ b/src/components/Table/Table.tsx
@@ -164,6 +164,7 @@ export const Table = ({
                                     <TableColumnHeader key={column.key} column={column} state={state} />
                                 ),
                             )}
+                            <td />
                         </TableHeaderRow>
                     ))}
                 </thead>
--- a/src/components/Table/TableHeaderRow.tsx
+++ b/src/components/Table/TableHeaderRow.tsx
@@ -13,7 +13,7 @@ export const TableHeaderRow: FC<TableHeaderRowProps> = ({ item, state, children 
     const { rowProps } = useTableHeaderRow({ node: item }, state, ref);
 
     return (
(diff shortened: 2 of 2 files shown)


src/components/Table/TableHeaderRow.tsx
```diff
@@ -13,7 +13,7 @@ export const TableHeaderRow: FC<TableHeaderRowProps> = ({ item, state, children
     const { rowProps } = useTableHeaderRow({ node: item }, state, ref);
 
     return (
-        <tr {...rowProps} ref={ref} className="tw-py-4 tw-px-8">
+        <tr {...rowProps} ref={ref} className="tw-py-4 tw-px-8 tw-sticky tw-top-0 tw-bg-white tw-z-10">
```
imoutaharik, line 16: Nice!! 🎉  I would just suggest to change the class `tw-bg-white` for `tw-bg-base`

src/components/Table/Table.tsx
```diff
@@ -164,6 +164,7 @@ export const Table = ({
                                     <TableColumnHeader key={column.key} column={column} state={state} />
                                 ),
                             )}
+                            <td />
```
imoutaharik, line 167: I think that closing tag is not necessary.

imoutaharik (changes requested):

src/components/Table/Table.tsx
joseazevedo88: It is a self closing tag in order to simulate the last column header, otherwise the background doesn't fill the entire container for some reason.

Without: 
image
 
With:
image

src/components/Table/TableHeaderRow.tsx
joseazevedo88: Will do! I thought we would need to use our color palette defined in `tailwind.config.js`. Isn't that our design system?

imoutaharik: We decided to use alias tokens instead of brand tokens. You can find them here: https://fondue-components.frontify.com/?path=/story/tokens--alias-tokens
Usually, you can find, on the Figma file, which token to use. Tokens can be confusing sometimes 😄

src/components/Table/Table.tsx
imoutaharik: Oh I see! But that just doesn't seem right to have that `td` there. Is there really no better way?

joseazevedo88: Actually it makes more sense to be an empty th! https://stackoverflow.com/a/19842702 Will update

imoutaharik (approved):

PR #923, opened 2022-08-18
chore(table): refactor components

joseazevedo88:
- Remove react-aria from table components
- Improve keyboard navigation, both checkboxes and action buttons can now be navigated to and interacted with using tabs and space(validated with Jamie)
- Improve code by removing the child components dependency on the huge state object created in the <Table> component.

### DIFF
--- a/src/components/Table/Table.tsx
+++ b/src/components/Table/Table.tsx
@@ -33,7 +33,7 @@ export type Column = {
 };
 
 export type Row = {
-    key: string | number;
+    key: Key;
     // cell keys have to correspond to column key values
     // e.g. Column { name: 'User', key: 'user' } ==> Row Cell { user: { id: 'anna', value: 'Anna' } }
     cells: Record<string, Cell>;
@@ -45,16 +45,19 @@ export type TableProps = PropsWithChildren<{
     rows: Row[];
     onSelectionChange?: (ids?: Key[]) => void;
     selectionMode?: SelectionMode;
-    selectedRowIds?: (string | number)[];
+    selectedRowIds?: Key[];
     ariaLabel?: string;
 }>;
 
-const DEFAULT_SORT_ORDER = 'descending';
+export enum SortDirection {
+    Ascending = 'ascending',
+    Descending = 'descending',
+}
 
-export type SortDirection = 'ascending' | 'descending';
+const DEFAULT_SORT_ORDER = SortDirection.Descending;
 
 type SortType = {
-    sortedColumnKey?: string | number;
+    sortedColumnKey?: Key;
     sortOrder?: SortDirection;
 };
 
@@ -79,11 +82,11 @@ const mapToTableAriaProps = (columns: Column[], rows: Row[]): TableStateProps<an
     ],
 });
 
-const getRowFromId = (rows: Row[], id: string | number) => rows.find(({ key }) => key === id) || null;
+const getRowFromId = (rows: Row[], id: Key) => rows.find(({ key }) => key === id) || null;
 
-const getAllRowIds = (rows: Row[]): (string | number)[] => rows.map(({ key: id }) => id);
+const getAllRowIds = (rows: Row[]): Key[] => rows.map(({ key: id }) => id);
 
-const sortRows = (rows: Row[], columnKey: string | number, isDescending: boolean) => {
+const sortRows = (rows: Row[], columnKey: Key, isDescending: boolean) => {
     const sort = (a: Row, b: Row) => {
         const keyA = a.cells[columnKey].sortId;
         const keyB = b.cells[columnKey].sortId;
--- a/src/components/Table/Table.tsx
+++ b/src/components/Table/Table.tsx
@@ -1,6 +1,5 @@
 /* (c) Copyright Frontify Ltd., all rights reserved. */
 
-import { useTable } from '@react-aria/table';
 import {
     Cell as AriaCell,
     Column as AriaColumn,
@@ -10,7 +9,7 @@ import {
     TableStateProps,
     useTableState,
 } from '@react-stately/table';
-import React, { PropsWithChildren, ReactNode, useRef, useState } from 'react';
+import React, { Key, PropsWithChildren, ReactNode, useRef, useState } from 'react';
 import { TableCell, TableCellType } from './TableCell';
 import { TableColumnHeader, TableColumnHeaderType } from './TableColumnHeader';
 import { TableHeaderRow } from './TableHeaderRow';
@@ -44,7 +43,7 @@ export type Row = {
 export type TableProps = PropsWithChildren<{
     columns: Column[];
     rows: Row[];
-    onSelectionChange?: (ids?: (string | number)[]) => void;
+    onSelectionChange?: (ids?: Key[]) => void;
     selectionMode?: SelectionMode;
     selectedRowIds?: (string | number)[];
     ariaLabel?: string;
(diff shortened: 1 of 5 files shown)


src/components/Table/Table.tsx
```diff
@@ -44,17 +43,19 @@ export type Row = {
...
     selectedRowIds?: (string | number)[];
     ariaLabel?: string;
 }>;
 
 const DEFAULT_SORT_ORDER = 'descending';
 
+export type SortDirection = 'ascending' | 'descending';
+
 type SortType = {
     sortedColumnKey?: string | number;
```
imoutaharik, line 57: It would make sense to use the `Key` type here as well. Also on type `Row` `key: Key` and other places. don't you think?

src/components/Table/Table.tsx
```diff
@@ -118,52 +119,57 @@ export const Table = ({
...
+                                        handleSortChange={onSortChange}
+                                        setSelectedRows={onSelectionChange}
                                     />
-                                ) : (
-                                    <TableColumnHeader key={column.key} column={column} state={state} />
-                                ),
-                            )}
+                                );
+                            })}
+                            <td />
```
imoutaharik, line 172: I think that closing `td` can be removed.

src/components/Table/TableCell.tsx
```diff
@@ -17,50 +11,69 @@ export enum TableCellType {
...
-    state: TableState<any>;
+    selectionMode: string;
     type?: TableCellType;
+    isChecked?: boolean;
+    selectedRows: Key[];
+    setSelectedRows?: (ids?: Key[]) => void;
 };
 
-export const TableCell: FC<TableCellProps> = ({ cell, state, type = TableCellType.Default }) => {
+export const TableCell: FC<TableCellProps> = ({
```
imoutaharik, line 21: `FC` can be removed here as well.

src/components/Table/TableCell.tsx
```diff
@@ -17,50 +11,69 @@ export enum TableCellType {
...
+            if (!setSelectedRows) {
+                return;
+            }
+
+            if (isChecked) {
+                const filteredRows = selectedRows.filter((row) => row !== cell.parentKey);
+                setSelectedRows(filteredRows);
+                return;
+            }
+            if (selectionMode === SelectionMode.SingleSelect) {
```
imoutaharik, line 43: That last condition could be simplified, WDYT?
```
const selectRows = selectionMode === SelectionMode.SingleSelect ? [cell.parentKey] : [...selectedRows, cell.parentKey];
setSelectedRows(selectRows);
```
giving the variable a better name of course. 😄

src/components/Table/TableCell.tsx
```diff
@@ -17,50 +11,69 @@ export enum TableCellType {
...
                 ref={ref}
                 className={merge([
                     'tw-pl-8 tw-py-4 tw-pr-4 tw-border-l-4',
-                    checked ? 'tw-border-violet-60' : 'tw-border-transparent',
+                    isChecked ? 'tw-border-violet-60' : 'tw-border-transparent',
                 ])}
                 data-test-id="table-select-cell"
             >
-                <Checkbox
+                <CheckboxCmp
```
imoutaharik, line 60: I'm curious, why was this renamed?

src/components/Table/TableCell.tsx
```diff
@@ -17,50 +11,69 @@ export enum TableCellType {
...
     return (
         <td
-            {...mergeProps(gridCellProps, focusProps)}
+            role="cell"
             ref={ref}
             className={merge([
                 'tw-p-4 tw-font-normal tw-text-xs focus:tw-outline-none',
-                checked ? 'tw-text-black-100 dark:tw-text-white' : 'tw-text-black-80 dark:tw-text-black-20',
-                isFocusVisible && FOCUS_STYLE_INSET,
+                isChecked ? 'tw-text-black-100 dark:tw-text-white' : 'tw-text-black-80 dark:tw-text-black-20',
```
imoutaharik, line 76: the classes here should be changed to use `Alias tokens` instead and avoid to use `dark` prefix.
We shouldn't have things like `tw-text-black-100` for example or violet-...

src/components/Table/TableColumnHeader.tsx
```diff
@@ -20,100 +14,105 @@ export enum TableColumnHeaderType {
...
     type?: TableColumnHeaderType;
+    rowIds: Key[];
+    sortDirection?: SortDirection;
+    selectionMode: string;
+    isColumnSorted?: boolean;
+    handleSortChange: (column: Key, direction: SortDirection) => void;
+    setSelectedRows?: (ids?: Key[]) => void;
 };
 
 export const TableColumnHeader: FC<TableColumnHeaderProps> = ({
```
imoutaharik, line 26: Same thing here `FC` can be removed.

src/components/Table/TableHeaderRow.tsx
```diff
@@ -1,19 +1,10 @@
...
-import { TableState } from '@react-stately/table';
 import React, { FC, useRef } from 'react';
 
-export type TableHeaderRowProps = {
-    item: any;
-    state: TableState<any>;
-};
-
-export const TableHeaderRow: FC<TableHeaderRowProps> = ({ item, state, children }) => {
+export const TableHeaderRow: FC = ({ children }) => {
```
imoutaharik, line 3: Another `FC`

src/components/Table/TableRow.tsx
```diff
@@ -1,30 +1,22 @@
...
 import React, { FC, useRef } from 'react';
 
 export type TableRowProps = {
-    item: any;
-    state: TableState<any>;
+    isSelected?: boolean;
 };
 
-export const TableRow: FC<TableRowProps> = ({ item, state, children }) => {
+export const TableRow: FC<TableRowProps> = ({ isSelected = false, children }) => {
```
imoutaharik, line 7: Same here with the `FC`

imoutaharik (changes requested):

imoutaharik:
Nice!! 🎉  I'm happy to see that component free of React-aria. I left a few comments here and there but it looks good overall.
I'm just not sure about the keyboard navigation, you can only select the checkbox by using `space`, I was trying to do it with `Enter`, that made more sense to me. 🤔

src/components/Table/Table.tsx
```diff
@@ -44,17 +43,19 @@ export type Row = {
...
-    onSelectionChange?: (ids?: (string | number)[]) => void;
+    onSelectionChange?: (ids?: Key[]) => void;
     selectionMode?: SelectionMode;
     selectedRowIds?: (string | number)[];
     ariaLabel?: string;
 }>;
 
 const DEFAULT_SORT_ORDER = 'descending';
 
+export type SortDirection = 'ascending' | 'descending';
```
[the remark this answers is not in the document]
joseazevedo88: Sounds good 👍

src/components/Table/Table.tsx
joseazevedo88: Total sense 👍

src/components/Table/TableCell.tsx
joseazevedo88: I also don't like it :( But it's a condition in the quality checks to not have sonarLint warnings and there was a warning in the `Checkbox` enum value because something called `Checkbox` already was being imported.

src/components/Table/TableCell.tsx
joseazevedo88: Sure 👍

src/components/Table/TableCell.tsx
joseazevedo88: Renamed to `CheckboxComponent` 👍

src/components/Table/TableCell.tsx
joseazevedo88: Will be resolved in another PR as per discussion in slack

PR #925, opened 2022-08-18
feat(RichTextEditor): Additional prop to be able to position toolbar

bojangles-m:


src/components/RichTextEditor/RichTextEditor.stories.tsx
```diff
@@ -15,11 +16,13 @@ export default {
         placeholder: 'Some placeholder',
         readonly: false,
         clear: false,
+        position: Position.FLOATING,
```
celine-s, line 19: Maybe add two stories to show the position differences of the toolbar.

src/components/RichTextEditor/EditorPositioningWrapper/ToolbarWrapper/ToolbarWrapperPositioningTop.tsx
```diff
@@ -0,0 +1,15 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+import React from 'react';
+import { ToolbarWrapperProps } from './types';
+
+export const ToolbarWrapperPositioningTop = ({ children }: ToolbarWrapperProps) => {
```
celine-s, line 6: I'm not sure if dropdown menu should close after selecting heading style in top/bottom position. Have you discussed this with Martin or Jamie?

celine-s (commented):
imo it looks good and the different toolbars are working smoothly within different window sizes. 👍

src/components/RichTextEditor/EditorPositioningWrapper/ToolbarWrapper/ToolbarWrapperPositioningTop.tsx
bojangles-m: The functionality is the same as before the change i made. Additional to it is only fixed position of menu.
If you look into the Fondu storybook it is the same behaviour.

src/components/RichTextEditor/RichTextEditor.stories.tsx
bojangles-m: I have concerns to such test they are prone to failures which are not even test failures but you have to adjust them even if it does not show the real failure. For example, the position is gonna be fixed on some text element and if the text changes that does not mean the Plugin is broken but the test is failing. Would like to avoid such tests which are in reality not a test. How do you feel about that?

rawi96: I guess what celine meant is not to add tests for that but add stories to have a demo in storybook where the position is top or botton.
But for me it is ok like it is now. Because you can change the position in the action menu in storybook.

src/components/RichTextEditor/RichTextEditor.cy.tsx
```diff
@@ -111,6 +115,10 @@ const RichTextWithChangeDesignTokensButton: FC = () => {
     );
 };
 
+const RichTextWithToolbarPositioning = ({ position }: { position?: Position }) => {
+    return <RichTextEditor position={position} />;
+};
+
```
triggertoo, line 121:
```suggestion
const RichTextWithToolbarPositioning = ({ position }: { position?: Position }) => <RichTextEditor position={position} />;
```

src/components/RichTextEditor/RichTextEditor.cy.tsx
```diff
@@ -539,3 +547,29 @@ describe('RichTextEditor Component', () => {
         );
     });
 });
+
+describe('RichTextEditor Component: Positioning of Toolbar', () => {
+    it('should render with floating toolbar', () => {
+        cy.mount(<RichTextWithToolbarPositioning />);
+
+        cy.get(TOOLBAR).should('not.visible');
```
triggertoo, line 555: Could you use different test id's in https://github.com/Frontify/fondue/pull/925/files#diff-0c679c41e3945db40c4d0d07465fe4b3be96d1285ed475c4d9d779ae5cb5e184R8 to distinguish if the correct wrapper is rendered?

rawi96 (approved):

src/components/RichTextEditor/context/RichTextEditorContext.ts
```diff
@@ -0,0 +1,22 @@
+import { createContext, useContext } from 'react';
+import { DesignTokens } from '../types';
+import { defaultDesignTokens } from '../utils/defaultDesignTokens';
+import { EditorPositioningWrapper, Position } from '../EditorPositioningWrapper';
+
+type RichTextEditorContextProps = {
+    designTokens: DesignTokens;
+    PositioningWrapper: typeof EditorPositioningWrapper[Position.FLOATING];
```
triggertoo, line 8: IMHO only the position of the toolbar should be provided by the context, since the component needs to rerender on every context change

triggertoo (commented):
looks good 👍 
one thing i noticed and we should probably take care of is, that the dropdown and flyout currently open at the bottom of the toolbar when it is positioned at the bottom of the editor.
positioning

src/components/RichTextEditor/RichTextEditor.cy.tsx
bojangles-m: Updated

src/components/RichTextEditor/RichTextEditor.cy.tsx
bojangles-m: Updated

src/components/RichTextEditor/RichTextEditor.stories.tsx
bojangles-m: Sorry misunderstood the point, Added a story for Positioning with fixed Top

triggertoo (approved):

triggertoo (approved):

PR #926, opened 2022-08-19
feat(OverflowMenu): add basic component implementation and tests

fabriciomendonca:
This PR:
- add the `OverflowMenu` component, subcomponent, stories, and tests
- export the `getItemElementType` function in the utilities folder
- refactor components to use the new exported function


src/components/OverflowMenu/OverflowMenuItem.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+export interface OverflowMenuItemProps {
+    label: string;
+    link?: string;
+    onClick?: (event?: MouseEvent<HTMLButtonElement>) => void;
+    decorator?: ReactElement<IconProps>;
+}
+
+// TODO - The same function is also used on BreadcrumbItem
+// should we export it on the utilities?
```
imoutaharik, line 15: Yes 😄

fabriciomendonca: Done :D

src/components/OverflowMenu/OverflowMenu.cy.tsx
```diff
@@ -0,0 +1,86 @@
...
+            cy.mount(<OverflowMenu items={SPAN_ITEMS} />);
+            cy.get(COMPONENT_TEST_ID).as('OverflowMenu');
+        });
+
+        it('renders the component with links', () => {
+            cy.get('@OverflowMenu').should('be.visible', 1);
+        });
+
+        it('opens the menu and checks if the spans are present', () => {
+            cy.get('@OverflowMenu').find('button').click();
```
imoutaharik, line 79: It would be nice to add some tests for the Tabbing into elements and check if the focus is visible where it should be.

src/components/OverflowMenu/OverflowMenu.stories.tsx
```diff
@@ -0,0 +1,96 @@
...
+                <BreadcrumbItem label="Breadcrumb item 4" showSeparator={false} />
+            </div>
+            <small>
+                The separator should be handled by the <strong>Breadcrumbs</strong> component
+            </small>
+        </div>
+    ),
+];
+
+export const BetweenBreadcrumbItems = Template.bind({});
```
imoutaharik, line 81: That's probably something to check if design but the huge focusRing (I still really think it's way too big) is cropping the next breadcrumbItem:
Screenshot 2022-09-06 at 17 03 39

src/components/OverflowMenu/OverflowMenu.tsx
```diff
@@ -0,0 +1,66 @@
...
+import { Button, ButtonRounding, ButtonSize, ButtonStyle, ButtonType } from '..';
+import { OverflowMenuItem, OverflowMenuItemProps } from './OverflowMenuItem';
+
+export interface OverflowMenuProps {
+    items: OverflowMenuItemProps[];
+}
+
+export const OverflowMenu = ({ items }: OverflowMenuProps) => {
+    const [isMenuOpened, setIsMenuOpened] = useState(false);
+    const [menuContainerRef, setMenuContainerRef] = useState<HTMLElement | null>(null);
```
imoutaharik, line 15: Here for the ref you can instead use `useRef`:
`const menuContainerRef = useRef<HTMLButtonElement | null>(null);`

imoutaharik (changes requested):
Very Good!! 🎉  This looks really nice 😄 I just left a few comments and also tomorrow I'll give another review for accessibility. I want to check if we miss some aria attributes here and where.

src/components/OverflowMenu/OverflowMenu.tsx
fabriciomendonca: As discussed on our call, I'm using the popper.js recommended hook for updating the container properly when the element changes. The docs link: https://popper.js.org/react-popper/v2/

src/components/OverflowMenu/OverflowMenu.stories.tsx
fabriciomendonca: Nice catch, since I'm using the `BreadcrumbItem` component directly, we see this issue. Will update the story and also add the separator like it should be done on the `Breadcrumbs` component itself.

imoutaharik:
@fabriciomendonca There is a code duplication failed check from sonarCloud for the cypress test. There is a repeated sequence for the keyboard navigation, It would be great if you could make that sequence reusable and reduce it.

src/components/OverflowMenu/OverflowMenu.stories.tsx
imoutaharik: Awesome! Thanks for that 🥳 Practically done with the first component! Can you just please add that focus ring styling to the button so then I think all is good and ready. @LidijaStoilova will have a check on accessibility tomorrow morning.

LidijaStoilova:
In general keyboard accessibility works pretty well - there is just one thing missing that when you hit the ESC key, the overflow menu should close. Also, I see some violations in the axe plugin, that would be nice to resolve. What do you think?

fabriciomendonca:
> In general keyboard accessibility works pretty well - there is just one thing missing that when you hit the ESC key, the overflow menu should close. Also, I see some violations in the axe plugin, that would be nice to resolve. What do you think?

Sorry for the late reply. I pushed the changes for closing the menu when pressing the Esc key. I checked the AXE plugin and couldn't see any violations there. Could you please share those violations, @LidijaStoilova?

LidijaStoilova:
I don't see any violations anymore, maybe it didn't load it correctly on Storybook when I checked. Good job on the ESC!

imoutaharik (approved):

PR #932, opened 2022-08-30
feat(RichTextEditor): integrating plugin mentions

bojangles-m:


src/components/RichTextEditor/components/mention.tsx
```diff
@@ -0,0 +1,45 @@
...
+    return (
+        <span
+            {...attributes}
+            data-slate-value={element.value}
+            data-slate-key={element.key}
+            data-slate-category={element.category}
+            contentEditable={false}
+            data-cy={`mention-${element.value?.replaceAll(' ', '-')}`}
+            style={{
+                padding: '3px 3px 2px',
```
rawi96, line 29: Could you change the inline styles to tailwind css classes?

src/components/RichTextEditor/utils/editorConfig.ts
```diff
@@ -169,3 +179,28 @@ export const getEditorConfig = () => {
         },
     );
 };
+
+export const EditorConfigAnnotation = () => {
```
rawi96, line 183: What is the reason for that?
Couldn't we just take the editorConfig like before and manage with the editorActions which plugins we want to use and which not?

rawi96 (commented):
Nice 😊 

2 little comments. 
And I would definitely recommend adding some cypress e2e Tests for this new mention-plugin.

But unfortunately, it doesn't work 100% correctly. Sometimes when I try to mention several people in a row, you suddenly get an @ but the dropdown doesn't appear. 🤔

src/components/RichTextEditor/RichTextEditor.tsx
```diff
@@ -108,12 +111,13 @@ export const RichTextEditor: FC<RichTextEditorProps> = ({
...
                     id={editorId}
-                    initialValue={parseRawValue(initialValue)}
+                    initialValue={parseRawValue(initialValue, config)}
                     onChange={onChange}
                     editableProps={editableProps}
-                    plugins={getEditorConfig()}
+                    plugins={getEditorConfig(config)}
                 >
                     <Toolbar editorId={editorId} actions={actions} editorWidth={editorWidth} />
+                    {config === EditorConfigType.ANNOTATIONS && <MentionCombobox items={mentionable} />}
```
rawi96, line 120: Should we not control that via the editorActions as well? 🤔 
Instead of introducing this new config "EditorConfigType"

bojangles-m: The way the Slate and Plate works is like this and cant be controlled over Actions. You need mention combobox and when you have it every RTE is dependent and mentions working in all. This is not a toolbar option, there is no action for mentions the actions are accepted by toolbar and if there is a missing action you can use it in a toolbar. It is not dependent on toolbar.

src/components/RichTextEditor/utils/editorConfig.ts
bojangles-m: The reason for that is, if I add to normal config I will have mentions always in every RTE even if I dont want to use it, cant control it over action.

rawi96 (approved):

PR #935, opened 2022-09-05
feat(EditableInput): Move EditableInput into own component

julianiff:
New Component EditableNode moved from Tree to own component

### DIFF
--- a/src/components/EditableInput/EditableInput.cy.tsx
+++ b/src/components/EditableInput/EditableInput.cy.tsx
@@ -32,12 +32,12 @@ describe('EditableInput', () => {
     });
 
     it('Event gets triggered when writing Hey in input form that nothing is sent and when clicking away that it sent an event', () => {
-        const onChangeSpy = cy.spy().as('onChangeSpy');
+        const onEditableSaveSpy = cy.spy().as('onChangeSpy');
         cy.mount(
             <EditableInput
                 name={'This Input'}
                 targetItemId={'id'}
-                onEditableSave={(one: string, two: string) => onChangeSpy(two)}
+                onEditableSave={(one: string, two: string) => onEditableSaveSpy(two)}
             ></EditableInput>,
         );
         cy.get(EDITABLE_LINK_NAME).dblclick();
--- a/src/components/EditableInput/EditableInput.cy.tsx
+++ b/src/components/EditableInput/EditableInput.cy.tsx
@@ -3,7 +3,7 @@ import { EditableInput } from '@components/EditableInput/EditableInput';
 
 const EDITABLE_CONTAINER = '[data-test-id=editable-node-container]';
 const EDITABLE_LINK_NAME = '[data-test-id=node-link-name]';
-const EDITABLE_NODE = '[data-test-id=node-editable]';
+const EDITABLE_NODE = '[data-test-id=editable-input]';
 
 describe('EditableInput', () => {
     beforeEach('Setup component', () => {
--- a/src/components/EditableInput/EditableInput.cy.tsx
+++ b/src/components/EditableInput/EditableInput.cy.tsx
@@ -30,6 +30,7 @@ describe('EditableInput', () => {
 
         cy.get(EDITABLE_NODE).should('exist');
     });
+
     it('Event gets triggered when writing Hey in input form that nothing is sent and when clicking away that it sent an event', () => {
         const onChangeSpy = cy.spy().as('onChangeSpy');
         cy.mount(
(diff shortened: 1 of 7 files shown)


src/components/EditableInput/EditableInput.cy.tsx
```diff
@@ -0,0 +1,66 @@
...
+                onEditableSave={(one: string, two: string) => onChangeSpy(two)}
+            ></EditableInput>,
+        );
+        cy.get(EDITABLE_LINK_NAME).dblclick();
+
+        cy.get(EDITABLE_NODE).type('Hey');
+
+        cy.get('@onChangeSpy').should('have.not.been.called.with');
+        cy.get(EDITABLE_CONTAINER).click();
+        cy.get('@onChangeSpy').should('have.been.called.with');
```
fabriciomendonca, line 48: Shouldn't we also check the value when using `should.have.been.called.with`? Like `cy.get('@onChangeSpy').should('have.been.called.with', 'Hey');` or something?

fabriciomendonca (approved):
lg2m. Left just one non-blocking question. Nice stuff 💪

src/components/EditableInput/EditableInput.cy.tsx
leko91: I think just `should.have.been.called` is enough here.

src/components/EditableInput/EditableInput.cy.tsx
```diff
@@ -0,0 +1,66 @@
...
+                onEditableSave={() => console.log('edit')}
+            ></EditableInput>,
+        );
+        cy.get(EDITABLE_LINK_NAME).should('exist');
+        cy.get(EDITABLE_LINK_NAME).dblclick();
+
+        cy.get(EDITABLE_NODE).should('exist');
+    });
+    it('Event gets triggered when writing Hey in input form that nothing is sent and when clicking away that it sent an event', () => {
+        const onChangeSpy = cy.spy().as('onChangeSpy');
```
leko91, line 35: We should name it `onEditableSaveSpy` so it's more readable.

src/components/EditableInput/EditableInput.cy.tsx
julianiff: thx nice input, ill added it

src/components/EditableInput/EditableInput.cy.tsx
```diff
@@ -0,0 +1,66 @@
+import React from 'react';
+import { EditableInput } from '@components/EditableInput/EditableInput';
+
+const EDITABLE_CONTAINER = '[data-test-id=editable-node-container]';
+const EDITABLE_LINK_NAME = '[data-test-id=node-link-name]';
+const EDITABLE_NODE = '[data-test-id=node-editable]';
+
+describe('EditableInput', () => {
+    it('renders without error', () => {
```
leko91, line 9: Here you can do a `beforeEach` and mount the component instead of in every test:

```
const EDITABLE_INPUT_NAME = 'Editable Input';
const EDITABLE_INPUT_ID = 'ID';

beforeEach('Mount component', () => {
    const onEditableSaveSpy = cy.spy().as('onEditableSaveSpy');
    
    cy.mount(
        <EditableInput
            name={EDITABLE_INPUT_NAME}
            targetItemId={EDITABLE_INPUT_ID}
            onEditableSave={onEditableSaveSpy}
        />,
    );
}
```

It's a bit cleaner maybe.

src/components/EditableInput/EditableInput.tsx
```diff
@@ -47,17 +47,17 @@ export const EditableNodeItem = ({ name, targetItemId, onEditableSave, children
     };
 
     return (
-        <div>
+        <div data-test-id="editable-node-container">
             {showInput ? (
                 <div className="tw-flex tw-items-center">
                     <div
                         data-test-id="node-editable"
```
leko91, line 54: This one stayed a `node-editable` and the new one is `editable-node-container`. For consistency could we rename them to `editable-input`?

leko91:
There are now some duplicate tests in the `Tree.cy.tsx`, but I guess never enough tests 😄 
Just make sure the old ones also work still.

leko91 (approved):

src/components/EditableInput/EditableInput.cy.tsx
julianiff: Thx for this input, was not thinking about mounting it in the beforeEach

julianiff:
Thx for your inputs, ill adjusted the files 👍

PR #941, opened 2022-09-05
feat(RichTextEditor): added padding property in rte

bojangles-m:


src/components/RichTextEditor/RichTextEditor.tsx
```diff
@@ -51,6 +53,7 @@ export const RichTextEditor: FC<RichTextEditorProps> = ({
         placeholder,
         readOnly: readonly,
         onBlur: () => onBlur && onBlur(JSON.stringify(localValue.current)),
+        className: `tw-p-${padding}`,
```
triggertoo, line 56: this won't work in production mode, since it get's purged: see https://v2.tailwindcss.com/docs/just-in-time-mode "Dynamic values"

triggertoo, line 56:
```suggestion
        className: `tw-p-[${padding}]`,
```

triggertoo:
Screenshot 2022-09-12 at 08 36 37

seems that XLarge does not apply properly in storybook

triggertoo (approved):

PR #949, opened 2022-09-07
feat(EditableText)!: add new properties

julianiff:
Component to switch between Label and Input state
 
Now Clones the children and adds clickHandlers to flip on Click.
 Events:
 - onEditableSave // When changing from input to label state
 - onModeChange // Indicates which state is active
 - onAdditionalValueSave // Similar to editable Save with additional Values provided
 
 Options properties:
 - customContainerClasses
 - customInputTextClasses
 - mode: enum EditableMode {INPUT = 'INPUT',LABEL = 'LABEL'}
 - enableDoubleClick: Should the input be triggered by a double click or single click
 - additional

### DIFF
--- a/src/components/EditableInput/EditableInput.tsx
+++ b/src/components/EditableInput/EditableInput.tsx
@@ -4,6 +4,7 @@ import React, {
     FocusEventHandler,
     KeyboardEvent,
     KeyboardEventHandler,
+    PropsWithChildren,
     useEffect,
     useRef,
     useState,
@@ -40,13 +41,12 @@ interface EditableOptionProps {
  * onModeChange callback when state changes
  * options to specify styling and additional behaviours
  */
-export interface EditableInputProps {
-    children?: React.ReactElement;
+export type EditableInputProps = PropsWithChildren<{
     onAdditionalValueSave?: (additionalValue: string, value: string) => void;
     onEditableSave?: (value: string) => void;
     onModeChange?: (editableState?: EditableMode) => void;
     options?: EditableOptionProps;
-}
+}>;
 
 /**
  * Component to switch between Label and Input state. Wraps the Children
--- a/src/components/EditableInput/EditableInput.tsx
+++ b/src/components/EditableInput/EditableInput.tsx
@@ -64,11 +64,6 @@ export interface EditableInputProps {
  * - enableDoubleClick:      boolean            // should the input be triggered by a double click or single click
  * - additionalValues:       string             // additional information to be passed via onClick Event
  *
- * @param onEditableSave
- * @param onModeChange
- * @param onAdditionalValueSave
- * @param children
- * @param options
  */
 export const EditableInput = ({
     onEditableSave,
--- a/src/components/EditableInput/EditableInput.tsx
+++ b/src/components/EditableInput/EditableInput.tsx
@@ -52,9 +52,9 @@ export interface EditableInputProps {
  *
  * Clones the children and adds clickHandlers to flip on Click.
  * Events:
- * - onEditableSave // When changing from input to label state
- * - onModeChange // Indicates which state is active
- * - onAdditionalValueSave // Similar to editable Save with additional Values provided
+ * - onEditableSave         // When changing from input to label state
+ * - onModeChange           // Indicates which state is active
+ * - onAdditionalValueSave  // Similar to editable Save with additional Values provided
  *
  * Options properties:
  * - customContainerClasses: string             // Tailwind Classes
--- a/src/components/EditableInput/EditableInput.tsx
+++ b/src/components/EditableInput/EditableInput.tsx
@@ -4,100 +4,151 @@ import React, {
     FocusEventHandler,
     KeyboardEvent,
     KeyboardEventHandler,
-    ReactNode,
+    useEffect,
     useRef,
     useState,
 } from 'react';
+import { EditableInputHelper } from '@components/EditableInput/lib/helper';
+import { DEFAULT_CONTAINER_CLASS, DEFAULT_INPUT_TEXT_CLASS } from '@components/EditableInput/constant';
 
 export enum EditableMode {
     INPUT = 'INPUT',
     LABEL = 'LABEL',
-    AUTO = 'AUTO',
 }
 
-export interface EditableInputProps {
-    name: string;
-    targetItemId: string;
-    onEditableSave: (targetItemId: string, value: string) => void;
-    onModeChange?: (editableState: EditableMode) => void;
+/**
+ * Define custom Styles with
+ * - customContainerClasses (<div>)
+ * - customInputTextClasses (<input>)
+ *
+ * mode: display as Input or Label first
+ * enableDoubleClick: Should the input be triggered by a double click or single click
+ * targeItemId: additional information to be passed via onClick Event
+ */
+interface EditableOptionProps {
+    customContainerClasses?: string;
+    customInputTextClasses?: string;
     mode?: EditableMode;
-    children?: ReactNode;
-    singleClick?: boolean;
+    enableDoubleClick?: boolean;
+    additionalValues?: string;
 }
 
+/**
+ * Add Children with badges or icons
+ * onEditableSave callback to get updated Value
+ * onAdditionalValueSave callback with more information (ex. ID)
+ * onModeChange callback when state changes
+ * options to specify styling and additional behaviours
+ */
+export interface EditableInputProps {
+    children?: React.ReactElement;
+    onAdditionalValueSave?: (additionalValue: string, value: string) => void;
+    onEditableSave?: (value: string) => void;
+    onModeChange?: (editableState?: EditableMode) => void;
+    options?: EditableOptionProps;
+}
+
+/**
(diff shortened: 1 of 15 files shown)


imoutaharik:
@julianiff I haven't gone through all the code but I have a question about the design. Is there any Figma you could share, please? Do the size of the input depend on font size? When clicking on the text (single click), the input is focused (the focus ring is not the same style as other Fondue inputs), but on double click, the input is not focused. also when the text is big, the pen icon looks a bit off and really small. 🤔

julianiff:
Hey @imoutaharik
- I dont think there really exists a figma design. The closest will be the one from the tree (https://www.figma.com/file/pevuvey1joCsopTRr9f4OE/Fondue---Components?node-id=12226%3A195573) its based on the work of Leko.
Screenshot 2022-09-13 at 10 21 01
- The text flips on click, but you can change the container and text of the input as it is needed (with the options prop). As the requirement.
- The icon is added outside of the component, and should be adjusted from the client. I can adjust the iconsize in the story.
- WIth double-click it also focuses when clicking, you can try it with this link (https://deploy-preview-949--fondue-components.netlify.app/?path=/story/components-editable-input--default&args=options.enableDoubleClick:true). In the docs its a bit weird, as there is a story about the focus on the input.

imoutaharik:
Thanks, @julianiff for the explanation and for sharing the links 😄 . I have another few questions and 1 suggestion 🤔  :
- At first glance, I thought there was a styling issue with the double click story but it actually has default input styling compared to the first one (single click) that has custom styling on the input. I would suggest switching so the "hey default" story is all default properties.
- What was the decision about making the Icon so flexible to be fully handled outside? Is there any case where the icon could change? As it is an editable input, it makes sense it's always the `Pen` icon. I'm wondering...
- In the Figma file, the Pen icon actually appears on hover which is more of a UX decision to hint to the user that the label is editable. But in the stories, it's more like we can either have it always present or not at all. I feel that in the first example, it's not really clear if it's editable. Should this really need to be done by the users or should already be addressed in the component?
- And Finally, I noticed the label is not accessible with tabbing. I can't access the editable input and update only using the keyboard. I have to click on it.
WDYT?

julianiff:
At first glance, I thought there was a styling issue with the double click story but it actually has default input styling compared to the first one (single click) that has custom styling on the input. I would suggest switching so the "hey default" story is all default properties.
- As the options object is not that obvious I thought ill fill it with values to have them displayed to the user immediately. But can also remove it. 

What was the decision about making the Icon so flexible to be fully handled outside? Is there any case where the icon could change? As it is an editable input, it makes sense it's always the Pen icon. I'm wondering...
- Only reasoning was to keep the component smaller with less properties. (as icon/badge would be needed to be added). But that can be added as a prop without any issues.

In the Figma file, the Pen icon actually appears on hover which is more of a UX decision to hint to the user that the label is editable. But in the stories, it's more like we can either have it always present or not at all. I feel that in the first example, it's not really clear if it's editable. Should this really need to be done by the users or should already be addressed in the component?
- Again it is done to have the component small, but that can be done also in the component (myself I have no preference). There exists a onModeChange event to hint the user the state of the component. As this component could also be put over a breadcrumb, or other elements. Could make a lot of sense to have the icon inside, but then for example in the tree it would need to be removable or exchangeable to fulfill their needs.

And Finally, I noticed the label is not accessible with tabbing. I can't access the editable input and update only using the keyboard. I have to click on it.
- Good Point, thx Ill check it out if i find smth like a aria-* that could be put on the children element to suggest screenreaders that it is a clickable element. In an earlier iteration we had a button present, maybe it would make sense to bring that one back.

julianiff:
@imoutaharik 
Check it out, newly added the focus-visible and fixed the accessibility issues. 
-> The children are now wrapped in a button that is accessible and has focus styling. Now that part works. That also removed children cloning. This Code change feels a bit less like a chunky design.

To avoid the PR getting stuck, I suggest we handle the icons and these questions in a separate PR. An Icon also introduces more states (size of the icon, icon selection (tree needs a different icon than the metadata management area), visible when editing, on hover or not). 

I am looking forward to the rest of the Code Review.

src/components/EditableInput/EditableInput.tsx
```diff
@@ -4,78 +4,156 @@ import React, {
...
+ *
+ * Options properties:
+ * - customContainerClasses: string             // Tailwind Classes
+ * - customInputTextClasses: string             // Tailwind Classes
+ * - mode:                   enum EditableMode  // {INPUT = 'INPUT',LABEL = 'LABEL'}
+ * - enableDoubleClick:      boolean            // should the input be triggered by a double click or single click
+ * - additionalValues:       string             // additional information to be passed via onClick Event
+ *
+ * @param onEditableSave
+ * @param onModeChange
```
imoutaharik, line 68: I'm personally not a big fan of JsDocs in React components, it looks noisy and redundant to name the params when we have interfaces. The code speaks of itself. And we don't really use it anywhere else. 🤔 But we can ask what others prefer.

src/components/EditableInput/lib/helper.ts
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
```
imoutaharik, line 1: You import ReactElement here so you don't have to use React.ReactElement for the typing.
`import React, {ReactElement} from 'react'`

src/components/Tree/Node.tsx
```diff
@@ -170,9 +170,15 @@ export const Node = ({
                             </span>
                             {icon && <span className="tw-flex tw-justify-center tw-items-center tw-w-5">{icon}</span>}
                             {editable && onEditableSave ? (
-                                <EditableInput name={name} targetItemId={node.id} onEditableSave={onEditableSave}>
+                                <>
+                                    <EditableInput
+                                        options={{ additionalValues: node.id, enableDoubleClick: true }}
+                                        onAdditionalValueSave={onEditableSave}
+                                    >
+                                        <div className={'tw-flex'}>{name}</div>
```
imoutaharik, line 178: No need to use the `{}` if you're passing just a string for the tailwind classes.

imoutaharik (changes requested):
Thanks for the accessibility update! It's so much better!! 🎉 
I still think the first story should be the one with default styling and the second one with custom.

I understand the need of having flexible styling here for the metadata form but as a Fondue component, I would still have the feedback of @thisguyscode or Johanna for at least the default styling. Because as it is right now in the stories it looks very premature in terms of design compared to the others. I would prefer to know what they think about it as there is no Figma for me to judge properly.

julianiff:
@imoutaharik thats a great idea, looking forward getting some updated designs for this component. 

For the PR, there were no styling changes introduced in this code (it is already like that in production). How do we want to tackle the new design adjustments? All together or separatly?

src/components/Tree/Node.tsx
julianiff:
```suggestion
                                        <div className='tw-flex'>{name}</div>
```

src/components/EditableInput/lib/helper.ts
julianiff:
```suggestion
import React, {ReactElement} from 'react';
```

src/components/EditableInput/lib/helper.ts
```diff
@@ -0,0 +1,25 @@
...
+export const EditableInputHelper = {
+    /**
+     * Recursively iterators through Children
+     * returns all instances of string amended together
+     *
+     * @param children
+     */
+    getLabel: (children?: React.ReactElement) => {
+        let label = '';
+        const iteratorStep = (children?: React.ReactElement | string) => {
```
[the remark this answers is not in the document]
julianiff:
```suggestion
        const iteratorStep = (children?: ReactElement | string) => {
```

src/components/EditableInput/EditableInput.tsx
```diff
@@ -4,78 +4,161 @@ import React, {
...
+ * @param onAdditionalValueSave
+ * @param children
+ * @param options
```
julianiff, line 71:
```suggestion
```

thisguyscode:
@imoutaharik @julianiff my understanding was that the styling would not be achieved by passing tailwind classes as a property, but rather by taking a text node as a child, and inferring the styling from that.

That would essentially mean that there is NO default styling (for the text), as the text node is a requirement, and text nodes always have styling.

I believe the only attributes we need to infer from the text node are:
- `font-size`
- `letter-spacing`
- `line-height`

julianiff:
@thisguyscode Ah aight, let me give it a try.

julianiff:
@imoutaharik If you find time, i adjusted the component for the Tree and it should now be ready to be reviewed/merged.

imoutaharik (approved):

thisguyscode:
Great work on this one @julianiff and @imoutaharik !! Sorry for the lack of design input, perhaps we can make more use of the alignment blockers in the mornings?

I've let Simone know about this as she was interested in its effect on the Tree. I suggested she take a look and give feedback on next steps for extending the properties, but that the new style will immediately affect the tree and any changes need to be properly considered.

PR #950, opened 2022-09-07
fix(badge)!: update spacing

SamCreasey:
Updated the badge spacing to match the design in Figma https://www.figma.com/file/pevuvey1joCsopTRr9f4OE/Fondue---Components?node-id=13856%3A256660 . This includes a single edge case where spacing is slightly narrower if all 3 components exist on a larger badge.

### DIFF
--- a/src/components/Badge/types.ts
+++ b/src/components/Badge/types.ts
@@ -4,16 +4,20 @@ import { IconProps } from '@foundation/Icon/IconProps';
 import { PropsWithChildren, ReactElement } from 'react';
 import { Color } from '../../types';
 
+export type BadgeSize = 's' | 'm';
+
 export type BadgeProps = PropsWithChildren<{
     style?: BadgeStyle;
     icon?: ReactElement<IconProps>;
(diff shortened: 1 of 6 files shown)


src/components/Badge/types.ts
```diff
@@ -4,31 +4,28 @@ import { IconProps } from '@foundation/Icon/IconProps';
 import { PropsWithChildren, ReactElement } from 'react';
 import { Color } from '../../types';
 
+export type BadgeSize = 's' | 'm';
```
imoutaharik, line 7: Thanks!! Looks good to me 🎉  I would just update that little type to be in full letters. Even if it's quite understandable, just for consistency as you already updated the RGBA. 😄

SamCreasey: Sure, fixed :) Will introduce a breaking change though

imoutaharik: Thanks! No problem so did Jonas by removing the self-alignment on that same component. 😄

imoutaharik (approved):

PR #958, opened 2022-09-13
chore: move sorting logic from out of fondue and allow optional sorting columns

joseazevedo88:
☝️ Now columns are optionally sortable with a prop and the sorting logic is exposed so that it can be customizable, which we need in analytics

image

### DIFF
--- a/src/components/Table/Table.stories.tsx
+++ b/src/components/Table/Table.stories.tsx
@@ -7,8 +7,8 @@ import { Button, ButtonSize, ButtonStyle } from '@components/Button/Button';
 import { IconSize } from '@foundation/Icon/IconSize';
 import { action } from '@storybook/addon-actions';
 import { Meta, Story } from '@storybook/react';
-import React, { FC, useEffect, useState } from 'react';
-import { Column, Row, SelectionMode, Table, TableProps } from './Table';
+import React, { FC, Key, useEffect, useState } from 'react';
+import { Column, Row, SelectionMode, SortDirection, Table, TableProps } from './Table';
 import { IconDotsVertical, IconFaceHappy } from '@foundation/Icon';
 
 export default {
@@ -47,7 +47,7 @@ const ActionButton: FC = () => (
 
 const columns: Column[] = [
     { name: 'User', key: 'user' },
-    { name: 'Active Sessions', key: 'activeSessions' },
+    { name: 'Active Sessions', key: 'activeSessions', sortable: true },
     { name: 'Last Active', key: 'lastActive' },
     { name: 'Regions', key: 'regions' },
     { name: 'Countries', key: 'countries' },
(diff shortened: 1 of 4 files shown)


src/components/Table/Table.stories.tsx
```diff
@@ -211,14 +211,35 @@ const rows: Row[] = [
...
 
     return (
         <Table
             {...args}
             columns={columns}
-            rows={rows}
+            rows={sortedRows}
             selectedRowIds={selectedRows}
             onSelectionChange={(ids) => setSelectedRows(ids || [])}
+            onSortChange={onSortChange}
```
fabriciomendonca, line 242: Should we add the `onSortChange` to the `TemplateWithSearch` story as well?

src/components/Table/Table.tsx
```diff
@@ -66,7 +68,7 @@ type SortType = {
 const mapToTableAriaProps = (columns: Column[], rows: Row[]): TableStateProps<any> => ({
     children: [
         <TableHeader key="table-header" columns={columns}>
-            {(column) => <AriaColumn allowsSorting>{column.name}</AriaColumn>}
+            {(column) => <AriaColumn allowsSorting={column.sortable}>{column.name}</AriaColumn>}
```
fabriciomendonca, line 71: Since the sorting function has to be to set by the component user, wouldn't it be better to only show the sorting arrow and/or allow sorting if the `onSortChange` callback is also provided?

src/components/Table/TableColumnHeader.tsx
```diff
@@ -93,12 +92,20 @@ export const TableColumnHeader = ({
     return (
         <th
             ref={ref}
-            className="tw-text-xs tw-font-medium tw-text-black-100 dark:tw-text-white tw-px-4 tw-py-3 tw-outline-none tw-cursor-pointer tw-group focus-visible:bg-violet-90"
+            className={`tw-text-xs tw-font-medium tw-text-black-100 dark:tw-text-white tw-px-4 tw-py-3 tw-outline-none tw-group focus-visible:bg-violet-90 ${
```
fabriciomendonca, line 95: Just for the sake of consistency, shouldn't we use the `merge` utility here?

fabriciomendonca (commented):
lg2m but I would add a sorting testing case as well. Great work! 💪

src/components/Table/Table.stories.tsx
joseazevedo88: Yup 👌

src/components/Table/Table.tsx
joseazevedo88: Yes, it's being done already on `<TableColumnHeader>`

src/components/Table/TableColumnHeader.tsx
joseazevedo88: Yes we should 👌

fabriciomendonca (approved):
Great work!

PR #977, opened 2022-09-28
feat(switch): Update styling, including sizes

thisguyscode:
@julianiff @imoutaharik - I think everything looks okay from a styling perspective. Open questions are:

- As the size has changed slightly, can we double-check some instances of the switch to see that nothing is breaking? Should be okay so I haven't labeled this as a breaking change for now.

- There is a strange vertical shift when switching between `hug=true`- `hug=false`. This is likely only relevant in storybook, and as the issue existed already before these changes, I haven't attempted to change this, but could be worth a quick look. (FYI it relates to the `flex` - `inline-flex` change).

- I didn't even look at the test files. Maybe you could teach me some time but for now I think I'd only make it worse ;)


julianiff (approved):

julianiff:
FOUND THEM FINALLY! xD To me on my dev env looks good. 

Large Switch - Collection sharing:
https://www.loom.com/share/62b9bfb4a30b40099eb1da78ce2a935f

Medium Switch - Settings: 
https://www.loom.com/share/f484be9b810746779016d54967173239

Small Switch - Settings:
https://www.loom.com/share/2a20f277933e45f4abf6a29f5cef7823

thisguyscode:
@julianiff great! Looks good to me. What do you think?

thisguyscode:
@julianiff Once the deploy preview is built could you double check that last commit please? Then I think it's good to go.

(I was too attached to the animation, but moved it to the active state and it still adds a bit of fun)

PR #980, opened 2022-09-29
fix(Button): remove button react-aria

julianiff:
React-aria always trigger the focusring when clicking the button. Removing react-aria fixes this issue and does not harm any accessibility criterias.


imoutaharik (approved):
I like this PR very much 😄

PR #982, opened 2022-09-30
feat(Tree): expand the tree to see active node(s) on mount

Kenny806:


src/components/Tree/Node.tsx
```diff
@@ -53,6 +53,31 @@ type NodeProps = {
     onEditableSave?: (targetItemId: string, value: string) => void;
 };
 
+const setShowNodesInitial = (currentNode: DraggableItem<TreeNodeItem>, activeNodeIds?: NullableString[]): boolean => {
```
imoutaharik, line 56: Would it be possible to find another name for this function? I think maybe en `setShowNodesOnInit` would be more semantic.

src/components/Tree/Node.tsx
```diff
@@ -53,6 +53,31 @@ type NodeProps = {
...
+        if (node.nodes) {
+            for (const childNode of node.nodes) {
+                if (activeIds.includes(childNode.id)) {
+                    return true;
+                }
+                if (hasActiveChildNodes(childNode, activeIds)) {
+                    return true;
+                }
+            }
+        }
```
imoutaharik, line 67: Also it would be great to simplify this function.

imoutaharik (changes requested):
Thanks for making the changes! 😄 I think it would be great to add stories when it's collapsed and when it's expanded on load. And a test. 😜 WDYT?

src/components/Tree/Node.tsx
Kenny806: What about `getShowNodesOnIt`? Now that I look at it, this function doesn't set a value, but gets one.

Kenny806: Renamed to `getInitialShowNodesValue`, as I realized it's not a setter, but rather a getter. And it seems descriptive, as it is getting the initial value for the `showNodes` state variable. But happy to rename to anything else

src/components/Tree/Node.tsx
Kenny806: Simplified it :-)

src/components/Tree/Tree.cy.tsx
```diff
@@ -7,11 +7,12 @@ import { Tree, TreeFlatListItem } from './Tree';
 
 type ComponentProps = {
     nodes: DraggableItem<TreeFlatListItem>[];
+    activeNodeIds?: NullableString[];
     onDrop?: (sourceItemId: string, parentId: NullableString, positionBeforeId: NullableString) => void;
     onEditableSave?: (itemId: string, value: string) => void;
 };
-const Component: FC<ComponentProps> = ({ nodes, onDrop, onEditableSave }) => {
-    const [selectedIds, setSelectedIds] = useState<NullableString[]>([]);
+const Component: FC<ComponentProps> = ({ nodes, onDrop, onEditableSave, activeNodeIds }) => {
```
imoutaharik, line 14: That would be great to emove `FC` from here too.

Kenny806: Done :-)

imoutaharik (approved):

PR #984, opened 2022-10-05
fix(asset-input): Different button styles

rawi96:
Fix styling issue in asset input block. Both buttons should have the same color and font-weight

### DIFF
--- a/src/components/AssetInput/AssetInput.tsx
+++ b/src/components/AssetInput/AssetInput.tsx
@@ -115,7 +115,7 @@ export const AssetInput: FC<AssetInputProps> = ({
                     data-test-id="asset-input-upload"
                 >
                     <label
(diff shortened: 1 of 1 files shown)


src/components/AssetInput/AssetInput.tsx
```diff
@@ -115,7 +115,7 @@ export const AssetInput: FC<AssetInputProps> = ({
                     data-test-id="asset-input-upload"
                 >
                     <label
-                        className="tw-relative tw-cursor-pointer tw-rounded tw-flex tw-items-center tw-justify-center tw-px-4 tw-h-9 tw-text-s tw-text-black-80 tw-bg-transparent hover:tw-bg-black-10 hover:tw-text-black active:tw-bg-black-20 active:tw-text-black dark:tw-text-white dark:hover:tw-bg-black-95 dark:active:tw-bg-black-superdark dark:hover:tw-text-white"
+                        className="tw-relative tw-cursor-pointer tw-rounded tw-flex tw-items-center tw-justify-center tw-px-4 tw-h-9 tw-text-s tw-text-black-80 tw-bg-transparent hover:tw-bg-black-10 hover:tw-text-black active:tw-bg-black-20 active:tw-text-black dark:tw-text-white dark:hover:tw-bg-black-95 dark:active:tw-bg-black-superdark dark:hover:tw-text-white tw-text-button-text tw-font-medium"
```
bojangles-m, line 118: I see there is some pseudo class 'dark', should that also be adjusted or not? If not, then it is ok :)
Screenshot 2022-10-05 at 19 13 36

bojangles-m (approved):

PR #990, opened 2022-10-11
refactor(Rich Text Editor): Refactor and remove old actions [#32ya11q]

bojangles-m:


src/components/RichTextEditor/utils/parseRawValue.ts
```diff
@@ -26,7 +26,10 @@ export const parseRawValue = ({ editorId = 'parseRawValue', raw }: ParseRawValue
     try {
         parsedValue = JSON.parse(raw);
     } catch {
-        const editor = createPlateEditor({ id: `${editorId}_parseRawValue`, plugins: getEditorConfig() });
+        const editor = createPlateEditor({
+            id: `${editorId}_parseRawValue`,
+            plugins: createPlatePlugins(defaultPlugins),
```
SamCreasey, line 31: Maybe we should be able to pass in an editor itself rather than an editor id. Not sure 100% on the internals of Udecode/plate but it seems like we could parse a raw value incorrectly if we have a different selection of plugins to what is available in the RTE where this function is used, no? But for external use we can create a default editor inside the function

bojangles-m: This needs to have it, because it creates two different editors. We tried to pass it but at guideline blocks went in infinite loop. At the same moment you use two plates and one of them is just to parse the passed value.

src/components/RichTextEditor/RichTextEditor.cy.tsx
```diff
@@ -148,309 +156,322 @@ const RichTextEditorWithValueSetOutside = ({ value }: { value: string }) => {
...
 
-    it('should type and insert text', () => {
-        cy.mount(<RichTextEditor />);
+        it('should type and insert text', () => {
+            cy.mount(<RichTextEditor />);
 
-        cy.get('[contenteditable=true]').click().type('hello');
+            cy.get('[contenteditable=true]').click().type('hello');
+        });
     });
```
SamCreasey, line 214: Add a space before every nested describe to make it easier to read

src/components/RichTextEditor/utils/parseRawValue.ts
SamCreasey: Ah ok, got it

src/components/RichTextEditor/RichTextEditor.cy.tsx
bojangles-m: Updated

SamCreasey (approved):
Nice work, looks a lot cleaner 👍

PR #991, opened 2022-10-11
test(TreeComponent): Improving tree component tests

dominiquemb:


pnpm-lock.yaml
```diff
@@ -0,0 +1,18479 @@
+lockfileVersion: 5.4
```
imoutaharik, line 1: This file shouldn't be added. We use `npm` in Fondue.

PR #1010, opened 2022-10-25
fix(Rich Text Editor): Removed editor id prop from parsing method

bojangles-m:
It is breaking the Clarify test with infinite rendering.
The Fondue beta 10 is working normally the 11 is breaking.

### DIFF
--- a/src/components/RichTextEditor/RichTextEditor.cy.tsx
+++ b/src/components/RichTextEditor/RichTextEditor.cy.tsx
@@ -156,7 +156,7 @@ describe('RichTextEditor Component', () => {
         cy.get(RICH_TEXT_EDITOR).should('contain.text', TEXT);
     });
 
-    it('should be possible to pass and change value', () => {
+    it('wraps the Editor in the component ', () => {
         const TEXT = 'This is new text';
 
         cy.mount(<RichTextEditorWithValueSetOutside value={TEXT} />);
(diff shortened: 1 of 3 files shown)


SamCreasey (approved):
@Walkeezy Seems like we might need an alternative solution to the infinite rendering caused by not supplying the id as it causes other problems, namely that the infinite loop is still visible within tests in Clarify but it's not 100% clear on why yet, only that it is caused by the parseRawValue function.

src/components/RichTextEditor/utils/parseRawValue.ts
```diff
@@ -21,7 +21,7 @@ export const parseRawValue = (editorId: string, raw?: string): Value => {
     try {
         parsedValue = JSON.parse(raw);
     } catch {
-        const editor = createPlateEditor({ id: editorId, plugins: getEditorConfig() });
+        const editor = createPlateEditor({ plugins: getEditorConfig() });
```
Walkeezy, line 24: Do the tests in clarify work when doing something like this?

```suggestion
        const editor = createPlateEditor({ id: editorId + '_parseRawValue', plugins: getEditorConfig() });
```

src/components/RichTextEditor/RichTextEditor.cy.tsx
```diff
@@ -150,6 +156,13 @@ describe('RichTextEditor Component', () => {
         cy.get(RICH_TEXT_EDITOR).should('contain.text', TEXT);
     });
 
+    it('should be possible to pass and change value', () => {
```
Walkeezy, line 159: The description of this test case seems missleading, as the actual test is just passing and not changing the value.

bojangles-m: Updated

Walkeezy (approved):

PR #1013, opened 2022-10-26
fix: provide unique id to createPlateEditor

Walkeezy:
Redo changes from https://github.com/Frontify/fondue/pull/1008, which were reverted in https://github.com/Frontify/fondue/pull/1010. This time it's an unique id for the `parseRawValue`-method and switched to an object for the parameters.

How to test this with clarify (as linking fondue to clarify is somehow not working currently):
- fondue: `npm run build`
- copy folder `dist` from **fondue** to **clarify/node_modules/@frontify/fondue**
- run tests


bojangles-m (approved):

SamCreasey (approved):
Kudos 👍

PR #1017, opened 2022-10-27
feat(Rich Text editor): MD to slate transformer [#2xjpz9h]

bojangles-m:


src/components/RichTextEditor/serializer/markdown/deserializer/plugin.ts
```diff
@@ -1,9 +1,9 @@
 /* (c) Copyright Frontify Ltd., all rights reserved. */
 
-import { MarkdownAstNode, OptionType } from '../types';
+import { MarkdownAstNode, PartialOptionType } from '../types';
 import deserialize from './deserialize';
 
-export default function plugin(opts?: OptionType) {
+export default function plugin(opts?: PartialOptionType) {
```
SamuelAlev, line 6:
```suggestion
export default function plugin(options?: PartialOptionType) {
```
Same for the others files 🙈

src/components/RichTextEditor/utils/exampleValues.ts
```diff
@@ -18,13 +18,16 @@ import { ELEMENT_CHECK_ITEM } from '../Plugins/CheckboxListPlugin/id';
 import { MentionComboboxItem, MentionableCategory } from '../Plugins/MentionPlugin/types';
 import { TextStyles } from '../Plugins/TextStylePlugin/TextStyles';
 import { DesignTokens } from '../types';
+import { mixedMarkdown } from '../serializer/markdown/__tests__/fixtures';
```
SamuelAlev, line 21: Mixed feeling about importing tests data in here
Is it used only in stories? If yes, why not importing it there?

src/components/RichTextEditor/serializer/markdown/options.ts
```diff
@@ -0,0 +1,49 @@
...
+        inline_code_mark: getPluginType(editor, MARK_CODE),
+        strong_mark: getPluginType(editor, MARK_BOLD),
+        emphasis_mark: getPluginType(editor, MARK_ITALIC),
+        delete_mark: getPluginType(editor, MARK_STRIKETHROUGH),
+        image: getPluginType(editor, ELEMENT_IMAGE),
+        code_block: getPluginType(editor, ELEMENT_CODE_BLOCK),
+        thematic_break: getPluginType(editor, ELEMENT_HR),
+        ul_list: getPluginType(editor, ELEMENT_UL),
+        ol_list: getPluginType(editor, ELEMENT_OL),
+        listItem: getPluginType(editor, ELEMENT_LI),
```
SamuelAlev, line 38: There is a mix of casing in the key, please align everything to camelCase

src/components/RichTextEditor/serializer/markdown/deserializer/plugin.ts
bojangles-m: Updated

src/components/RichTextEditor/utils/exampleValues.ts
bojangles-m: No, it is not. it is used in test. So to not create a duplicate and if this works in test it will work in RTE :)

src/components/RichTextEditor/serializer/index.ts
```diff
@@ -0,0 +1,3 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+export * from './transform';
```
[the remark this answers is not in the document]
bojangles-m: Yes we do, you give the transformer what king of transformation you want to do. And transformer gives the rules what it needs so you have to create it for the interface. You pass the MD to slate or Slate to MD. I would prefer if we have the same for HTML serializer ;)

`export class Transform {
    public static use(transformer: Transformer) {
        return transformer;
    }
}
`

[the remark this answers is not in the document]
bojangles-m: So we are consistent in the transformers we have.

src/components/RichTextEditor/serializer/markdown/options.ts
bojangles-m: Updated

src/components/RichTextEditor/serializer/markdown/deserializer/plugin.ts
bojangles-m: Updated

PR #1024, opened 2022-11-02
feat(Tooltip): add enterDelay prop

dominiquemb:
The optional enterDelay prop defines the amount of time that should pass before the tooltip becomes visible. If not defined, the tooltip will render immediately.

https://app.clickup.com/t/32yf5y9


src/components/Tooltip/Tooltip.cy.tsx
```diff
@@ -50,6 +53,40 @@ describe('Tooltip Component', () => {
         cy.get(TOOLTIP_ID).should('contain', TOOLTIP_TEXT);
     });
 
+    it('should render a tooltip after a timeout if enterDelay is set to a number greater than 0', () => {
+        initTooltip({ content: TOOLTIP_TEXT, enterDelay: 1000 });
```
fabriciomendonca, line 57: You could use the `ENTER_DELAY` const here

fabriciomendonca (changes requested):

fabriciomendonca (approved):

PR #1026, opened 2022-11-04
refactor(Rich Text Editor): Fix type errors

imoutaharik:
Fixing Type errors listed when running `npm run typecheck`
Because of the complexity and not being able to easily define a type and using `unknown` is not an option, type `any` will be used.

The decision being that is preferable to have a lint warning for `no-explicit-any` than a type error.

### DIFF
--- a/src/components/RichTextEditor/serializer/utils/serializeLeafToHtml.ts
+++ b/src/components/RichTextEditor/serializer/utils/serializeLeafToHtml.ts
@@ -11,7 +11,7 @@ import {
 import escapeHtml from 'escape-html';
 
 export const serializeLeafToHtml = (node: TDescendant): string => {
-    let string = escapeHtml(node.text);
+    let string = escapeHtml(node.text as string);
     if (node.bold) {
         string = `<span class="${BOLD_CLASSES}">${string}</span>`;
     }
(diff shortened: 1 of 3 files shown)


src/components/RichTextEditor/serializer/utils/serializeLeafToHtml.ts
```diff
@@ -11,7 +11,7 @@ import {
 import escapeHtml from 'escape-html';
 
 export const serializeLeafToHtml = (node: TDescendant): string => {
```
bojangles-m, line 13: instead of TDescendant make any

bojangles-m: and also in file serializeNodeToHtmlRecursive.ts where prop is passed

bojangles-m (changes requested):

src/components/RichTextEditor/serializer/utils/serializeLeafToHtml.ts
imoutaharik: Thanks, just replaced `TDescendant` for `any`

bojangles-m (approved):

PR #1030, opened 2022-11-07
feat(Menu): add new menu component and use existing MenuItem

fabriciomendonca:


src/components/Menu/Menu.tsx
```diff
@@ -0,0 +1,89 @@
...
+interface Props {
+    triggerRef?: RefObject<Element>;
+    open?: boolean;
+    children?: ReactElement<MenuItemProps> | ReactElement<MenuItemProps>[];
+    onClose?: () => void;
+}
+
+export type MenuProps = PropsWithChildren<Props>;
+
+const CONTAINER_BASE_CLASSES = 'tw-relative tw-bg-base tw-rounded tw-py-2 tw-shadow-mid tw-w-max';
```
joseazevedo88, line 17: To me this indicates once again, that we need to have a reusable building blocks generic component, in this case, container.

src/components/Menu/Menu.tsx
```diff
@@ -0,0 +1,89 @@
...
+const CONTAINER_POSITION_CLASSES = ' tw-left-0 tw-top-7';
```
joseazevedo88, line 18: Is the at the whitespace needed in the beginning of the string? Also, is this menu supposed to be used with any trigger? The `top: 7` doesn't allow that, but maybe I'm misunderstanding here :)

src/components/Menu/Menu.tsx
```diff
@@ -0,0 +1,89 @@
...
+const INSET_BORDER =
```
joseazevedo88, line 19: Shouldn't this live in the theme?

src/components/MenuItem/MenuItem.tsx
```diff
@@ -82,25 +83,61 @@ export const MenuItem: FC<MenuItemProps> = ({
 
     const textClass = menuItemTextColorRecord[style][textState];
 
+    const [mainElementType, setMainElementType] = useState('span');
+
+    useEffect(() => {
+        if (link || onClick) {
+            setMainElementType(getItemElementType(link, onClick));
+        }
+    }, [link, onClick]);
```
joseazevedo88, line 92: Do we need this? Seems we could do with: 
`const elementType = getItemElementType(link, onClick)`

src/components/MenuItem/MenuItem.tsx
```diff
@@ -1,28 +1,21 @@
 /* (c) Copyright Frontify Ltd., all rights reserved. */
 
-import { MenuItemContent, MenuItemContentProps, MenuItemContentSize } from '@components/MenuItem/MenuItemContent';
+import React, { FC, MouseEvent, useEffect, useState } from 'react';
```
imoutaharik, line 3: Could you please remove `FC` here please. thanks.

src/components/MenuItem/MenuItemContent.tsx
```diff
@@ -1,22 +1,12 @@
 /* (c) Copyright Frontify Ltd., all rights reserved. */
 
+import React, { FC, HTMLAttributes, ReactElement, ReactNode, cloneElement } from 'react';
```
imoutaharik, line 3: Same here with `FC`

imoutaharik (commented):
Thanks! I made some comments and have a general question: Shouldn't the menu close when you click outside?

src/components/Menu/Menu.tsx
fabriciomendonca: I agree with you. This is a specific one with the shadow and an inset shadow for the border, which I had to use a pseudo element to make it possible. But I totally get your point.
It would be great have some basic reusable elements.

src/components/Menu/Menu.tsx
fabriciomendonca: You are right, I will remove that because popper is positioning the element. And the whitespace shouldn't be there at all. Thanks for the catch.

src/components/Menu/Menu.tsx
fabriciomendonca: This is something I have planned but the Flyout wasn't merged already when I created this. I also want to create a helper function to add pseudo prefixes to a set of classes.
Like:
```
addPseudoCssPrefix('before', 'tw-pointer-events-none tw-block tw-rounded tw-top-0');
```
Would return
```
'before:tw-pointer-events-none before:tw-block before:tw-rounded tw-top-0'
```

src/components/MenuItem/MenuItem.tsx
fabriciomendonca: Having the `useEffect` would avoid any unnecessary call to `getItemElementType`. Since it is dependent of the `link` and `onClick` props which are used only on the new implementation of `MenuItem`, a side effect hook here is the proper way to go, imho.

src/components/Menu/Menu.stories.tsx
```diff
@@ -0,0 +1,34 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+import React from 'react';
+import { Meta, Story } from '@storybook/react';
```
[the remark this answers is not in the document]
fabriciomendonca: Great. Thanks for pointing it out. :)

src/components/Menu/Menu.cy.tsx
```diff
@@ -0,0 +1,304 @@
...
+    cy.focused().trigger('keydown', { key: 'Tab' });
+    cy.focused().should('have.attr', 'href', ITEMS_WITH_LINKS[2].link);
+
+    closeMenuWithKeyboard(['Tab'], hasTrigger);
+};
+
+const buttonItemsKeyboardNavigation = (hasTrigger = true) => {
+    openWithKeyboard(hasTrigger);
+    verifyButtonItem(1);
+    cy.focused().trigger('keydown', { key: 'ArrowDown' });
```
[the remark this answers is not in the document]
fabriciomendonca: Somehow, Cypress interpret the `realPress` and the `trigger` differently. I tried to use the `realPress` there and the tests were failing. You can see on the close function that when we have a TriggerElement we should use `realPress`, without that element, we should use `.trigger`. Weird, but it is Cypress...

joseazevedo88:
> @imoutaharik
> 
> > Thanks! I made some comments and have a general question: Shouldn't the menu close when you click outside?
> > Thanks for reviewing. This was created from the `OverflowMenu` implementation and I can't remember why we don't have it implemented there. I will work on this. Do you know if we have a custom hook for handling the click outside?

Yup, there's a `useClickOutside` hook in fondue 😄

src/components/Menu/Menu.tsx
```diff
@@ -0,0 +1,89 @@
...
+        strategy: 'fixed',
+        modifiers: [
+            {
+                name: 'flip',
+                enabled: true,
+            },
+            {
+                name: 'offset',
+                options: {
+                    offset: [0, 8],
```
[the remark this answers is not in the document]
fabriciomendonca: Nice one, will take a look how it is done there

src/components/MenuItem/MenuItem.tsx
fabriciomendonca: Sure :)
With the new changes we also need to update our local node version to 18.
I believe that worths an announcement on Slack. WDYT?

src/components/Menu/Menu.tsx
fabriciomendonca: Actually, we don't need the `CONTAINER_POSITION_CLASSES` anymore. I will just remove it. That's why I like so separate TW classes in some variables. It is too hard to catch those cases when we are dealing with a super long string.

src/components/MenuItem/MenuItem.tsx
imoutaharik: I plan to make the announcement today.

src/components/Menu/Menu.tsx
fabriciomendonca: Added to the utilities. The helper function we can implement later if needed since we are using prefixes everywhere

joseazevedo88 (approved):
Didn't test but code looks good 🎉

src/components/MenuItem/MenuItem.tsx
```diff
@@ -1,28 +1,21 @@
 /* (c) Copyright Frontify Ltd., all rights reserved. */
```
imoutaharik, line 2: I just noticed that the styling of disabled menuItem on the dropdown is rounded now.

Screenshot 2022-11-24 at 10 57 21

imoutaharik (changes requested):

src/components/MenuItem/MenuItem.tsx
fabriciomendonca: Fixed. Idk why I added a `tw-rounded` class to the `MenuItem` container. Probably on the first tests and I forgot to remove 🤦‍♂️ Thank you so much for catching this 💪

imoutaharik (approved):

PR #1040, opened 2022-11-09
fix(dropdown): add alias tokens and remove brand tokens

julianiff:
Hey,

I changed the dropdown component to now use only alias tokens. Hope everything looks like before but with other tokens.

- To test its best to add the tw-dark class to the body of the iframe in the storybook, because the portal moves the html end of the document and is not covered by a tw-dark.

### DIFF
--- a/src/components/Dropdown/Dropdown.tsx
+++ b/src/components/Dropdown/Dropdown.tsx
@@ -122,7 +122,7 @@ export const Dropdown: FC<DropdownProps> = ({
 
     const textColorClass = activeItem
         ? menuItemTextColorRecord[activeItem.style || MenuItemStyle.Primary][textState]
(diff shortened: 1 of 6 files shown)


fabriciomendonca (approved):
It looks good to me. We could create a doc with the mapping for the substitutions somewhere. So it would be easier for other people also to replace the brand tokens by the alias ones. WDYT?
Like:
`tw-text-black-60 -> tw-text-text-x-weak`

julianiff:
> It looks good to me. We could create a doc with the mapping for the substitutions somewhere. So it would be easier for other people also to replace the brand tokens by the alias ones. WDYT? Like: `tw-text-black-60 -> tw-text-text-x-weak`

Sounds like a great idea!

PR #1049, opened 2022-11-14
fix(Flyout): missing outline and rounded corners

dominiquemb:


src/components/ColorPicker/ColorPreview.tsx
```diff
@@ -15,8 +15,8 @@ export const ColorPreview: FC<ColorPreviewProps> = ({ color }) => {
 
     return (
         <div
-            className="tw-sticky tw-box-content tw-shrink-0 tw-h-2 tw-top-0 tw-bg-white tw-z-20 dark:tw-bg-black-95"
-            style={{ backgroundColor }}
+            className="tw-sticky tw-box-content tw-shrink-0 tw-h-2 tw-top-0 tw-bg-white tw-z-20 dark:tw-bg-black-95 tw-shadow-inner tw-border-l tw-border-t tw-border-r tw-border-line"
+            style={{ backgroundColor, borderTopLeftRadius: '3px', borderTopRightRadius: '3px' }}
```
fabriciomendonca, line 19: Is there any special reason why we are using 3px here?
The Flyout has a `4px` rounded corners... Maybe we should use 4px here too, shouldn't we?

thisguyscode:
@dominiquemb - the border is missing from the top edge in regular flyout
Screenshot 2022-11-14 at 16 31 32

src/components/ColorPicker/ColorPreview.tsx
dominiquemb: This was added inline because I was having problems with tw-rounded-t borders being more rounded than the outer container's borders, but this seems to have resolved itself with the latest changes, so I've removed the inline CSS.

dominiquemb:
> @dominiquemb - the border is missing from the top edge in regular flyout Screenshot 2022-11-14 at 16 31 32

Fixed in last commit

thisguyscode (dismissed):

src/components/Flyout/FlyoutFooter.tsx
```diff
@@ -14,7 +14,7 @@ export const FlyoutFooter: FC<FlyoutFooterProps> = ({ buttons, border = true })
     return (
         <div
             className={merge([
-                'tw-flex tw-gap-x-3 tw-justify-end tw-py-5 tw-px-8 tw-bg-white dark:tw-bg-black-95',
+                'tw-flex tw-gap-x-3 tw-justify-end tw-py-5 tw-px-8 tw-bg-white dark:tw-bg-black-95 tw-rounded-bl tw-rounded-br tw-shadow-mid tw-border-l tw-border-b tw-border-r tw-border-line',
```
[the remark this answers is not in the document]
dominiquemb: This class has been removed

thisguyscode:
Screenshot 2022-11-16 at 14 26 35
@dominiquemb - sorry I missed this in the screenshot in slack, but the shadow should apply to the whole flyout's outer edge.

So when I said it shouldn't be on the footer I just meant the top edge of the footer.

Overall I think the approach of applying the shadow to the body and footer independently is just generally going to cause challenges. There should be one shadow applied to the whole flyout. Please refer to Figma

https://www.figma.com/file/cU8YFwN6fqiEK2sM7i6ffo/Fondue---Flyout?node-id=10%3A14048&t=ZK6PxrWChB1o1jOa-0

dominiquemb:
> Screenshot 2022-11-16 at 14 26 35
> 
> @dominiquemb - sorry I missed this in the screenshot in slack, but the shadow should apply to the whole flyout's outer edge.
> So when I said it shouldn't be on the footer I just meant the top edge of the footer.
> 
> Overall I think the approach of applying the shadow to the body and footer independently is just generally going to cause challenges. There should be one shadow applied to the whole flyout. Please refer to Figma
> 
> https://www.figma.com/file/cU8YFwN6fqiEK2sM7i6ffo/Fondue---Flyout?node-id=10%3A14048&t=ZK6PxrWChB1o1jOa-0

I have tried playing around with it, and it seems that it's not possible for the shadow to apply to the whole Flyout without seeing the bug where the shadow disappears as you make the screen shorter than the height of the Flyout, in cases when there is a minContentHeight. 

This is why the header, body, and footer of the Flyout need to have their own TW shadow classes.

thisguyscode:
@dominiquemb - I see the issue, react-aria is adding a `max-height` to the entire overlay, which is hard to control.

For now I would still move the shadow to the outer element (as in commit), but maybe @imoutaharik can clean all this up with her scroll work.

As far as I can tell, anyone currently using contentMinHeight may have buttons that are entirely inaccessible outside of the viewport, as they can't scroll to them.

dominiquemb:
> @dominiquemb - I see the issue, react-aria is adding a `max-height` to the entire overlay, which is hard to control.
> 
> For now I would still move the shadow to the outer element (as in commit), but maybe @imoutaharik can clean all this up with her scroll work.
> 
> As far as I can tell, anyone currently using contentMinHeight may have buttons that are entirely inaccessible outside of the viewport, as they can't scroll to them.

I have re-added the `tw-shadow-mid` class to the outer container. Please have a look now.

thisguyscode:
@dominiquemb - I think this is the right compromise. The shadow is broken when using ContentMinHeight but I think that the flyout is generally broken when using ContentMinHeight because the user is currently unable to scroll the buttons into view.

@imoutaharik FYI to bear this mind when working on scroll bug fix. We should also ensure the shadow renders correctly with ContentMinHeight set.

fabriciomendonca:
@thisguyscode Just added to the Flyout refactor task description
https://app.clickup.com/t/30zjuba

src/components/ColorPicker/ColorPreview.tsx
```diff
@@ -15,7 +15,7 @@ export const ColorPreview: FC<ColorPreviewProps> = ({ color }) => {
 
     return (
         <div
-            className="tw-sticky tw-box-content tw-shrink-0 tw-h-2 tw-top-0 tw-bg-white tw-z-20 dark:tw-bg-black-95"
+            className="tw-sticky tw-box-content tw-shrink-0 tw-h-2 tw-top-0 tw-bg-white tw-z-20 dark:tw-bg-black-95 tw-border-t tw-mt-[-1px] tw-border-l tw-border-r tw-border-line tw-rounded-t"
```
fabriciomendonca, line 18: `tw-border-t tw-mt-[-1px] tw-border-l tw-border-r` could be `tw-mt-[-1px] tw-border tw-border-b-0`
Maybe it would be easier to read, WDYT?

src/components/Flyout/FlyoutFooter.tsx
```diff
@@ -14,7 +14,7 @@ export const FlyoutFooter: FC<FlyoutFooterProps> = ({ buttons, border = true })
     return (
         <div
             className={merge([
-                'tw-flex tw-gap-x-3 tw-justify-end tw-py-5 tw-px-8 tw-bg-white dark:tw-bg-black-95',
+                'tw-flex tw-gap-x-3 tw-justify-end tw-py-5 tw-px-8 tw-bg-white dark:tw-bg-black-95 tw-rounded-bl tw-rounded-br tw-border-l tw-border-b tw-border-r tw-border-line',
```
fabriciomendonca, line 17: Same here
`tw-border-l tw-border-b tw-border-r` could be also `tw-border tw-border-t-0`

fabriciomendonca (approved):

PR #1052, opened 2022-11-14
fix: paragraph crash

SamCreasey:
Fixes an issue where the defaults for a paragraph were not correctly applied to the plugin


Walkeezy (approved):
👍🏻

PR #1059, opened 2022-11-17
chore: upgrade to React 18

imoutaharik:
**Done:**
- Upgrade React and React types to v18
- Update React-aria: [PR](https://github.com/Frontify/fondue/pull/1062)
- Fixed `./jsx-runtime.js is not exported from package` error by updating `react-dnd`: [PR](https://github.com/Frontify/fondue/pull/1061)

Test Fondue(React17) on React 18 app -> Working
Test Fondue(React18) on React 17 app -> Type Errors (a lot)
Test Fondue(React18) on React 18 app -> Working

!!! Waiting for upgrade in Clarify !!!


package.json
```diff
@@ -163,7 +163,7 @@
         "xstate": "4.28.1"
     },
     "peerDependencies": {
-        "react": "^17.0.2",
-        "react-dom": "^17.0.2"
+        "react": "^16.8 || 17 || 18",
```
jonasbuechel, line 166: let's remove 16.x as we never test it

jonasbuechel (changes requested):

src/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -28,7 +28,7 @@ export type BreadcrumbsProps = {
     items: Breadcrumb[];
 };
 
-export const Breadcrumbs: FC<BreadcrumbsProps> = ({ items }) => {
```
SamuelAlev, line 31: When removing `FC<Props>`, don't forget to set the return type of the function to `ReactElement` otherwise we might return boolean and it won't complain

src/components/Tooltip/Tooltip.tsx
```diff
@@ -205,7 +205,7 @@ export const Tooltip = ({
     }, [enterDelay]);
 
     const checkIfHovered = useCallback(
-        (event) => {
+        (event: any) => {
```
SamuelAlev, line 208:
```ts
      const checkIfHovered = useCallback(
        (event: MouseEvent) => {
            const hoveredElement = event.composedPath?.();
            const hoverSources = [triggerRefElement, triggerElementContainerRef, tooltipContainerRef];

            if (hoveredElement && hoverSources.some((el) => el.current && hoveredElement.includes(el.current))) {
                handleShowTooltipOnHover();
            }
        },
        [handleShowTooltipOnHover],
    );
```
You can do this

src/components/Breadcrumbs/Breadcrumbs.tsx
imoutaharik: Interesting! I guess I have to add the return type in every component `FC` have been removed. Might need to update the Fondue docs as well. And the template when running `npm run component:create`.

src/components/Tooltip/Tooltip.tsx
imoutaharik: Thanks for the suggestion! I just updated that one. 😄

src/components/Breadcrumbs/Breadcrumbs.tsx
imoutaharik: Done! Added in some components and I created a task in clickup to remove `FC` in remaining components still using it. So this PR doesn't get too big.

imoutaharik:
@jonasbuechel @SamuelAlev I've updated and fixed the type errors, test are passing as well. We just have sonarLint failing because of the Flyout stories. I tried to find a way to fix it but didn't succeed. If you have any ideas, I'm all ears. 😃 
Otherwise, we also have a full Flyout refactoring planned. 
I'm requesting a re review if you don't mind. Thanks!

jonasbuechel:
Closed because we already covered type adaptions.

PR #1062, opened 2022-11-18
chore(deps): upgrade react-aria

imoutaharik:
Upgrading all React-aria dependencies.


fabriciomendonca (approved):
I checked all components and they seem to be working as before. Thanks for handling this :)

PR #1063, opened 2022-11-21
chore: update to node 18

SamuelAlev:
- Update to node 18 by default as it is LTS
- Update Storybook to next (it is quite stable and close to release) to fix the TLS deprecation coming with node 17
  - Update config to ESM 
  - Replace `Story` with `StoryFn` as it is getting deprecated ([migration docs](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#story-type-deprecated))

**To test:**
Update to node 18 (`nvm install 18 && nvm use` in the root of Fondue) and `npm ci`, then you can try:
- `npm run build`
- `npm run build:storybook`
- `npm run storybook`


package.json
```diff
@@ -29,7 +29,7 @@
         "build": "npm run clean && npm run generate:icons && npm run build:library && npm run build:extractedThirdPartyStyles",
         "build:extractedThirdPartyStyles": "concat src/components/**/*.css | postcss > dist/extractedThirdPartyStyles.css",
         "build:library": "vite build",
-        "build:storybook": "NODE_OPTIONS=--max_old_space_size=6144 build-storybook",
+        "build:storybook": "storybook build",
```
imoutaharik, line 32: We added `NODE_OPTIONS=--max_old_space_size=6144` because we were running out of memory in huge PR like re rendering all icons. Should we delete it? 🤔

src/components/Accordion/Accordion.stories.tsx
```diff
@@ -33,7 +33,7 @@ export default {
     component: AccordionComponent,
 } as Meta<AccordionProps>;
 
-export const WithDifferentAccordionItems: Story<AccordionProps> = () => {
+export const WithDifferentAccordionItems: StoryFn<AccordionProps> = () => {
```
imoutaharik, line 36: Controls are looking funky, stuck in dark mode 🌚 
Screenshot 2022-11-22 at 10 38 11

imoutaharik (commented):
The components are working well, I just noticed a few things in Storybook.

The Accessibility add-on is not working for me:
Screenshot 2022-11-22 at 10 40 39

package.json
SamuelAlev: Let's add it back if it bugs us?

src/components/Accordion/Accordion.stories.tsx
SamuelAlev: Fixed

package.json
imoutaharik: Yes, we can do that!

imoutaharik (approved):

src/components/Checklist/Checklist.tsx
```diff
@@ -36,7 +36,6 @@ type ChecklistVertical = ChecklistBase & {
 
 type ChecklistHorizontal = ChecklistBase & {
     direction: ChecklistDirection.Horizontal;
-    columns?: undefined;
```
imoutaharik: @SamuelAlev Are we sure we don't need the columns anymore?

SamuelAlev: Yup, sonarcloud was complaining so I had to fix it

PR #1065, opened 2022-11-21
fix(TextInput): tabbing should not highlight two elements at the same time

dominiquemb:


fabriciomendonca (approved):
Great work with the new test cases :)

PR #1069, opened 2022-11-23
fix: rte space wrapping

SamCreasey:
Fixes an issue caused by the regex test not accounting for attributes on the HTML elements, (eg. style attributes set by the design tokens). This would cause a string like `<p style="color:red">text</p>` to get wrapped in tags and become `<p><p style="color:red">text</p></p>` Instead checking if the element has any child nodes is a safe method of assessing whether the content is a plain string or not


Walkeezy:
Nice 👍🏻

Walkeezy (approved):

PR #1070, opened 2022-11-23
fix(DatePicker): clear button was not clearing the input field

dominiquemb:


src/components/TextInput/TextInput.tsx
```diff
@@ -128,7 +128,11 @@ export const TextInput: FC<TextInputProps> = ({
                     : merge([
                           'focus-within:tw-border-black-90 hover:tw-border-black-90',
                           validationClassMap[validation],
-                          isFocusVisible && FOCUS_STYLE,
+                          isFocusVisible &&
```
fabriciomendonca, line 131: We should avoid on next branches to branch from the last one you worked. Those changes were already merged to `main` on the `TextInput` PR. If you are branching from `main` I believe it would be better to have a fresh installation, because it can lead to some conflicts in the future. :)

fabriciomendonca (approved):

PR #1072, opened 2022-11-23
fix: Error validation state icon, add validation state to Dropdown, MultiSelect, and DatePicker( + updated clearing )

valokin:


src/components/MultiSelect/MultiSelect.cy.tsx
```diff
@@ -32,13 +34,18 @@ const ITEMS = {
     ],
 };
 
-const Component: FC = () => {
+type Props = {
+    validation?: Validation;
+};
+
+const Component: FC<Props> = ({ validation = Validation.Default }: Props) => {
```
imoutaharik, line 41: I think we can remove the `FC<Props>`.

src/components/TextInput/TextInput.tsx
```diff
@@ -218,6 +226,14 @@ export const TextInput: FC<TextInputProps> = ({
                     <LoadingCircle size={LoadingCircleSize.ExtraSmall} />
                 </span>
             )}
+            {validation === Validation.Error && (
```
imoutaharik, line 229: I think this works pretty well now that we have just the error icon in place but we will need soon to have the 3 options, Icon for success and loading. Can we format that to switch the different state easily.
Maybe `validation !== Validation.Default` and render inside the according icon and styling. something like that, WDYT?

src/components/DatePicker/DatePickerTrigger.tsx
```diff
@@ -12,28 +13,74 @@ type DatePickerTriggerProps = PropsWithChildren<{
...
+        };
+
+        const carrotOffsetForClearButton = value && isClearable ? 1 : 0;
+        const carrotOffsetForValidationErrorIcon = isErrorValidationState ? 1 : 0;
+
+        const carrotRightSideTWPositionClass = carrotRightSideTWPositionClassByOffset(
+            carrotOffsetForClearButton + carrotOffsetForValidationErrorIcon,
+        );
+
+        /* eslint-disable jsx-a11y/click-events-have-key-events,jsx-a11y/no-static-element-interactions */
```
imoutaharik, line 50: There is no way we can fix this accessibility issue and not disable eslint? 🤔

imoutaharik (commented):
Thanks for taking care of it! great stuff! 😄 
I just left a few comments.

src/components/DatePicker/DatePickerTrigger.tsx
valokin: I wouldn't add any keyboard interaction to this element as it's already possible to open the the DatePicker by hitting enter when you focus the whole input. Making this element focusable would make it look strange.

The purpose of this element is to handle clicks for opening when the clear icon (or clear icon and error validation icon both ) are present, but leaving out space for the clear icon to be clickable as well.

src/components/MultiSelect/MultiSelect.cy.tsx
valokin: Done

src/components/TextInput/TextInput.tsx
valokin: While I see the valid point of having the code more ready for the next change. I think that change(`validation !== Validation.Default`) should be made when the time comes to implement the loading and success state icons there, otherwise the `render inside the according icon and styling` expression would return nothing in the two other cases. 
Also, the loading state exists atm and it shows the loader on the top right corner. Meaning, the current actual version of only having the error state icon positioned in this place would be less explicit with `validation !== Validation.Default` as the condition.

imoutaharik (approved):

imoutaharik:
@valokin Just noticed a small styling issue here with the focus ring on the clear button.
Screenshot 2022-11-24 at 16 13 52

valokin:
> 


Let me explain why I added the **white background to the arrow icon**. In the case when the DatePicker has a small width, the text for the date value and the arrow overlap, 
Screenshot 2022-11-24 at 16 45 06

But now I see this seems to be the issue even before my changes, so I guess I we should remove the white background from the icon?
Screenshot 2022-11-24 at 16 47 26


And should we put a 
1. minimum width on the DatePicker component to ensure no UI overlap
or
2. let the user of the component make sure the width of the container is never too small?

imoutaharik:
> > 
> 
> Let me explain why I added the **white background to the arrow icon**. In the case when the DatePicker has a small width, the text for the date value and the arrow overlap, Screenshot 2022-11-24 at 16 45 06
> 
> But now I see this seems to be the issue even before my changes, so I guess I we should remove the white background from the icon? Screenshot 2022-11-24 at 16 47 26
> 
> And should we put a
> 
> 1. minimum width on the DatePicker component to ensure no UI overlap
>    or
> 2. let the user of the component make sure the width of the container is never too small?

Hello,
Yes, we still don't have anything responsive in Fondue, Not everything is working on small devices or even small sizes.

Thanks for removing the white background 🙌🏼 , I believe it's the best option so it's stay the way it was before. and we don't have that weird styling on focus.

For the minimum width, I would say option 2 and let user decide. But would love to know what you think about it @thisguyscode

imoutaharik (approved):

PR #1074, opened 2022-11-24
feat (Rich Text Editor): emoji plugin

bojangles-m:
Inline plugin which activates with ':' and two chars.

https://www.loom.com/share/bab6b9d8ded044558ec4db0cfa87218f


src/components/RichTextEditor/Plugins/EmojiPlugin/EmojiCombobox/EmojiComboboxStyles.ts
```diff
@@ -0,0 +1,45 @@
...
+            z-index: 500;
+        `,
+    ],
+    item: [
+        ...item,
+        `
+            background: transparent;
+
+            :hover {
+                background: rgb(234, 235, 235, 1);
```
SamuelAlev, line 31: You could use `var()` with fondue tokens instead of hardcoded value

bojangles-m: Updated

SamuelAlev (approved):

PR #1075, opened 2022-11-24
fix(Dropdown): onChange is firing twice

imoutaharik:


mikeyrayvon (changes requested):
Thanks for looking into this :) 

The `onChange` callback still fires twice like this. `onChange()` is called by `onSelectionChange` on `line 94`, and this gets called once when selecting an item from the dropdown, and again when `state.setSelectedKey` is called on `line 118` in this `useEffect`. 

I think actually this `useEffect` is redundant, because when I log `state` in the callback instead of calling `setSelectedKey`, the `state.selectedKey` is already set correctly. 

Try it like this:
```
useEffect(() => {
        if (state.disabledKeys.has(activeItemId as string)) {
            return;
        }

        console.log(state);

        //state.setSelectedKey(activeItemId as string);
    }, [activeItemId]);
 ```
 
 My suggestion is we take out this `useEffect` altogether. But I'm not sure the full intention of it, so maybe this has unintended consequences... Thoughts?

imoutaharik:
@mikeyrayvon It's still firing twice?

I actually reverted a recent change I did with the dependencies of that `useEffect`, I added `state` and it was causing to fire twice.

I tried it by logging in the onChange function directly on the Story:
`return <Dropdown {...args} activeItemId={active} onChange={(id) => console.log(id)} />;`

It was firing twice and when I removed the `state` dependency, it was firing only once.

Now maybe there is some refactoring to be done on the component, but I'm not sure what would the full consequences of removing the `useEffect` and why it was implemented in the first place.

The dropdown is the same now than a release ago, we could dig deeper in the issue and improve it but we also plan to create a new `Select` Component with the new Inputs Design that's going to replace the `Dropdown`.

imoutaharik:
@mikeyrayvon Yes, you were right, I tested in Clarify with a dropdown used in the marketplace and it was firing twice. Now only once when I removed the useEffect. But I'm wondering why it was added and if it would break somewhere if removed.

@charlietango You added it like a year ago, (i know it's been a long time ago) but do you recall why and tell me if it's ok that we remove it.

mikeyrayvon (approved):
Works as expected with useEffect removed. Thanks :)

charlietango:
@imoutaharik IIRC I've done that change so the active item can be set from a prop.

PR #1076, opened 2022-11-25
feat(Dropdown): add emphasis

Walkeezy:
This PR adds the `emphasis` prop to the `Dropdown` component, as we need it for the new Code Snippet block.

### DIFF
--- a/src/components/Dropdown/Dropdown.stories.tsx
+++ b/src/components/Dropdown/Dropdown.stories.tsx
@@ -95,7 +95,7 @@ const DropdownWithBackgroundTemplate: StoryFn<DropdownProps> = (args: DropdownPr
 
     return (
         <div className="tw-p-8 tw-bg-black-5 tw-flex tw-gap-2">
-            <div className="tw-max-w-[300px]">
+            <div className="tw-max-w-[350px]">
                 <Dropdown
                     {...args}
                     placeholder="select default item"
@@ -104,7 +104,7 @@ const DropdownWithBackgroundTemplate: StoryFn<DropdownProps> = (args: DropdownPr
                     onChange={(id) => setActive(id)}
                 />
             </div>
-            <div className="tw-max-w-[300px]">
+            <div className="tw-max-w-[350px]">
                 <Dropdown
                     {...args}
                     placeholder="select weak item"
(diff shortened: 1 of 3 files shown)


src/components/Dropdown/Dropdown.stories.tsx
```diff
@@ -80,6 +87,36 @@ const DropdownWithinOverflownContainer: StoryFn<DropdownProps> = (args: Dropdown
...
 
+const DropdownWithBackgroundTemplate: StoryFn<DropdownProps> = (args: DropdownProps) => {
+    const [active, setActive] = useState(args.activeItemId);
+    const [active2, setActive2] = useState(args.activeItemId);
+    useEffect(() => setActive(args.activeItemId), [args.activeItemId]);
+    useEffect(() => setActive2(args.activeItemId), [args.activeItemId]);
+
+    return (
+        <div className="tw-p-8 tw-bg-black-5 tw-flex tw-gap-2">
+            <div className="tw-max-w-[350px]">
```
imoutaharik, line 98: For the story here, I would remove the default emphasis dropdown and have the week one a bit larger than 350px

src/components/Trigger/Trigger.tsx
```diff
@@ -14,6 +14,11 @@ export enum TriggerSize {
     Large = 'Large',
 }
 
+export enum TriggerEmphasis {
```
imoutaharik, line 17: Instead of having 2 types with the same values, I would create a general one `InputEmphasis`, as this new property is going to be used in `Combobox`, `Select` and `MultiSelect`

imoutaharik (changes requested):
Thanks for the contribution! I just left a couple of comments 😄

src/components/Dropdown/Dropdown.stories.tsx
Walkeezy: Makes sense, removed.

src/components/Trigger/Trigger.tsx
Walkeezy: Changed 👍🏻

imoutaharik (approved):
Great! Thanks for making the changes 😄

PR #1077, opened 2022-11-25
feat(multiselect): add multiSelect weak emphasis

joseazevedo88:
New variant for asset audit trail in `MultiSelect` component. This component should still work retroactively

image

### DIFF
--- a/src/components/Menu/Menu.cy.tsx
+++ b/src/components/Menu/Menu.cy.tsx
@@ -59,9 +59,10 @@ const closeMenuWithKeyboard = (keys: any, hasTrigger = true) => {
 const openWithKeyboard = (hasTrigger = true) => {
     if (hasTrigger) {
         cy.get(MENU_SELECTOR).find('button').eq(0).focus().realPress('Enter');
-    } else {
-        cy.get('body').trigger('keydown', { key: 'Tab' });
     }
+    // else {
+    // cy.get('body').trigger('keydown', { key: 'Tab' });
+    // }
 
     cy.get(MENU_SELECTOR).find('nav').should('have.class', 'tw-block');
 };
--- a/src/components/Menu/Menu.cy.tsx
+++ b/src/components/Menu/Menu.cy.tsx
@@ -59,10 +59,9 @@ const closeMenuWithKeyboard = (keys: any, hasTrigger = true) => {
 const openWithKeyboard = (hasTrigger = true) => {
     if (hasTrigger) {
         cy.get(MENU_SELECTOR).find('button').eq(0).focus().realPress('Enter');
+    } else {
+        cy.get('body').focus();
     }
-    // else {
-    // cy.get('body').trigger('keydown', { key: 'Tab' });
-    // }
 
     cy.get(MENU_SELECTOR).find('nav').should('have.class', 'tw-block');
 };
--- a/src/components/Menu/Menu.tsx
+++ b/src/components/Menu/Menu.tsx
@@ -14,19 +14,21 @@ interface Props {
     children?: ReactElement<MenuItemProps> | ReactElement<MenuItemProps>[];
     onClose?: () => void;
     offset?: [number, number];
+    type?: 'default' | 'multiselect';
 }
 
 export type MenuProps = PropsWithChildren<Props>;
 
-const CONTAINER_BASE_CLASSES = 'tw-relative tw-bg-base tw-rounded tw-py-2 tw-shadow-mid tw-min-w-max tw-z-[120000]';
-const CONTAINER_POSITION_CLASSES = ' tw-left-0 tw-top-7';
+const CONTAINER_BASE_CLASSES = 'tw-relative tw-bg-base tw-rounded tw-py-2 tw-shadow-mid tw-z-[120000]';
+const CONTAINER_POSITION_CLASSES = 'tw-left-0 tw-top-7';
 
 const CONTAINER_CLASSES = merge([CONTAINER_BASE_CLASSES, INSET_BORDER]);
 
-export const Menu = ({ triggerRef, children, open = true, offset = [0, 8], onClose }: MenuProps) => {
+export const Menu = ({ triggerRef, children, open = true, offset = [0, 8], onClose, type = 'default' }: MenuProps) => {
     const [isMenuOpen, setIsMenuOpen] = useState(open);
     const [menuContainerRef, setMenuContainerRef] = useState<HTMLElement | null>(null);
     const [menuOpenerRef, setMenuOpenerRef] = useState<HTMLElement | null>(null);
+    const MIN_WIDTH_CLASS = type === 'default' ? 'tw-min-w-max' : 'tw-min-w-[18rem]';
 
     const menuKeyboardNavigationAction = useMenuKeyboardNavigation(
         isMenuOpen,
@@ -87,6 +89,7 @@ export const Menu = ({ triggerRef, children, open = true, offset = [0, 8], onClo
         <nav
             className={merge([
                 CONTAINER_CLASSES,
+                MIN_WIDTH_CLASS,
                 menuOpenerRef ? CONTAINER_POSITION_CLASSES : '',
                 isMenuOpen ? 'tw-block' : 'tw-hidden',
             ])}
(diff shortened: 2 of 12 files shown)


src/components/Menu/Menu.tsx
```diff
@@ -18,7 +18,7 @@ interface Props {
 
 export type MenuProps = PropsWithChildren<Props>;
 
-const CONTAINER_BASE_CLASSES = 'tw-relative tw-bg-base tw-rounded tw-py-2 tw-shadow-mid tw-w-max';
+const CONTAINER_BASE_CLASSES = 'tw-relative tw-bg-base tw-rounded tw-py-2 tw-shadow-mid tw-min-w-max';
```
fabriciomendonca, line 21: Just saw a problem with the `Menu` z-index I let pass during the development. You can see on the component Docs page on Storybook by opening the Basic Menu With Links. It will be behind the Menu Without Trigger section.
Adding the class `tw-z-[120000]` to the base classes will solve the issue (this is the same class applied on the `Tooltip` component)

src/components/Menu/Menu.cy.tsx
```diff
@@ -59,9 +59,10 @@ const closeMenuWithKeyboard = (keys: any, hasTrigger = true) => {
 const openWithKeyboard = (hasTrigger = true) => {
     if (hasTrigger) {
         cy.get(MENU_SELECTOR).find('button').eq(0).focus().realPress('Enter');
-    } else {
-        cy.get('body').trigger('keydown', { key: 'Tab' });
     }
+    // else {
+    // cy.get('body').trigger('keydown', { key: 'Tab' });
```
fabriciomendonca, line 64: This block was supposed to add focus to the menu without a trigger. I saw now it is auto focusing the first element by default because of the keyboard navigation hook.
I was thinking here: maybe this can be an issue if in a page that uses the `Menu` without a trigger, the keyboard navigation will start on that menu first item. 
WDYT about changing the keyboard navigation hook to not autofocus it if no trigger element?
cc: @imoutaharik

src/components/Trigger/Trigger.tsx
```diff
@@ -14,7 +14,7 @@ export enum TriggerSize {
     Large = 'Large',
 }
 
-export enum InputEmphasis {
+export enum TriggerEmphasis {
```
fabriciomendonca, line 17: This will add a breaking change, if it is being used outside of Fondue.
I checked Clarify and couldn't find any usage of it.
Also checked that it was created by @Walkeezy [here](https://github.com/Frontify/fondue/pull/1076/files).
Idk if he is using outside of Fondue, so it would be nice to double check if these changes will not break anything before merging them.

Walkeezy: Going to use it in Frontify/guideline-blocks. It's fine to rename it from my side. :-)

joseazevedo88: Nice, thanks. Let's do `TriggerEmphasis` for consistency then 👍

thisguyscode:
@joseazevedo88 There are quite a few interaction issues that I think need addressing before we release. I don't think that a comment is going to do it really. Happy to do a loom but best would be a meeting.

src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -1,17 +1,21 @@
...
-import { mergeProps } from '@react-aria/utils';
 import { merge } from '@utilities/merge';
 import { Validation } from '@utilities/validation';
 import { AnimatePresence, motion } from 'framer-motion';
-import React, { FC, useRef, useState } from 'react';
-import { getPaddingClasses } from './helpers';
+import React, { FC, KeyboardEvent, useRef, useState } from 'react';
+import { getInputWidth, getPaddingClasses } from './helpers';
+import { Menu } from '@components/Menu';
+import { MenuItem } from '..';
```
fabriciomendonca, line 16: maybe we should use from `@components/MenuItem` for consistency?

src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -96,72 +144,174 @@ export const MultiSelect: FC<MultiSelectProps> = ({
...
+    const handleSpacebarToggle = (e: KeyboardEvent<HTMLDivElement>) => {
+        if (e.code === 'Space' && (filterInputRef.current !== document.activeElement || !filterable)) {
+            toggleOpen();
+        }
+    };
+
+    const getDecoratorClasses = () => {
+        if (emphasis === MultiSelectEmphasis.Weak) {
+            return hasSelectedItems ? '' : 'tw-text-text-weak';
+        } else {
```
fabriciomendonca, line 156: you can remove the `else` statement here and just use `return ''` since it will return inside the `if`

src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -96,72 +144,174 @@ export const MultiSelect: FC<MultiSelectProps> = ({
...
+    const getTagType = () => {
+        if (emphasis === MultiSelectEmphasis.Weak) {
+            return TagType.Selected;
+        } else {
+            if (open) {
+                return TagType.SelectedWithFocus;
+            } else {
+                return TagType.Selected;
+            }
+        }
```
fabriciomendonca, line 170: I believe you can simplify this logic, WDYT?

```suggestion
        if (emphasis === MultiSelectEmphasis.Weak || !open) {
            return TagType.Selected;
        }

        return TagType.SelectedWithFocus;
```

src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -47,33 +73,55 @@ export const MultiSelect: FC<MultiSelectProps> = ({
...
+
+    const handleClickOutside = () => {
+        if (emphasis === MultiSelectEmphasis.Weak) {
+            setOpen(false);
+        }
+    };
+
+    useClickOutside(menuRef?.current, handleClickOutside, [multiSelectRef?.current as HTMLElement]);
+
+    const toggleOpen = () => setOpen((open) => !open);
```
fabriciomendonca, line 107: The `ESC` key when focusing on the menu was supposed to close the menu but it is not working here. You can see it working on the `Menu` component stories

fabriciomendonca (changes requested):

joseazevedo88:
@thisguyscode this is not helpful. First, I don't know who you are. Second, I'm not part of fondue team and my task is to quickly prototype something so we can release and then fondue team can iterate over it and accomplish the original design vision. Therefore, this will need to be merged fast so we can get rid of fondue dependencies on my(analytics) side. Third, if there are broken interactions with the previous implementation kindly let me know via comment, or loom, so I can address them. Thanks

thisguyscode:
@joseazevedo88 Hey, it's Jamie :D

Didn't mean to offend you or anything, just though it might be easier to jump on a call, and if you were not available for a call then I could record a loom.

I'll record a loom.

joseazevedo88:
Hey @thisguyscode Jamie 👋 😄  No offense taken, I was just wondering if who commented had the whole context 😅 Please also sync with Nikolai on it, having the input inside a button is an a11y violation(and just code in general). Also in this case(see 1st screenshot) how can the user know there's an input there? That's why I added the possibility to pass a label to that input(see 2nd screenshot). I'm up for a meeting to discuss this if needed.
image

image

src/components/MultiSelect/MultiSelect.tsx
fabriciomendonca: It is not about saving two lines of code. It is about avoiding future unreachable code if any written after the `else`. And It would be ok if you decided not to change it now. I know there is more than one solution for a problem, and it is great when we use this space for at least pointing them and learning from each other. I'm sorry if it looked I was only trying to save two lines of code.

thisguyscode:
Bugs/regressions:
https://www.loom.com/share/c5eb1dcf6b144adebb37f8d57a13dc95
https://www.loom.com/share/9416ded66ae9404f9e138e9bcdae38ca
https://www.loom.com/share/a2f3b5a2a5d642c2b5c0497be89019b7

UX suggestion:
https://www.loom.com/share/e07847f51d404dbe82d69515940d6486

There may be more but those ones make it hard to fully assess from a usability standpoint.

fabriciomendonca (approved):
Approved after a peer review with @joseazevedo88

PR #1085, opened 2022-12-02
fix(Divider): adding test cases for all props

dominiquemb:


src/components/Divider/Divider.cy.tsx
```diff
@@ -1,23 +1,73 @@
...
+        cy.mount(<Divider style={DividerStyle.Solid} />);
+
+        cy.get(DIVIDER_HR_SELECTOR).should('exist').and('have.class', 'tw-border-solid');
+    });
+
+    it('allows the divider border style to be solid', () => {
+        cy.mount(<Divider style={DividerStyle.Solid} />);
+
+        cy.get(DIVIDER_HR_SELECTOR).should('exist').and('have.class', 'tw-border-solid');
+    });
```
fabriciomendonca, line 66: This case is duplicated, right?

fabriciomendonca (changes requested):

src/components/Divider/Divider.cy.tsx
dominiquemb: Yup, fixed

fabriciomendonca (approved):

PR #1090, opened 2022-12-07
fix(RTE): toolbar border is visible on zero toolbar width

fulopdaniel:
image


bojangles-m (commented):
I would suggest to update GeneratePlugins and pluginComposer.
The plugin composer gets the noToolbar prop, which can be used inside GeneratePlugins to show or hide toolbar. Could that work?

bojangles-m (approved):

PR #1098, opened 2022-12-12
fix: add rules to  RTE serializer for bold, italic, underline

SamCreasey:
Html strings that have been created by `serializeRawToHtml` were not accurately parsing the bold, italic, and underline plugins due to the fact that the default plugin rules do not check for the format we have created for each markup element. 
Adding a new rule means that it will be picked up by the RTE.
To test, copy the below into the RichTextEditor Storybook. 
On main branch this will appear as a plain paragraph, and here it will have all 3 styles applied
```
const VAL =
    '<p style="font-size: 14px; font-style: normal; font-weight: normal;"><span class="tw-font-bold"><span class="tw-underline"><span class="tw-italic">text</span></span></span></p>';

export const SerializerTest = (args: RichTextEditorProps) => (
    <RichTextEditorComponent
        {...args}
        plugins={new PluginComposer().setPlugin([
            new InitPlugin(),
            new ItalicPlugin(),
            new BoldPlugin(),
            new UnderlinePlugin(),
        ])}
        value={VAL}
    />
);

```


src/components/RichTextEditor/Plugins/BoldPlugin/createBoldPlugin.ts
```diff
@@ -0,0 +1,15 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+import { ToggleMarkPlugin, createBoldPlugin as createPlateBoldPlugin, createPluginFactory } from '@udecode/plate';
+import { BOLD_CLASSES } from './BoldMarkupElement';
+
+export const createBoldPlugin = createPluginFactory<ToggleMarkPlugin>({
```
bojangles-m, line 6: I think you dont need to create createPluginFactory because the Plugin already exist you can just call the createBoldPlugin and add object with props like: 
```
createBoldPlugin({
    deserializeHtml: {
            rules: [
                { validNodeName: ['SPAN'], validClassName: BOLD_CLASSES },
            ],
        },
    }
})
```
the default values should be overwritten with yours. Let me know

SamCreasey: fixed

src/components/RichTextEditor/Plugins/ItalicPlugin/createItalicPlugin.ts
```diff
@@ -0,0 +1,17 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+import { ToggleMarkPlugin, createItalicPlugin as createPlateItalicPlugin, createPluginFactory } from '@udecode/plate';
+import { ITALIC_CLASSES } from './ItalicMarkupElement';
+
+const plateItalicPlugin = createPlateItalicPlugin();
+
+export const createItalicPlugin = createPluginFactory<ToggleMarkPlugin>({
```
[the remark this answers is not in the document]
SamCreasey: fixed

src/components/RichTextEditor/Plugins/UnderlinePlugin/index.ts
```diff
@@ -2,10 +2,21 @@
 
 import { createUnderlinePlugin } from '@udecode/plate';
 import { UNDERLINE_PLUGIN } from './id';
-import { UnderlineMarkupElement } from './UnderlineMarkupElement';
+import { UNDERLINE_CLASSES, UnderlineMarkupElement } from './UnderlineMarkupElement';
 import { UnderlineButton } from './UnderlineButton';
 import { Plugin, PluginProps } from '../Plugin';
 
+const PLATE_UNDERLINE_RULES = [
+    {
```
bojangles-m, line 10: Is this necessary if you overwrite it afterwards?

bojangles-m: And why dont you put everything in the `PLATE_UNDERLINE_RULES `?

bojangles-m (approved):

src/components/RichTextEditor/Plugins/UnderlinePlugin/index.ts
SamCreasey: I added the const there to explicitly say which rules were coming from plate and not us, since I thought there should be some degree of separation between the rules so it's easier to tell which rule is important for the markup element

PR #1118, opened 2022-12-30
test(Tooltip): add test cases for all props

fabriciomendonca:


src/components/Tooltip/Tooltip.cy.tsx
```diff
@@ -191,4 +198,111 @@ describe('Tooltip Component', () => {
         cy.get(TOOLTIP_ID).should('exist');
         cy.get(TOOLTIP_ID).should('have.class', 'tw-hidden');
     });
```
dominiquemb, line 200: I think there should be a test added for the content prop itself.

src/components/Tooltip/Tooltip.cy.tsx
```diff
@@ -191,4 +198,111 @@ describe('Tooltip Component', () => {
...
+
```
dominiquemb, line 201: Also missing test a case for triggerElement

dominiquemb (changes requested):
See comments inline

src/components/Tooltip/Tooltip.cy.tsx
fabriciomendonca: The first test case has only the `content` prop in it, so it is already there. I will update the title to `should render a tooltip with content`.
BTW, `content` is the only required prop.

dominiquemb (approved):

PR #1121, opened 2023-01-02
fix(trigger): fix padding on small variant

joseazevedo88:
The height wasn't matching the new design, here, should be 36px instead of 42px: https://www.figma.com/file/WfEbqZI0nnPWHN43MCtvnm/Fondue---Inputs?node-id=43%3A32279&t=1JXIVp1n0yuj7EAa-4

CC @SamCreasey 

image


SamCreasey (changes requested):
Screenshot 2023-01-03 at 09 37 01
No padding means on a multi-line the input looks wrong

SamCreasey (approved):

PR #1132, opened 2023-01-06
chore(Slider): Making activeItemId optional

selimovicz:


SamCreasey (approved):
Nice change 👍

thisguyscode:
@SamCreasey Nice one, this also solves the funky animation on first render I think.

PR #1137, opened 2023-01-09
chore: new plate 19.0.3

bojangles-m:


src/components/RichTextEditor/Plugins/EmojiPlugin/EmojiPicker/EmojiPicker.tsx
```diff
@@ -54,13 +54,13 @@ export const EmojiPicker = ({
...
             <EmojiPickerContent
                 i18n={i18n}
-                setEmoji={setEmoji}
                 emojiLibrary={emojiLibrary}
                 isSearching={isSearching}
                 searchResult={searchResult}
                 visibleCategories={visibleCategories}
-                selectEmoji={selectEmoji}
                 settings={settings}
+                onSelectEmoji={onSelectEmoji}
```
fabriciomendonca, line 62: Won't these changes be breaking changes since we are changing the props names?

bojangles-m: No, it is internal and we are not using it anywhere yet. So it wont break anything

fabriciomendonca (approved):

PR #1140, opened 2023-01-09
feat!: add optional rich text border

SamCreasey:
Separate border from Position classes so that a border can be added optionally for RichTextEditors with a floating toolbar. This also adds a hover class to the border when it not in readonly mode

BREAKING CHANGE: 
The border will always be enabled by default, whereas previously if the toolbar position was set to `Position.Floating` the border would not be shown.


src/components/RichTextEditor/EditorPositioningWrapper/PlateWrapper.tsx
```diff
@@ -1,17 +1,23 @@
...
+import { Position, WrapperProps } from './types';
 import { EditorPositioningWrapper } from '.';
 import { useRichTextEditorContext } from '../context/RichTextEditorContext';
+import { merge } from '@utilities/merge';
 
 export const PlateWrapper = forwardRef<HTMLDivElement, WrapperProps>(({ children }, ref) => {
-    const { position } = useRichTextEditorContext();
+    const { position, border } = useRichTextEditorContext();
 
-    const classNames = EditorPositioningWrapper[position].PlateWrapperClassNames;
```
bojangles-m, line 11: My suggestion, put the logic into the context and exchange position with classNames so that context return you only classNames instead of position and border and you get this independent.

src/components/RichTextEditor/EditorPositioningWrapper/index.ts
```diff
@@ -10,15 +10,15 @@ import {
 export { Position } from './types';
 export const EditorPositioningWrapper: EditorPositioningWrapperProps = {
```
bojangles-m, line 11: This you can make function which gets two params border and position and returns the classes you need.

src/components/RichTextEditor/context/RichTextEditorContext.tsx
```diff
@@ -10,28 +10,28 @@ import { EditorResizeContextProvider } from './EditorResizeContext';
 export type RichTextEditorContextProps = {
     designTokens: DesignTokens;
     position: Position;
+    border: boolean;
 };
 
 const RichTextEditorContext = createContext<RichTextEditorContextProps>({
     designTokens: defaultDesignTokens,
     position: Position.FLOATING,
```
bojangles-m, line 18: Here you need to adjust the right return from context.

src/components/RichTextEditor/EditorPositioningWrapper/index.ts
SamCreasey: fixed

src/components/RichTextEditor/context/RichTextEditorContext.tsx
SamCreasey: fixed

src/components/RichTextEditor/EditorPositioningWrapper/PlateWrapper.tsx
SamCreasey: fixed

bojangles-m:
Thnx and great job 🥳👏🏻 What do you think is it possible to make that we control border from outside with this border prop. So that I still have menu on top or bottom but without border. Why would this be necessary? in such cases where i need to set some other border maybe rounded on top but not on bottom like in pic.
Screenshot 2023-01-17 at 09 02 30

bojangles-m (commented):
Everything look great, just one question left.

src/components/RichTextEditor/RichTextEditor.stories.tsx
```diff
@@ -57,7 +57,8 @@ export default {
         readonly: false,
         clear: false,
         position: Position.FLOATING,
-        padding: PaddingSizes.None,
+        padding: PaddingSizes.Medium,
```
bojangles-m, line 60:
```suggestion
        padding: Object.keys(PaddingSizes)[2],
```

bojangles-m (approved):
Look great, thnx, just one suggestion maybe, that right radio button is active

PR #1143, opened 2023-01-10
feat(TextInput): Introducing selectable and onFocus props

selimovicz:


imoutaharik (commented):
Thanks for that! That looks good to me. I think it would be nice to also add a story for that behaviour. 😄

imoutaharik (approved):

PR #1146, opened 2023-01-10
chore(multiselect): add prop to pass custom avatar components to menu item

joseazevedo88:
image

### DIFF
--- a/src/components/MultiSelect/MultiSelect.stories.tsx
+++ b/src/components/MultiSelect/MultiSelect.stories.tsx
@@ -194,3 +194,17 @@ WithMixedValues.args = {
     summarizedLabel: 'Mixed values',
     indeterminateItemKeys: ['This is a long tag'],
 };
+
+export const WithAvatar = MultiSelectTemplate.bind({});
+
+WithAvatar.args = {
+    emphasis: MultiSelectEmphasis.Weak,
+    items: WithFilter.args.items,
+};
+
+WithAvatar.argTypes = {
+    emphasis: {
+        options: [],
+        control: { type: 'none' },
+    },
+};
--- a/src/components/MultiSelect/MultiSelect.stories.tsx
+++ b/src/components/MultiSelect/MultiSelect.stories.tsx
@@ -11,7 +11,7 @@ import {
     MultiSelectSize,
     MultiSelectType,
 } from './MultiSelect';
-import { IconPerson16 } from '@foundation/Icon';
+import { IconNook16, IconPerson16 } from '@foundation/Icon';
 
 export default {
     title: 'Components/Multi Select',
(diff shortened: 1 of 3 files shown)


charlietango (commented):
Looks good to me 👍🏻 I recommend you wait for @fabriciomendonca's approval since he's more familiar with the codebase.

imoutaharik (commented):
The code looks good to me but I can't see it in Storybook in the stories. How is it supposed to look like? I don't see the avatar, just the checkbox. 🤔

joseazevedo88:
> The code looks good to me but I can't see it in Storybook in the stories. How is it supposed to look like? I don't see the avatar, just the checkbox. 🤔

To check this variant in storybook, you need to select the `withFilter` story and the `weak` emphasis in the props 👍

imoutaharik:
> > The code looks good to me but I can't see it in Storybook in the stories. How is it supposed to look like? I don't see the avatar, just the checkbox. 🤔
> 
> To check this variant in storybook, you need to select the `withFilter` story and the `weak` emphasis in the props 👍

I see, thanks! Then I believe we should have this option on both emphasis.
It doesn't makes sense to have it only when the emphasis is weak. This feature is hidden like this.
I would also add a story for that -> `WithAvatar` `WithImage`.
The new design for the multiSelect is actually like that with the option to have an icon, an image on the left and the checkmark on the right. But we also want to support checkboxes as some team needs it.
We can quickly discuss that on a call if you want.

joseazevedo88:
> > > The code looks good to me but I can't see it in Storybook in the stories. How is it supposed to look like? I don't see the avatar, just the checkbox. 🤔
> > 
> > 
> > To check this variant in storybook, you need to select the `withFilter` story and the `weak` emphasis in the props 👍
> 
> I see, thanks! Then I believe we should have this option on both emphasis. It doesn't makes sense to have it only when the emphasis is weak. This feature is hidden like this. I would also add a story for that -> `WithAvatar` `WithImage`. The new design for the multiSelect is actually like that with the option to have an icon, an image on the left and the checkmark on the right. But we also want to support checkboxes as some team needs it. We can quickly discuss that on a call if you want.

Thanks! I did it like that as there is two different issues: support the old multiselect, where I didn't change any code in order to not break previous functionality and the new multiselect with the `weak` emphasis where I did the changes. I don't think it should work for the legacy one as I don't think that's a future requirement, but correct me if I'm wrong. What I think might make sense in order to fix this is to have a story `withImage` where the emphasis is `weak` to show the  new multiSelect but don't allow changing the emphasis prop. What do you think?

imoutaharik:
> > > > The code looks good to me but I can't see it in Storybook in the stories. How is it supposed to look like? I don't see the avatar, just the checkbox. 🤔
> > > 
> > > 
> > > To check this variant in storybook, you need to select the `withFilter` story and the `weak` emphasis in the props 👍
> > 
> > 
> > I see, thanks! Then I believe we should have this option on both emphasis. It doesn't makes sense to have it only when the emphasis is weak. This feature is hidden like this. I would also add a story for that -> `WithAvatar` `WithImage`. The new design for the multiSelect is actually like that with the option to have an icon, an image on the left and the checkmark on the right. But we also want to support checkboxes as some team needs it. We can quickly discuss that on a call if you want.
> 
> Thanks! I did it like that as there is two different issues: support the old multiselect, where I didn't change any code in order to not break previous functionality and the new multiselect with the `weak` emphasis where I did the changes. I don't think it should work for the legacy one as I don't think that's a future requirement, but correct me if I'm wrong. What I think might make sense in order to fix this is to have a story `withImage` where the emphasis is `weak` to show the new multiSelect but don't allow changing the emphasis prop. What do you think?

That would be nice to display that story to see that option is actually available for the weak emphasis, yes. 🙏🏼 
 I'll then approve 🚀  because I think there is still a few things we need to decide for that component and I don't want to block this. From Jamie's feedback, we apparently need to support both (checkboxes and new design) and they need to update the Figma file. I'll clarify that with Johanna as it's still not clear and final. anyway I will let you know if major changes will be applied that may affect you in the near future. Does it sound good to you?

joseazevedo88:
Sounds great :) Added the `withAvatar` story with only the `weak` emphasis possible

imoutaharik (approved):

PR #1152, opened 2023-01-13
feat(Rich Text Editor): check for local value

bojangles-m:
Refactor and moved config into useEditorState.
Implemented check if the local value is empty and exposed this method.


src/components/RichTextEditor/RichTextEditor.tsx
```diff
@@ -43,11 +44,18 @@ export const RichTextEditor: FC<RichTextEditorProps> = ({
...
+    const { localValue, onChange, memoizedValue, config } = useEditorState({
+        editorId,
+        initialValue,
+        onTextChange,
+        plugins,
+        onValueChanged,
+    });
 
-    const editableProps: EditableProps = {
+    const editableProps = useRef<EditableProps>({
```
SamuelAlev, line 58: Why the `useRef`?

src/components/RichTextEditor/helpers/isLocalValueEmpty.ts
```diff
@@ -0,0 +1,10 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
```
SamuelAlev, line 1: File should be within the markdown folder

src/components/RichTextEditor/helpers/isLocalValueEmpty.ts
```diff
@@ -0,0 +1,10 @@
...
+
+import { SlateToMarkdown, Transform } from '../serializer';
+import { TreeOfNodes } from '../types';
+
+export const isLocalValueEmpty = (value: TreeOfNodes): Promise<boolean> =>
+    new Promise((resolve) => {
+        const result = Transform.use<TreeOfNodes, string>(new SlateToMarkdown()).process(value);
+        resolve(result.length === 0);
+    });
```
SamuelAlev, line 10: Why `Promise`?
```suggestion
export const isLocalValueEmpty = (value: TreeOfNodes): boolean => {
    const result = Transform.use<TreeOfNodes, string>(new SlateToMarkdown()).process(value);
    return result.length === 0;
}
```

src/components/RichTextEditor/RichTextEditor.tsx
```diff
@@ -28,10 +28,11 @@ export type RichTextEditorProps = {
     padding?: PaddingSizes;
     position?: Position;
     plugins?: PluginComposer;
-    onKeyDown?: (event: KeyboardEvent<HTMLDivElement>, value: TreeOfNodes | null) => void;
```
SamuelAlev, line 31: Breaking change to remove the `value`

src/components/RichTextEditor/helpers/isLocalValueEmpty.ts
SamuelAlev: If `process` is not async then wrapping it with a `Promise` has not effect

bojangles-m: But hen you call it it has effect because you are using it like 
```
isLocalValueEmpty().then() ... 
```
and it wont stop the process nevertheless how long it makes the transformation.
Or do I understand it wrong?

src/components/RichTextEditor/serializer/markdown/isValueEmpty.ts
```diff
@@ -0,0 +1,10 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+import { SlateToMarkdown, Transform } from '..';
+import { TreeOfNodes } from '../../types';
+
+export const isValueEmpty = (value: TreeOfNodes): Promise<boolean> =>
+    new Promise((resolve) => {
+        const result = Transform.use<TreeOfNodes, string>(new SlateToMarkdown()).process(value);
+        resolve(result.length === 0);
```
SamuelAlev, line 9: Can't you check if there is an array of one node and if first node is empty?
It would be way faster

bojangles-m: Updated with new Function :)

SamuelAlev (approved):

PR #1158, opened 2023-01-17
fix: modify tooltip to enable Keyboard Navigation

SamCreasey:
Portal has been removed from the tooltip component. Although it provided some useful escaping from the current stacking context, it seems that this currently isn't really an issue, and for all cases on Clarify this can be resolved with using the 'fixed' style of popper's placement.
What is more important is that the element exists within the correct tab order, and positioning it after the element ensures that, however in order to be tabbable is must exist in the DOM, and switch to a visible style when an element inside the tooltip is focused. To do this, the visibility is controlled instead by opacity and setting the height/width to 0 when the tooltip is not currently visible.

Since the tooltip must always be rendered, there is no longer a difference between `disabled` and `hidden` and so one of them should be deprecated (hidden). 

Note: This pr does not solve the issue of describing the tooltip trigger with an aria-label, or using aria-described by, it just enables keyboard accessibility


src/components/Tooltip/Tooltip.tsx
```diff
@@ -47,6 +48,7 @@ export type TooltipProps = PropsWithChildren<{
     enterDelay?: number;
     open?: boolean;
     disabled?: boolean;
+    /** @deprecated use disabled */
```
fabriciomendonca, line 51: Maybe we could have a more detailed explanation here? `use disabled since the tooltip is always present in the DOM now so hidden has no effect anymore` or something like that... Not necessary tho. WDYT?

fabriciomendonca (approved):

src/components/Tooltip/Tooltip.tsx
SamCreasey: Good call, Updated it!

PR #1167, opened 2023-01-20
feat: WIP Overlays Refactoring

imoutaharik:
### First Draft

- We decided not to refactor the Flyout but keep it as legacy component. We'll take a different approach and create a more flexible Dialog Component. This way, we can have a smoother transition and not a big breaking change to deal with at once as the Flyout is widely used.

Link to the new Design: https://www.figma.com/file/cU8YFwN6fqiEK2sM7i6ffo/Fondue---Flyout?node-id=10%3A14048&t=2QSqMpLi4Ocz6ZMf-1

## Proposed usage
```
<Dialog position={DialogPosition.TopStart} ...etc >
  <DialogHeader
      title="Whatever Header Title"
      close={true}
      size={DialogHeaderSize.Large}
      onClose={() => someFunction()}
   />
   <DialogContent padding={DialogContentPadding.Default}>
     Whatever content you pass as children
   <DialogContent />
   <DialogFooter buttons={[
                            { label: 'Secondary Button', action: () => someFunction() },
                            {
                                label: 'Primary Button',
                                action: () => someFunction(),
                            },
                        ]}
  <DialogFooter />
<Dialog />
```

## Still Needs to be done / Open questions:
**Dialog**
- Still need to add the overflow to scroll content
- Click outside: Add `transactional` property to determine if the dialog should close when clicking outside.
- Should `height`, `min-height`, `max-height` be exposed?
- positioning: Add `triggerRef` prop and handle popper.js positioning logic in the `Dialog` component -> options to flip, offset, (arrow?)

**Content**
Is `DialogContent`/ `DialogBody` really necessary? In the design it just have a padding prop.

**Header**
Use of the Header is optional, devs can also build a custom header to nest inside the Dialog if necessary.
- Add decorator and badge. Most likely limit to only 1 badge and only 1 icon.
`Icon - title - badge`

**Footer**
- Buttons limited to max 2.
- Styling is limited to 1 Secondary and 1 Primary. (Could be useful to add styling options to choose between danger, positive, etc... but set the emphasis within the component).

Sorry for the long text 😆


imoutaharik:
@leko91 @GoranRibic I'm just getting started but would love to have early feedback as I'm progressing on this.
Thanks a lot.

src/components/DialogFooter/DialogFooter.tsx
```diff
@@ -0,0 +1,43 @@
...
+import React, { ReactElement } from 'react';
+import { Button, ButtonEmphasis, ButtonSize, ButtonStyle } from '@components/Button';
+
+export type DialogButton = {
+    label: string;
+    action: () => void;
+};
+
+export type DialogFooterProps = {
+    buttons: [DialogButton] | [DialogButton, DialogButton];
```
leko91, line 12: Any reason why cant this just be a `DialogButton[]`? I know `cancel` and `confirm` are used in 99% of cases, but I don't feel like it would hurt anything. And then just do `buttons.map(...)`.
This is just a nitpick, but I love the direction it's going to for now 😄

imoutaharik: Yes, It was a design decision to limit to only 2 but they weren't against having more. In this case, I would need to give the option to choose the styling of the button. 🤔

thisguyscode:
@imoutaharik I don't know how this fits in with you work so far, but I suggest a slightly different composition:

- `Flyout` or `Popover` for the generic inline overlay element
- `Modal` becomes more generic too (i.e. empty, without all the content properties)
- `DialogHeader`, `DialogBody` and `DialogFooter` can be used inside either of those

thisguyscode:
> Is DialogContent/ DialogBody really necessary? In the design it just have a padding prop.

@imoutaharik - I think it's useful in allowing us to handle scroll behaviour consistently. Even if we don't support much in the beginning it would be helpful to be able to target if we choose to.

imoutaharik:
> > Is DialogContent/ DialogBody really necessary? In the design it just have a padding prop.
> 
> @imoutaharik - I think it's useful in allowing us to handle scroll behaviour consistently. Even if we don't support much in the beginning it would be helpful to be able to target if we choose to.

@thisguyscode Yes, it makes sense. I'll update Dialog to be called popover and add the DialogBody to have the scroll behaviour there.
I may have some coding time tomorrow morning to work on it. 
But TBH, I think changes on the modal should be done on another PR. I would rather introduce the new ones in this one and let gradually refactor and replace in other components.

I still want to keep the Flyout as legacy for smooth transition. So it's not a huge breaking change to deal with.

src/components/DialogHeader/DialogHeader.tsx
```diff
@@ -0,0 +1,60 @@
...
+import React, { ReactElement } from 'react';
+import { merge } from '@utilities/merge';
+import { IconCross } from '@foundation/Icon';
+import { Badge, BadgeProps } from '@components/Badge';
+
+export type DialogHeaderProps = {
+    title: string;
+    size?: DialogHeaderSize;
+    collapseBottom?: boolean;
+    close?: boolean;
```
GoranRibic, line 12: Having `close` and `onClose` like this is not good IMO. 

Reason: What if the user puts close to true but does not provide the `onClose` function? 

What you should do is either only have onClose which is optional and a display button once that is provided, or you need to tightly couple those two props in such a way that if a user passes `close={true}` then `onClose` should be mandatory.

src/components/DialogHeader/DialogHeader.tsx
```diff
@@ -0,0 +1,60 @@
...
+    onClose?: () => void;
+    icon?: ReactElement;
+    badge?: BadgeProps;
```
GoranRibic, line 15: Why not pass the `badge` prop as a component just like an `icon`? This introduces confusion when consuming components, sticking to one pattern removes it. And the better one IMO is by passing component.

src/components/DialogHeader/DialogHeader.tsx
```diff
@@ -0,0 +1,60 @@
...
+};
+
+export enum DialogHeaderSize {
+    Null = 'Null',
```
GoranRibic, line 19: Null?

src/components/DialogHeader/DialogHeader.tsx
```diff
@@ -0,0 +1,60 @@
...
+    Large = 'Large',
+}
+
+export const DialogHeader = ({
+    title,
+    collapseBottom = false,
```
GoranRibic, line 25: collapseBottom? What is the point of this?

GoranRibic (changes requested):
IMHO there are some conceptual issues regarding Dialog component that needs to be defined and think trough before going forward.

src/components/DialogHeader/DialogHeader.tsx
imoutaharik: collapseBottom removes the border and spacing below the header so you could use tabs for example. It's in the Figma file.

src/components/DialogFooter/DialogFooter.tsx
```diff
@@ -0,0 +1,21 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+import React, { ReactElement } from 'react';
+import { Button, ButtonProps, ButtonSize } from '@components/Button';
+
+export type DialogFooterProps = {
+    buttons: [ButtonProps] | [ButtonProps, ButtonProps];
+};
+
+export const DialogFooter = ({ buttons }: DialogFooterProps): ReactElement => {
```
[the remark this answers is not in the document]
imoutaharik: Per Design, it was better to limit to 2 buttons. Although we can open the discussion if we allow more.
I didn't want to limit the styling to primary and secondary. Some use cases may need a danger button in the footer. Like a DELETE button.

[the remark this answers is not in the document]
GoranRibic: I don't think we understood each other. What I meant is that you have primary and secondary as props like this:
`
<DialogFooter primary={<Button variant="primary" />} secondary={<Button variant="danger" />} />
`
Where primary and secondary refers to the position of the button and its purpose.

src/components/DialogHeader/DialogHeader.tsx
GoranRibic: Isn't it better to be `headerSize = 'large' | 'medium' | 'small'`? Otherwise instead of `null` maybe `default`?

src/components/DialogHeader/DialogHeader.tsx
GoranRibic: IMHO I would remove that property and other stuff,  as DialogHeader should only be responsible for styling e.g. padding or w/e, and leave it to the end user to compose what Header should have via children. 

In this case: 

```
<DialogHeader>
   <Collapsable in={expanded} ...>
     ...
   </Collapsable>
</DialogHeader>

```

GoranRibic (changes requested):

src/components/DialogHeader/DialogHeader.tsx
imoutaharik: We provide A DialogHeader component with strict styling. You can't pass children. That's what Designers would like to see in all overlays components (including Modal) But it's optional to use it. You can pass a Custom Header to the Dialog component if you want to. Same with Body and Footer. If it happens Designers use the Fondue Designs, it's ready for you to use.

src/components/DialogHeader/DialogHeader.tsx
imoutaharik: Changed to Default.

src/components/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,33 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+import React, { PropsWithChildren, ReactElement } from 'react';
+import { merge } from '@utilities/merge';
+import { CONTAINER_CLASSES } from '@utilities/borderStyle';
+
+export type DialogProps = PropsWithChildren<{
```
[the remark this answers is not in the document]
imoutaharik: Still defining pattern with Jamie. We will have a component in charge of positioning and it's still in thinking process where to put the properties.

imoutaharik:
Closing as we will work on all overlays by chunks.

PR #1170, opened 2023-01-23
feat: add updateValueOnChange property

triggertoo:


src/components/RichTextEditor/RichTextEditor.cy.tsx
```diff
@@ -959,7 +998,7 @@ describe('RichTextEditor Component', () => {
         );
     };
 
-    describe.only('column break plugin', () => {
```
rawi96, line 962: 🤣

rawi96 (approved):

PR #1174, opened 2023-01-24
fix(MultiSelect): MultiSelect component overflow

selimovicz:


imoutaharik (approved):
Thanks for debugging this!!! 🎉  I'm approving this as there is an annoying issue in Clarify to fix urgently but be aware a refactoring is on its way soon.
We'll have to clean up the component and unify both emphasis.

Some comments:
- We use portals in both dropdown and multiSelect so they can work in a container with overflow. This leads to an accessibility issue but as discussed with @fabriciomendonca, we need to test thoroughly if it's safe to remove portal and if using a fixed strategy of popper.js is enough.  
- A new design and approach will be proposed but we'll make sure to keep you in the loop.

PR #1175, opened 2023-01-24
chore: sync node test command with clarify

SamCreasey:


imoutaharik:
@SamCreasey Can we close this Pr? As I believe we'll keep things in Singular. :)

imoutaharik:
@SamuelAlev Should we close this PR? I thought we agreed it's better to have it singular on both repository.

PR #1177, opened 2023-01-24
chore(Badge): increase color contrast when Badge has positive styling

dominiquemb:


imoutaharik:
@dominiquemb Maybe we can close this PR as we are getting new alias-tokens from design for the color contrast.

PR #1187, opened 2023-01-27
fix(Tooltip): icon alignment on Safari and arrow styling

fabriciomendonca:


src/components/Tooltip/BrightHeader.tsx
```diff
@@ -47,7 +47,7 @@ export const BrightHeader: FC<BrightHeaderProps> = ({ headerStyle }) => {
         <div
             data-test-id="bright-header"
             className={merge([
-                'tw-h-9 tw-relative tw-z-10 tw-flex tw-justify-center tw-items-center tw-rounded-t-md tw-text-white',
+                'tw-h-9 tw-relative tw-z-[120000] tw-z-10 tw-flex tw-justify-center tw-items-center tw-rounded-t-md tw-text-white',
```
imoutaharik: There is 2 tailwind classes for z-index here `tw-z-[120000] tw-z-10`

src/components/Tooltip/Tooltip.stories.tsx
```diff
@@ -90,7 +90,9 @@ export const TooltipComponent: StoryFn<TooltipProps> = (args: TooltipProps) => {
                 {...args}
                 triggerElement={
                     <button className="tw-flex tw-justify-center tw-items-center">
-                        <IconExclamationMarkCircle size={IconSize.Size16} filled />
+                        <span className="tw-flex tw-leading-3">
+                            <IconExclamationMarkCircle size={IconSize.Size16} filled />
```
imoutaharik, line 94: Here you can use the `IconExclamationMarkCircle16Filled` component directly, as we know already the size we want to use. We'd like that to be the norm and encourage people to use the sized Icon component whenever possible.

imoutaharik (approved):

PR #1188, opened 2023-01-28
fix(Checkbox): fixing accessibility errors

dominiquemb:


src/components/TooltipIcon/TooltipIcon.tsx
```diff
@@ -41,6 +41,7 @@ export const TooltipIcon: FC<TooltipIconProps> = ({
                             <button
                                 type="button"
                                 data-test-id="tooltip-icon-trigger"
+                                aria-label="More info"
```
imoutaharik, line 44: I think it's better to expose the `aria-label` and not have it hardcoded as it might not always be "more info". Like that they can also pass a string with translation.

src/components/TooltipIcon/TooltipIcon.tsx
```diff
@@ -28,10 +29,12 @@ const tooltipTriggerStyleClass: Record<TooltipIconTriggerStyle, string> = {
 
 export const TooltipIcon: FC<TooltipIconProps> = ({
     tooltip,
+    'aria-label': ariaLabel = 'More info',
     iconSize = IconSize.Size16,
     triggerIcon = <IconQuestionMarkCircle />,
     triggerStyle = TooltipIconTriggerStyle.Primary,
 }: TooltipIconProps) => {
+    const ariaLabelToUse: string = ariaLabel ?? String(tooltip?.content);
```
imoutaharik, line 37: I think this should be removed then.

imoutaharik (changes requested):
I noticed alignment of the input and text is off.

Screenshot 2023-02-02 at 10 56 29

dominiquemb:
> I noticed alignment of the input and text is off.
> 
> Screenshot 2023-02-02 at 10 56 29

I think that the alignment matches the design. What do you think? 

Screenshot 2023-02-02 at 12 26 04

imoutaharik:
@dominiquemb There is a test failing on MultiSelect.cy.tsx. It's not a flaky one. There is some error for the checkbox.

imoutaharik:
@dominiquemb There is a test failing on MultiSelect.cy.tsx. It's not a flaky one. There is some error for the checkbox.

rawi96 (approved):
Change looks good 👍

fabriciomendonca (approved):

imoutaharik (approved):

PR #1190, opened 2023-01-30
chore(Checkbox): moving note to be directly under label to match design

dominiquemb:


src/components/Checkbox/Checkbox.tsx
```diff
@@ -41,6 +47,7 @@ const isCheckedOrMixed = (checked: CheckboxState): boolean => {
 const CheckboxComponent: ForwardRefRenderFunction<HTMLInputElement, CheckboxProps> = (
     {
         id: propId,
+        emphasis = 'Default',
```
fabriciomendonca, line 50: Why not to use `emphasis = CheckboxEmphasis.Default`?

src/components/Checkbox/Checkbox.tsx
```diff
@@ -130,31 +137,43 @@ const CheckboxComponent: ForwardRefRenderFunction<HTMLInputElement, CheckboxProp
                                   !isCheckedOrMixed(state) &&
                                       'tw-border-black-80 tw-bg-white hover:tw-border-black dark:tw-border-white dark:tw-bg-black dark:hover:tw-border-black-20 dark:hover:tw-bg-black-90 group-hover:tw-bg-white group-hover:tw-border-black dark:group-hover:tw-border-black-20 dark:group-hover:tw-bg-black-90',
                                   isCheckedOrMixed(state) &&
+                                      emphasis === CheckboxEmphasis.Weak &&
+                                      'tw-bg-box-neutral-strong tw-text-white hover:tw-bg-box-neutral-inverse dark:tw-text-black group-hover:tw-text-white group-hover:tw-bg-box-neutral-strong dark:group-hover:tw-text-black',
```
fabriciomendonca, line 141: We are mixing Alias tokens and the Brand tokens here... If I'm not wrong, when we use Alias tokens we should get rid of those styles for `dark:...`. Idk which Alias token would replace the `tw-text-white dark:tw-text-black`, maybe @thisguyscode (Jamie)?

fabriciomendonca (changes requested):

src/components/Checkbox/Checkbox.tsx
thisguyscode: `tw-text-white` -> `tw-bg-box-neutral-strong-inverse`

Don't need any `dark:...` stuff

fabriciomendonca (approved):
lg2m

PR #1196, opened 2023-02-02
refactor(RichTextEditor): Use editor.apply in column break plugin

triggertoo:
- Refactor: Use `editor.apply` instead key event handling.
- Validation: Check number of columns when applying column break to a selection of nodes
- Allow list elements to break


Walkeezy:
Works really well 👍🏻

src/components/RichTextEditor/Plugins/ColumnBreakPlugin/withColumnBreak.ts
```diff
@@ -0,0 +1,33 @@
...
+
+export const withColumnBreak =
+    (columns: number) =>
+    <E extends PlateEditor = PlateEditor>(editor: E) => {
+        const { apply } = editor;
+
+        editor.apply = (operation) => {
+            switch (operation.type) {
+                case 'split_node':
+                    operation.properties = { ...operation.properties, breakAfterColumn: false };
```
triggertoo, line 15: avoid deleting newline here

celine-s (commented):
hotkeys [Shift, Ctrl, Enter] are not working atm.

backspace & delete: if there is a columnbreak inbetween it should only delete the line and keep the break.

celine-s (commented):
@triggertoo I tested the del and backspace and it works perfectly well 👍 :).

Just something nit: While deleting the columnbreak from the paragraph before the cursor jumps to this paragraph and I think it should stay at the paragraph we where.  Do you know if we easily say where the cursor stays after the backspace?

I added the hotkeys and some test about the key events (hotkey, backspace, del). Wdyt? Shall we add some more?

triggertoo:
> @triggertoo I tested the del and backspace and it works perfectly well 👍 :).
> 
> Just something nit: While deleting the columnbreak from the paragraph before the cursor jumps to this paragraph and I think it should stay at the paragraph we where. Do you know if we easily say where the cursor stays after the backspace?
> 
> I added the hotkeys and some test about the key events (hotkey, backspace, del). Wdyt? Shall we add some more?

👍 [fixed](https://github.com/Frontify/fondue/pull/1196/commits/770a8a2656699e549d6fa066eb91012bb51b3c67)

src/components/RichTextEditor/RichTextEditor.cy.tsx
```diff
@@ -1100,6 +1104,80 @@ describe('RichTextEditor Component', () => {
             checkPosition('be.gt', 100, 'second');
             checkPosition('be.gt', 100, 'Level 5');
         });
+
+        it('it should add only one column break after the first break, when there are only two columns', () => {
```
triggertoo, line 1108:
```suggestion
        it('should add only one column break after the first break, when there is a maximum of two columns', () => {
```

src/components/RichTextEditor/RichTextEditor.cy.tsx
```diff
@@ -1100,6 +1104,80 @@ describe('RichTextEditor Component', () => {
...
+            checkPosition('be.lessThan', 100, 'hello');
+
+            // break after should not be enabled
+            selectTextValue('last newline');
+            cy.get(TOOLBAR_GROUP_1).children().eq(-1).realHover().should('not.include.html', activeButtonClassNames);
+            checkPosition('be.lessThan', 500, 'last newline');
+            checkPosition('be.lessThan', 500, 'World');
+        });
+
+        it('the button should be disabled when already at max column breaks', () => {
```
triggertoo, line 1127:
```suggestion
        it('should disable the column break button when already at max column breaks', () => {
```

src/components/RichTextEditor/RichTextEditor.cy.tsx
```diff
@@ -1100,6 +1104,80 @@ describe('RichTextEditor Component', () => {
...
+        it('should delete a column break with del key', () => {
+            cy.mount(<RichTextEditorWithTwoColumns />);
+            cy.get('[contenteditable=true]').click().type('hello{shift+ctrl+enter}{enter}W{leftArrow}{leftArrow}{del}');
+            cy.get('[contenteditable=true]').should('not.include.html', columnBreakClassNames);
+            // checks that the words are still separated 'hello' and 'W', as only the column break should be deleted
+            selectTextValue('hello');
+            cy.get(TOOLBAR_FLOATING).should('be.visible');
+        });
+
+        it('it should only add one column break with hotkeys, when there are two columns', () => {
```
triggertoo, line 1169:
```suggestion
        it('should only add one column break with hotkeys, when there is a maximum of two columns', () => {
```

triggertoo (commented):
nice work @celine-s

src/components/RichTextEditor/Plugins/ColumnBreakPlugin/createColumnBreakPlugin.ts
```diff
@@ -1,135 +1,40 @@
...
-import { setBreakAfter } from './utils/setBreakAfter';
+import { withColumnBreak } from './withColumnBreak';
+import { onKeyDownColumnBreak } from './onKeyDownColumnBreak';
 
 export const KEY_ELEMENT_BREAK_AFTER = 'breakAfterColumn';
 
 export class BreakAfterPlugin extends Plugin {
     private columns: number;
-    private gap: number;
+    private gap: string | number;
```
triggertoo, line 13: why do we need to support string?

celine-s (approved):
love it! 🤩

Walkeezy (approved):

PR #1211, opened 2023-02-07
refactor: register tree items

SamuelAlev:


GoranRibic (approved):

src/components/Tree/Tree.tsx
```diff
@@ -1,93 +1,258 @@
...
         return () => {
-            window.removeEventListener('keydown', downKeyHandler);
-            window.removeEventListener('keyup', upKeyHandler);
+            window.removeEventListener('keydown', keyDownHandler);
+            window.removeEventListener('keyup', keyUpHandler);
         };
-    }, []);
+    }, [keyDownHandler, keyUpHandler]);
+
+    useEffect(() => {
```
GoranRibic, line 185: Isn't this `useEffect` redundant?

src/components/Tree/TreeItem.tsx
```diff
@@ -9,33 +19,66 @@ import { DraggableItem, DropZonePosition } from '@utilities/dnd';
...
+import { getAcceptTypes, getItemPositionInParent, getNextItemToFocus, getPreviousItemToFocus } from './helpers';
+import { cloneThroughFragments, flattenChildren } from './utils';
 
 const DRAGGING_OPACITY = 0.4;
 const DEFAULT_OPACITY = 1;
 
 export const TreeItem = ({
     id,
-    sort,
+    parentId,
```
GoranRibic, line 33: why does TreeItem have a `parentId` prop? Should that be inside treeState?

GoranRibic: What is level used for?

src/components/Tree/TreeItem.tsx
```diff
@@ -44,50 +87,122 @@ export const TreeItem = ({
...
+                        onExpand(id, false);
+                    } else if (treeItemState?.parentId) {
+                        const parentId = treeItemState.parentId;
+                        const parentState = treeState.items.get(parentId);
+                        parentState?.domElement?.focus();
+                    }
+                    break;
+
+                case 'ArrowUp':
+                    event.preventDefault();
```
GoranRibic, line 167: event.preventDefault() can be put on top

src/components/Tree/helpers/index.ts
```diff
@@ -0,0 +1,6 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+export * from './getAcceptedTypes';
+export * from './getItemPositionInParent';
+export * from './getNextItemToFocus';
+export * from './getPreviousItemToFocus';
```
GoranRibic, line 6: IMO these should be different files and some of them combined. 
getAcceptedTypes -> dnd.ts
getItemPositionInParent -> position.ts
getNextItemToFocus, getPreviousItemToFocus -> focus.ts

or something similar in nature.

src/components/Tree/types.ts
```diff
@@ -40,6 +44,32 @@ export type TreeItemProps = {
...
+    childrenIds?: string[];
+    level: number;
+    domElement?: HTMLElement;
+};
+
+export type TreeState = {
+    items: Map<string, TreeItemState>;
+    selectedIds: Set<string>;
+    expandedIds: Set<string>;
+    selectionMode: 'single' | 'multiselect';
```
GoranRibic, line 64: multi-select should be allowed or disallowed through property multi-select in the tree component.

src/components/Tree/TreeItem.tsx
SamuelAlev: > why does TreeItem have a parentId prop? Should that be inside treeState?
Passed by its parent when rendering happens

> What is level used for?
[aria-level](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-level)

src/components/Tree/TreeItem.tsx
SamuelAlev: It prevents tabbing

src/components/Tree/helpers/index.ts
SamuelAlev: Easier to find the methods IMO
What would be the benefit?

src/components/Tree/Tree.tsx
SamuelAlev: Different use-case

PR #1215, opened 2023-02-08
docs(Tooltip): add tooltip composition in context stories

fabriciomendonca:
Added new stories for Composition in context.

**Findings**
Due to the usage of `createPortal` the `Dropdown` overlay will always be on top of the `Tooltip` if both are open at the same time


imoutaharik (approved):
Thanks! and good findings -> will be useful when we reevaluate the use of Portal in the MultiSelect and Dropdown.

PR #1217, opened 2023-02-08
fix(Checkbox)!: adding emphasis styles, moving notes to below label

dominiquemb:


SamuelAlev:
Renaming/deleting props is a breaking change and breaking changes needs a `!`

imoutaharik (commented):
Screenshot 2023-02-08 at 16 35 31

I believe that according to Figma, the background colour for the checkbox should be base and change to box-neutral on hover.

dominiquemb:
> Screenshot 2023-02-08 at 16 35 31
> 
> I believe that according to Figma, the background colour for the checkbox should be base and change to box-neutral on hover.

Good catch. Fixed!

imoutaharik (approved):
Once this gets released, It would be nice to quickly fix the types in Clarify. 
`/client/Common/Form/Checkbox`

PR #1219, opened 2023-02-08
fix(Tree): Drag and drop fix

leko91:


SamuelAlev:
Fixes DnD issues caused by registration of items, now async items are correctly registered based on their index in JSX
```tsx
<Tree>
	<TreeItem> {/* sort: 0 */}
		<TreeItem /> {/* sort: 0 */}
		<TreeItem /> {/* sort: 1 */}
		<TreeItem /> {/* sort: 2 */}
	</TreeItem>
	<TreeItem /> {/* sort: 1 */}
	<TreeItem /> {/* sort: 2 */}
</Tree>
```

src/components/Tree/Tree.tsx
```diff
@@ -64,40 +64,107 @@ const reducer = (state: TreeState, action: TreeStateAction): TreeState => {
...
-            }
-
             return {
                 ...state,
                 items: state.items.set(id, {
                     parentId,
-                    childrenIds,
                     level,
                     domElement: ref,
+                    childrenIds: undefined,
```
GoranRibic, line 151: why not pass an empty array? So you don't have to do `?` checks

src/components/Tree/TreeItem.tsx
```diff
@@ -84,16 +80,17 @@ export const TreeItem = ({
...
     });
 
     const handleDrop = useCallback(
         (
-            targetItem: DraggableItem<{ id: string; sort: Nullable<number> }>,
-            sourceItem: DraggableItem<{ id: string; sort: Nullable<number> }>,
-            position: DropZonePosition,
+            targetItem: DraggableItem<{ id: string; sort: number }>,
+            sourceItem: DraggableItem<{ id: string; sort: number }>,
+            position: DropZonePosition, // [abc, def, ghi]
```
GoranRibic, line 90: // [abc, def, ghi] ?

src/components/Tree/TreeItem.tsx
```diff
@@ -126,162 +123,165 @@ export const TreeItem = ({
...
-                    aria-level={treeItemState?.level}
-                    aria-selected={treeState.selectedIds.has(id)}
-                    aria-expanded={treeState.expandedIds.has(id)}
-                    tabIndex={0}
-                    onClick={handleSelect}
-                    onKeyDown={handleKeyDown}
-                    onMouseEnter={handleMouseEnter}
-                    onMouseLeave={handleMouseLeave}
+        <>
+            {(propsDraggable ?? draggable) && sort === 0 ? (
```
GoranRibic, line 194: what is propsDraggable?

GoranRibic (approved):

src/components/Tree/TreeItem.tsx
SamuelAlev: `Add draggable on a per item basis.`
Allows remove dragging on one specific item (cover page and brandlink shouldn't be draggable for example)

src/components/Tree/Tree.tsx
SamuelAlev: I prefer `undefined` for the leafs of a tree

src/components/Tree/TreeItem.tsx
GoranRibic: I would give it a more meaningful name then, as it is confusing in my opinion. But that is not urgent ATM anyway. 👍

PR #1232, opened 2023-02-10
chore: use `pnpm`

SamuelAlev:
Migrate from `npm` to `pnpm` (+ adjust CI)
Make tests run in parallel
Add all `peerDependencies`
Push the latest icons (seems forgotten)
Ensure workflow run once per PR (abort previous commits workflows)

**To test**
[Install pnpm](https://pnpm.io/installation) then proceed with:
```shell
rm -rf node_modules
pnpm i
```


package.json
```diff
@@ -168,7 +184,8 @@
         "tinycolor2": "1.4.2",
         "unified": "10.1.2",
         "unist-util-visit": "4.1.2",
-        "xstate": "4.28.1"
+        "xstate": "4.28.1",
+        "zustand": "4.3.3"
```
imoutaharik, line 188: Where do we need Zustand in Fondue? I'm wondering.

package.json
```diff
@@ -160,6 +175,7 @@
         "react-textarea-autosize": "8.4.0",
         "remark-gfm": "3.0.1",
         "remark-parse": "10.0.1",
+        "scheduler": "0.23.0",
```
imoutaharik, line 178: And that scheduler? Do we need to execute a cron job?

.github/workflows/pull-request.yml
```diff
@@ -87,22 +111,28 @@ jobs:
...
-                  restore-keys: node_modules-
+                      ${{ steps.pnpm-cache.outputs.pnpm_cache_dir }}
+                      ~/.cache/Cypress
+                  key: pnpm-with-cypress-${{ runner.os }}-${{ hashFiles('pnpm-lock.yaml') }}
+                  restore-keys: pnpm-with-cypress-${{ runner.os }}-${{ hashFiles('pnpm-lock.yaml') }}
 
             - name: Install dependencies
               if: steps.npm_cache.outputs.cache-hit != 'true'
-              run: npm ci
+              run: pnpm install --frozen-lockfile
```
imoutaharik, line 135: So `pnpm install --frozen-lockfile` is the new `npm ci`, should we run it everytime we update/switch our branches? Or pnpm install is enough? Otherwise it would be nice to have a script with a shorter name for it. 😅 no?

package.json
SamuelAlev: These are `peerDependencies`, `npm@>7` install them automatically where `pnpm` don't.

package.json
SamuelAlev: Same as previous comment

.github/workflows/pull-request.yml
SamuelAlev: > should we run it everytime we update/switch our branches? Or pnpm install is enough?

`pnpm i` is enough

imoutaharik (approved):

PR #1234, opened 2023-02-11
chore: updated vite with lib name

bojangles-m:
Vite needs lib name to build the library


bojangles-m:
New version of vite 4, no need for this.

PR #1239, opened 2023-02-17
feat(new-component): Context-Menu

rawi96:


imoutaharik:
Hello @rawi96, Thanks for taking initiative but I think that the addition of new components like that should be discussed first with the Team. We agreed a Toast component could live in Fondue but I believe it shouldn't be called `Context Menu` and the example here and implementation is strongly `Guidelines block` specific, too. And we try to avoid that 🤷🏼‍♀️ .

PR #1242, opened 2023-02-20
docs(Menu): fix code snippets

fabriciomendonca:
The docs addon was supposed to show the full source, but there is a known issue and a suggested workaround.
However, that workaround didn't work in Fondue.

https://github.com/storybookjs/storybook/tree/master/addons/storysource#displaying-full-source

Another workaround I found was to add the `BasicMenuComponent` as the `docs.source.code` parameter value, but it is injecting the transpiled code instead of the `tsx` code.

The best solution I could find right now, is to create a string with the component source code so it can be presented exactly how it was coded in the stories. Also added a `Usage Example`


imoutaharik (commented):
Thanks Fabricio. I took a look at it and I still think there is a lot going on in the example code. It's not cristal clear how the component actually is.
I also noticed both example `with buttons` and `with links` have the same code snippet and yet renders different content.

As a suggestion, I would do it more this way :
1- Reorder the stories
first the Menu itself without trigger and show the snippet as simple as possible like this
```
<Menu>
  <MenuItem>label 1<MenuItem/>
  <MenuItem>label 2<MenuItem/>
  <MenuItem>label 3<MenuItem/>
<Menu/>
```
Something like that, to reflect the structure of the component itself.

And then, have examples with a button trigger and how it could be used.

The goal being, you should understand at first glance the component and then if you want to know exactly how you can use it with a trigger and close/open function, you can go through the second example if you'd like to.

fabriciomendonca:
> Thanks Fabricio. I took a look at it and I still think there is a lot going on in the example code. It's not cristal clear how the component actually is. I also noticed both example `with buttons` and `with links` have the same code snippet and yet renders different content.
> 
> As a suggestion, I would do it more this way : 1- Reorder the stories first the Menu itself without trigger and show the snippet as simple as possible like this
> 
> ```
> <Menu>
>   <MenuItem>label 1<MenuItem/>
>   <MenuItem>label 2<MenuItem/>
>   <MenuItem>label 3<MenuItem/>
> <Menu/>
> ```
> 
> Something like that, to reflect the structure of the component itself.
> 
> And then, have examples with a button trigger and how it could be used.
> 
> The goal being, you should understand at first glance the component and then if you want to know exactly how you can use it with a trigger and close/open function, you can go through the second example if you'd like to.

Forgot to change the With Links Usage Example with the right items list... And since they are strings now, the rendered result is different than the code. I will fix this. Thanks for the catch.

I will convert the `Menu Without Trigger` story into that simple one and render it first, will also rename it to `Component Structure` or sth

src/components/Menu/Menu.stories.tsx
```diff
@@ -2,34 +2,71 @@
...
     component: Menu,
     tags: ['autodocs'],
 } as Meta<MenuProps>;
 
-const TemplateWithoutTrigger: StoryFn<MenuProps & { items: Item[] }> = (args) => (
+const TemplateWithoutTrigger: StoryFn = () => (
     <div className="tw-p-4">
-        <NoTriggerMenuComponent {...args} />
+        <Menu>
+            <MenuItem link="https://www.frontify.com">Item 1</MenuItem>
```
imoutaharik, line 26: I would remove the link from the actual story implementation. I was testing the component and clicked on the ltems. It can be annoying to get redirected to Frontify. Especially because it does not open on another tab.

imoutaharik (approved):

PR #1247, opened 2023-02-21
fix(Rich Text Editor): removed rte init plugin

bojangles-m:
Removed init plugin because it interferes with the List plugin and it does not work correctly. Plugins also work with the soft break, for further use it can also be implemented separately.


SamCreasey (approved):
Looks good to me 👍

triggertoo (approved):
👍

PR #1248, opened 2023-02-21
fix(Checkbox)!: accessibility, design and cypress test updates

dominiquemb:


imoutaharik:
@dominiquemb Is this Pr still needed or should we close it? I think all updates for the checkbox have already been merged right?

PR #1249, opened 2023-02-21
feat: add `displayName`

SamuelAlev:
Add `displayName` for easier debugging
Screenshot 2023-02-22 at 19 13 44


triggertoo (approved):

imoutaharik:
@SamuelAlev The downside of that as spotted by Fabricio is that the displayName appears in the storybook docs. 
I'm not sure if we can change again the displayName just in the story file after the import. 🤔 (just thinking out loud) haven't tried.

src/components/Button/Button.tsx
```diff
@@ -135,3 +135,4 @@ const ButtonComponent: ForwardRefRenderFunction<HTMLButtonElement | null, Button
 };
 
 export const Button = forwardRef(ButtonComponent);
+Button.displayName = 'Button';
```
imoutaharik, line 138: Why `AddBlockButton` and `Button` don't get the prefix?

SamuelAlev: Added manually, forgotten to add the prefix.

imoutaharik (approved):

PR #1250, opened 2023-02-21
feat(Rich Text Editor): list plugin with soft break

bojangles-m:
Added option for soft break to behave exactly as break.


src/components/RichTextEditor/Plugins/ListPlugin/ListPlugin.ts
```diff
@@ -15,15 +17,28 @@ export const createLicStylePlugin = createPluginFactory<ToggleMarkPlugin>({
...
+        this.isSoftBreak = props?.isSoftBreak ?? false;
     }
 
     plugins() {
-        return [createListPlugin(), createLicStylePlugin()];
+        return [
+            createListPlugin({
+                withOverrides: withList,
+                options: {
+                    isSoftBreak: this.isSoftBreak,
```
triggertoo, line 38: maybe rename to `useSoftBreak`?

triggertoo (commented):
looks good 👍 Could you add a test for the behaviour?

src/components/RichTextEditor/Plugins/ListPlugin/ListPlugin.ts
bojangles-m: Would you have some other name for it, this is more like Hook name :)

bojangles-m:
> looks good 👍 Could you add a test for the behaviour?

Do you have an idea how this test would look like?

src/components/RichTextEditor/Plugins/ListPlugin/ListPlugin.ts
triggertoo: imho it's unrelated to react, but we can also keep it that way.

triggertoo:
> > looks good 👍 Could you add a test for the behaviour?
> 
> Do you have an idea how this test would look like?

you could check if a softbreak (shift-enter) in a list creates a new list item

src/components/RichTextEditor/__tests__/ListPlugin.cy.tsx
```diff
@@ -0,0 +1,88 @@
...
+        cy.get(TEXTSTYLE_DROPDOWN_TRIGGER).click({ force: true });
+        cy.get(TEXTSTYLE_OPTION).first().click();
+        cy.get(firstListItemSelector).should(
+            'have.attr',
+            'style',
+            'font-size: 48px; font-style: normal; font-weight: 700;',
+        );
+    });
+
+    it.skip('renders the list item without text decoration underline', () => {
```
triggertoo, line 56: please remove skip to execute the test

src/components/RichTextEditor/__tests__/ListPlugin.cy.tsx
```diff
@@ -0,0 +1,88 @@
...
+        cy.mount(<RichTextEditorWithUnorderedListStyles />);
+
+        cy.get('[contenteditable=true] li').should('have.class', '!tw-no-underline');
+    });
+
+    it.skip('renders custom styled list items', () => {
```
triggertoo, line 62: please remove skip to execute the test

src/components/RichTextEditor/__tests__/ListPlugin.cy.tsx
```diff
@@ -0,0 +1,88 @@
...
+    it('renders new list item with soft break', () => {
+        cy.mount(<RichTextEditorWithUnorderedListStyles isSoftBreak={true} />);
+        const itemValue = 'New item in list';
+
+        const firstListItemSelector = '[contenteditable=true] ul:first-child > li:first-child';
+        const secondListItemSelector = '[contenteditable=true] ul:first-child > li:nth-child(2)';
+
+        cy.get(firstListItemSelector).type(`{end}{shift+enter}${itemValue}`);
+
+        cy.get(firstListItemSelector).children().should('have.length', 1);
```
triggertoo, line 85: you could check if the number of list items is equal 2 to make sure, a list item was added

src/components/RichTextEditor/__tests__/ListPlugin.cy.tsx
bojangles-m: of course, thanx, forget when i suppressed the test

src/components/RichTextEditor/__tests__/ListPlugin.cy.tsx
bojangles-m: udpated

src/components/RichTextEditor/__tests__/ListPlugin.cy.tsx
bojangles-m: udpated

triggertoo: it seems a bit confusing if the length is 4 here. what if you instead just initialise the editor with just one children li element (not loading `unorderedListValue` from the helper but defining it yourself. it's easier to understand if the test fails.

triggertoo (approved):

PR #1256, opened 2023-02-22
feat(toast): add new toast component

rawi96:
New generic toast component.
[Slack Conversation](https://frontify.slack.com/archives/C02PSJTPA3D/p1676879350059769)


src/components/Toast/Toast.tsx
```diff
@@ -0,0 +1,34 @@
...
+import { AnimatePresence, motion } from 'framer-motion';
+import React, { ReactElement } from 'react';
+import { ToastProps, toastStylesBackgroundColorsMap } from './types';
+
+export const Toast = ({ isOpen, style, icon, children }: ToastProps): ReactElement => {
+    return (
+        <AnimatePresence>
+            {isOpen && (
+                <motion.div
+                    className="tw-fixed tw-bottom-0 tw-left-[60px] tw-min-h-[4.25rem] tw-m-5 tw-rounded-lg tw-border tw-shadow-lg tw-bg-white tw-border-black-10"
```
SamCreasey, line 13: Positioning of the toast component should be done externally I think. Toast should be able to be placed anywhere. In this case we might need to add an opacity change too 🤔

src/components/Toast/Toast.tsx
```diff
@@ -0,0 +1,34 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+import { merge } from '@utilities/merge';
+import { AnimatePresence, motion } from 'framer-motion';
+import React, { ReactElement } from 'react';
+import { ToastProps, toastStylesBackgroundColorsMap } from './types';
+
+export const Toast = ({ isOpen, style, icon, children }: ToastProps): ReactElement => {
+    return (
```
SamCreasey, line 9:
```suggestion
export const Toast = ({ isOpen, style, icon, children }: ToastProps): ReactElement => (
```

src/components/Toast/types.tsx
```diff
@@ -0,0 +1,24 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+import { ReactChild } from 'react';
+
+export type ToastProps = {
+    isOpen: boolean;
+    style: ToastStyles;
+    icon: ReactChild;
+    children?: ReactChild;
```
SamCreasey, line 9:
```suggestion
    icon: ReactChild;
    children?: ReactNode;
```
I think ReactChild will throw a typeerror if you add more than one child, but maybe that's intended?

src/components/Toast/types.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+export enum ToastStyles {
+    Information = 'Information',
+    Warning = 'Warning',
+    Tip = 'Tip',
+    Note = 'Note',
+}
+
+export const toastStylesBackgroundColorsMap: Record<ToastStyles, string> = {
+    [ToastStyles.Information]: 'tw-bg-violet-60',
```
SamCreasey, line 20: use alias tokens, not brand tokens here :)

src/components/Toast/types.tsx
```diff
@@ -0,0 +1,24 @@
...
+    style: ToastStyles;
+    icon: ReactChild;
+    children?: ReactChild;
+};
+
+export enum ToastStyles {
+    Information = 'Information',
+    Warning = 'Warning',
+    Tip = 'Tip',
+    Note = 'Note',
```
SamCreasey, line 16:
```suggestion
    Loud = 'Loud',
    Danger = 'Danger',
    Success = 'Success',
    Warning = 'Warning',
```

SamCreasey (changes requested):

src/components/Toast/types.tsx
SamuelAlev, line 9:
```suggestion
    icon: ReactNode;
    children?: ReactNode;
```

`ReactChild` is deprecated in React 18

src/components/Toast/types.tsx
```diff
@@ -0,0 +1,24 @@
...
+import { ReactChild } from 'react';
+
+export type ToastProps = {
+    isOpen: boolean;
+    style: ToastStyles;
+    icon: ReactChild;
+    children?: ReactChild;
+};
+
+export enum ToastStyles {
```
SamuelAlev, line 12:
```suggestion
export enum ToastStyle {
```

src/components/Toast/types.tsx
imoutaharik: I agree it should be ReactNode

src/components/Toast/types.tsx
imoutaharik: The idea was to have the same color option as the Tooltip: Purple, green, red and yellow. But I agree with that naming. Design in Figma as a Fondue component is not done yet and I believe there might be a possibility the colors won't be that bright.

src/components/Toast/Toast.tsx
```diff
@@ -0,0 +1,32 @@
...
+import { merge } from '@utilities/merge';
+import { AnimatePresence, motion } from 'framer-motion';
+import React, { ReactElement } from 'react';
+import { ToastProps, toastStylesBackgroundColorsMap } from './types';
+
+export const Toast = ({ isOpen, style, icon, children }: ToastProps): ReactElement => (
+    <AnimatePresence>
+        {isOpen && (
+            <motion.div
+                className="tw-min-h-[4.25rem] tw-rounded-lg tw-border tw-shadow-lg tw-bg-white tw-border-black-10"
```
imoutaharik, line 12: As all our overlays, you can use the `CONTAINER_CLASSES` like in the `Menu` Component.

imoutaharik:
@rawi96 I wouldn't call the Story LeftLoud and RightDanger. 
I would have:
1- informative Toast at the bottom
2- Warning Toast at the Top (as an example)
3- With interactive content

-> Adding some button in the last one. And also don't forget all Fondue components should be AA level for accessibility so I would also add a test case in cypress for being able to focus on the interactive elements.

src/components/Toast/Toast.stories.tsx
```diff
@@ -0,0 +1,72 @@
...
+const WithRightBottomContainer: StoryFn<ToastProps> = (args) => (
+    <div className="tw-fixed tw-bottom-0 tw-right-0 tw-m-3">
+        <ToastComponent {...args} />
+    </div>
+);
+
+export const InformativeToastAtTheBottom = WithLeftBottomContainer.bind({});
+InformativeToastAtTheBottom.args = {
+    isOpen: true,
+    style: ToastStyle.Loud,
```
imoutaharik, line 37: I would add argTypes to the story with the possibility to select between all style options, and animationDirection as well.

Screenshot 2023-03-02 at 15 32 14

Because now it's just a string. Would be better to display all options available.

src/components/Toast/Toast.tsx
```diff
@@ -0,0 +1,49 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+import { INSET_BORDER } from '@utilities/borderStyle';
+import { merge } from '@utilities/merge';
+import { AnimatePresence, motion } from 'framer-motion';
+import React, { ReactElement } from 'react';
+import { ToastAnimationDirection, ToastProps, toastStylesBackgroundColorsMap } from './types';
+
+const CONTAINER_BASE_CLASSES = 'tw-min-h-[4.25rem] tw-bg-base tw-rounded-lg tw-shadow-mid tw-z-[120000]';
+const CONTAINER_CLASSES = merge([CONTAINER_BASE_CLASSES, INSET_BORDER]);
```
imoutaharik, line 10: I know it was repeated in the Menu and Flyout component, but to avoid duplicated code, I would move those classes to utilities.

imoutaharik (commented):
Thanks for all the changes!! 🎉  It looks good to me, I just added a couple of comments.
I'm adding @LidijaStoilova also to review the a11y part.

Thanks for taking care!

src/components/Toast/Toast.tsx
rawi96: `INSET_BORDER` from `@utilities/borderStyle` is used in:
- `Overlay.tsx`
- `Menu.tsx`
- `Toast.tsx`

In `Overlay.tsx` it is merged with:
- `tw-flex tw-flex-col tw-flex-auto tw-min-h-0 tw-rounded
`


In `Menu.tsx` it is merged with:
- `tw-relative tw-bg-base tw-rounded tw-py-2 tw-shadow-mid tw-z-[120000]
`


And i merged it in the `Toast.tsx` with:
- `tw-min-h-[4.25rem] tw-bg-base tw-rounded-lg tw-shadow-mid tw-z-[120000]
`

So it looks like there is no class that is used in all 3 components (except  INSET_BORDER which is already in utilities).
So i guess it does not make any sense to move something to utilities right?

imoutaharik: Mmm it's not really consistent. Ok, don't worry about that, we'll take care.
This will be standardized when we work on the new `Container` component. 😉

SamCreasey (approved):
Nice work 👍

PR #1257, opened 2023-02-23
test: remove loops and timeout

SamuelAlev:
and replace `esno` with `tsx` (Since v0.15, `esno` is essentially an alias of `tsx`)


fabriciomendonca (approved):
We created tasks for recreating the removed tests without using loops and actually testing the component changes.

PR #1258, opened 2023-02-23
feat(RichTextEditor): Add tooltips

triggertoo:


celine-s (approved):
The design for the ones with hotkeys is slightly different (first word is bold) and not sure about the padding. 

But besides that it works perfect and is looks great 🚢 😍

triggertoo:
> The design for the ones with hotkeys is slightly different (first word is bold) and not sure about the padding.

this one is only about information & hotkeys

PR #1259, opened 2023-02-23
feat(date picker): add range variant for date picker

joseazevedo88:
image

Some new props were exposed to accomplish this, please refer to the `DatePicker` type changes

### DIFF
--- a/src/components/DatePicker/DatePicker.tsx
+++ b/src/components/DatePicker/DatePicker.tsx
@@ -38,6 +38,9 @@ export type DatePickerProps = {
     children?: React.ReactNode;
     hasPopperArrow?: boolean;
     filterDate?: (date: Date) => boolean;
+    onOpen?: () => void;
+    onClose?: () => void;
+    onBlur?: () => void;
 } & (SingleDatePickerProps | RangeDatePickerProps);
 
 const getDayClasses = (variant: DatePickerProps['variant'], date: Date) => {
@@ -52,6 +55,9 @@ export const DatePicker: FC<DatePickerProps> = ({
     isClearable,
     shouldCloseOnSelect,
     onChange,
+    onOpen,
+    onClose,
+    onBlur,
     dateFormat = 'dd MMM yyyy',
     value,
     startDate,
@@ -67,6 +73,16 @@ export const DatePicker: FC<DatePickerProps> = ({
 }) => {
     const [isCalendarOpen, setIsCalendarOpen] = useState<boolean>(false);
 
+    const handleOpen = () => {
+        setIsCalendarOpen(true);
+        onOpen?.();
+    };
+
+    const handleClose = () => {
+        setIsCalendarOpen(false);
+        onClose?.();
+    };
+
     return (
         <div data-test-id="date-picker">
             <DatepickerComponent
@@ -78,6 +94,7 @@ export const DatePicker: FC<DatePickerProps> = ({
                 maxDate={maxDate}
                 calendarStartDay={1}
                 onChange={onChange}
+                onBlur={onBlur}
                 selectsRange={variant === 'range' ? true : false}
                 showPopperArrow={hasPopperArrow}
                 filterDate={filterDate}
@@ -95,8 +112,8 @@ export const DatePicker: FC<DatePickerProps> = ({
                 formatWeekDay={(day) => day.slice(0, 1)}
                 isClearable={isClearable}
                 dateFormat={dateFormat}
-                onCalendarClose={() => setIsCalendarOpen(false)}
-                onCalendarOpen={() => setIsCalendarOpen(true)}
+                onCalendarClose={handleClose}
+                onCalendarOpen={handleOpen}
                 shouldCloseOnSelect={shouldCloseOnSelect}
                 dayClassName={(date) => getDayClasses(variant, date)}
                 renderCustomHeader={({ date, decreaseMonth, increaseMonth, increaseYear, decreaseYear }) => (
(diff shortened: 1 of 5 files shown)


src/components/DatePicker/DatePicker.tsx
```diff
@@ -10,14 +10,42 @@ import { DatePickerTrigger } from './DatePickerTrigger';
 import { IconCaretLeft, IconCaretLeftDouble, IconCaretRight, IconCaretRightDouble } from '@foundation/Icon';
 import { Validation } from '@utilities/validation';
 
+type SingleDatePickerProps = {
+    variant?: 'single';
+    onChange: (date: Date | null) => void;
+    selectsRange?: false;
```
imoutaharik, line 16: is selectsRange property necessary? Is is not just enough to set it's value if variant is "Single" or "Range"

src/components/DatePicker/DatePicker.tsx
```diff
@@ -10,14 +10,42 @@ import { DatePickerTrigger } from './DatePickerTrigger';
...
+};
+
 export type DatePickerProps = {
     placeHolder?: string;
     isClearable?: boolean;
     shouldCloseOnSelect?: boolean;
-    onChange?: (date: Date | null) => void;
     dateFormat?: string;
-    value?: Date;
+    value?: Date | null;
```
imoutaharik, line 34: If I understood correctly. `value` is used in the single variant and for range, you get the values from startDate and endDate. To avoid confusion, I would recommend adding a comment about it in the type definition using JsDocs -> `/** @description Some info there */`

src/components/DatePicker/DatePicker.tsx
```diff
@@ -10,14 +10,42 @@ import { DatePickerTrigger } from './DatePickerTrigger';
...
+    minDate?: Date;
```
imoutaharik, line 35: What the decision behind using minDate and maxDate properties instead of having the restriction property as `Past | Future | Past&Future | None` ?
What's the default? 
As a side note: I also notice a bug but that was already there apparently on beta, past date are greyed out but you can still select them.

src/components/DatePicker/DatePicker.tsx
```diff
@@ -10,14 +10,42 @@ import { DatePickerTrigger } from './DatePickerTrigger';
...
+    maxDate?: Date;
     validation?: Validation;
+    customTrigger?: React.ReactNode;
+    children?: React.ReactNode;
+    hasPopperArrow?: boolean;
+    filterDate?: (date: Date) => boolean;
```
imoutaharik, line 41: What's filterDate for?

src/components/DatePicker/DatePicker.tsx
Kenny806: If it's not specified here, we wouldn't be able to assign it by destructuring props and we would need to check if it exists before assigning. Something like this: `const selectsRange = restProps.selectRange ?? false;` Specifying it in the type definition seems better. Also helps to clarify the various combinations of props that are accepted.

Kenny806: A possible improvement would be `selectsRange?: undefined;`, stating that this prop is expected to be undefined if `variant` is `'single'`, while not breaking the destructuring assignment.

src/components/DatePicker/DatePicker.tsx
Kenny806: This option makes it more flexible. For example, in Asset Audit Trail we want to limit the selectable dates so that user cannot select anything before the date when the tracking started. This cannot be accomplished without setting the dates explicitly. If they're not specified, then any date is selectable

src/components/DatePicker/DatePicker.tsx
Kenny806: Would it be clearer to remove `value` from `DatePickerProps` and add `value?: Date` to `SingleDatePickerProps` and `value?: null` to `RangeDatePickerProps`? This would clarify the relationship + make it more type safe

src/components/DatePicker/DatePicker.tsx
Kenny806: Allows further narrowing of the selectable dates: https://reactdatepicker.com/#example-filter-dates

src/components/DatePicker/DatePicker.tsx
imoutaharik: Even better.

src/components/DatePicker/DatePicker.tsx
imoutaharik: What I mean is that I don't think this property is necessary for in the DatePicker type as it's not really something you should change or control. If your variant is "Single", then selectRange must be false and vice versa. In the DatePickerComponent which the third party one, I would set the selectRange value depending on the variant. I'm not sure why we are exposing it. 
What if f I set variant to "Single" and selectsRange to true? 🤔

joseazevedo88: You can't because of the type restrictions :) But you're right! IT can be controlled via the variant prop:
`selectsRange={variant === 'range' ? true : false}`

src/components/DatePicker/DatePicker.tsx
joseazevedo88: https://reactdatepicker.com/#example-date-range-for-one-datepicker  -  the range variant uses: `startDate`, `endDate` and `selected`(which) we call value. It uses the `selected` to add some css classes, so this typing makes sense for both variants

imoutaharik: In this case, I think that some JSDocs on the property wouldn't hurt. I would easily get confused switching variant.

joseazevedo88: Done 👍

imoutaharik (approved):
Thanks for making the changes. 🎉 
It would be nice also to style the trigger on the range story and format the date.

PR #1266, opened 2023-02-27
feat(Slider): add new component

fabriciomendonca:
The renaming changes can be ignored on this PR since they're gonna be handled by another PR https://github.com/Frontify/fondue/pull/1285

### DIFF
--- a/src/components/Slider/Slider.tsx
+++ b/src/components/Slider/Slider.tsx
@@ -68,11 +68,8 @@ export const Slider = ({
     const id = useMemoizedId(propId);
 
     const onInputChange = (inputValue: string) => {
-        const valueWithoutSuffix = inputValue.replace(valueSuffix, '');
-
-        // make sure the suffix will be removed and added to the field value
-        // only when leaving it
-        setValueWithSuffix(valueWithoutSuffix);
+        // make sure the suffix will be added to the input field
+        setValueWithSuffix(inputValue.replace(valueSuffix, '') + valueSuffix);
     };
 
     const addValueUnitSuffix = () => {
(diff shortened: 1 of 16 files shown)


src/components/Slider/Slider.tsx
```diff
@@ -1,159 +1,376 @@
...
+                        onFocus={onFocus}
+                        onBlur={onBlur}
+                        onMouseDown={startDrag}
+                        onMouseUp={stopDrag}
+                        onTouchStart={startDrag}
+                        onTouchEnd={stopDrag}
+                        onKeyDown={onKeyDown}
+                        disabled={disabled}
+                    >
+                        <span className="tw-absolute tw-block tw-top-1/2 tw--translate-y-1/2 tw-w-full tw-h-1 tw-rounded-sm tw-bg-box-neutral tw-flex-1"></span>
```
imoutaharik, line 327: @fabriciomendonca I just have a question you can surely clarify. What was the reason to better use span and button instead of input type=range as starting point?

fabriciomendonca: I used as a base the HTML structure of the Carbon `Slider` component and the recommendations for accessible slider with `aria-` properties, instead of using the input type=range and keeping it hidden.
I decided for the `button` because it is focusable and tabbable by default, instead of using a div and adding the `tabIndex=-1`. Actually, I've tried it and it was not focusing the element properly. :)
For reference, the Carbon Slider HTML structure
```html
<div class="cds--form-item">
  <label for="slider" class="cds--label" id="slider-label">Slider label</label>
  <div class="cds--slider-container">
    <span class="cds--slider__range-label">0</span>
    <div class="cds--slider" role="presentation" tabindex="-1" stepmuliplier="1" value="50">
      <div class="cds--slider__thumb" role="slider" id="slider" tabindex="0" aria-valuemax="100" aria-valuemin="0" aria-valuenow="34" aria-labelledby="slider-label" style="left: 34%;"></div>
      <div class="cds--slider__track"></div>
      <div class="cds--slider__filled-track" style="transform: translate(0%, -50%) scaleX(0.34);"></div>
    </div>
    <span class="cds--slider__range-label">100</span>
    <input type="number" id="slider-input-for-slider" class="cds--text-input cds--slider-text-input" aria-label="Label for slider value" min="0" max="100" step="1" value="34">
  </div>
</div>
```

fabriciomendonca:
Will close this PR and create a fresh branch from `beta` to avoid dealing with too much conflicts.

PR #1270, opened 2023-03-01
feat: add variable tree item padding

SamCreasey:
Added a new property to the `Tree` component `baseItemPadding` in order to achieve the hover & active styles outlined in the Nav Manager. These styles will now be full-width, but the dropzone will still maintain a visually nested appearance when dragging an item.
The new prop has been added to the `Tree` component as adding a `padding` attribute to the `TreeItem` itself might be confusing for the developer, since the level that is supplied to the `TreeItem` also impacts the padding, this way they are separated and the padding will be consistent for every item under the `Tree` context

### DIFF
--- a/src/components/Tree/Tree.cy.tsx
+++ b/src/components/Tree/Tree.cy.tsx
@@ -7,7 +7,7 @@ import { treeNodesMock } from '@components/Tree/utils/mocks';
 
 const TreeComponent = ({ onSelect }: { onSelect?: (id: Nullable<string>) => void }) => {
     return (
-        <Tree id="treeId" onSelect={onSelect}>
+        <Tree id="treeId" onSelect={onSelect} basePadding={{ x: 50, y: 20 }}>
             {treeNodesMock.map((node) => (
                 <TreeItem key={node.id} id={node.id} contentComponent={() => <span>{node.label}</span>}>
                     {node.nodes?.map((node) => (
@@ -33,6 +33,7 @@ const TreeComponent = ({ onSelect }: { onSelect?: (id: Nullable<string>) => void
 
 const TREE_ID = '[data-test-id=tree]';
 const TREE_ITEM_ID = '[data-test-id=tree-item]';
+const TREE_ITEM_CONTENT_ID = '[data-test-id=tree-item-content]';
 const TREE_ITEM_TOGGLE_ID = '[data-test-id=tree-item-toggle';
 const SUB_TREE_ITEMS_ID = '[data-test-id=sub-tree-items]';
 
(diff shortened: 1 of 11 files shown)


src/components/Tree/Tree.cy.tsx
```diff
@@ -7,7 +7,7 @@ import { treeNodesMock } from '@components/Tree/utils/mocks';
 
 const TreeComponent = ({ onSelect }: { onSelect?: (id: Nullable<string>) => void }) => {
     return (
-        <Tree id="treeId" onSelect={onSelect}>
+        <Tree id="treeId" onSelect={onSelect} baseItemPadding={{ x: 50, y: 20 }}>
```
SamuelAlev, line 10: As we are the only one to use the tree for now, we could set it as default wdyt?
Designers should align that, having different paddings on tree is weird.

SamCreasey: Already aligned, custom is the way forwards since it's used in Analytics

src/components/Tree/Tree.stories.tsx
```diff
@@ -15,6 +15,7 @@ export default {
     tags: ['autodocs'],
     args: {
         id: 'treeId',
+        baseItemPadding: { x: 10, y: 8 },
```
SamuelAlev, line 18:
```suggestion
        baseItemPadding: { top: 8, right: 10, bottom: 8, left: 10 },
```

src/components/Tree/types.ts
```diff
@@ -18,8 +18,12 @@ export type TreeProps = {
     onExpand?: OnExpandCallback;
     onDrop?: OnDropCallback<{ id: string; sort: number }>;
     children: ReactNode;
+    /** Padding added to each TreeItem in the TreeContext. Default: {"top": 10, "right": 8, "bottom": 10, "left": 8} */
```
SamuelAlev, line 21:
```suggestion
    /** 
      * Padding added to each TreeItem in the TreeContext.
      * @default {"top": 10, "right": 8, "bottom": 10, "left": 8}
      */
```
(Indentation might be incorrect in the suggestion)
=> https://jsdoc.app/tags-default.html

SamuelAlev (approved):

src/components/Tree/types.ts
SamCreasey: fixed

src/components/Tree/Tree.stories.tsx
SamCreasey: fixed

GoranRibic (approved):

PR #1277, opened 2023-03-01
refactor: use optimized path

SamuelAlev:
Use the `/Generated` path whereever possible to have the cached bundled available
Use regex instead of 2 replace in enum generation

### DIFF
--- a/scripts/generateIconEnum.ts
+++ b/scripts/generateIconEnum.ts
@@ -7,7 +7,7 @@ const ICON_ENUM_OUTPUT_PATH = './src/foundation/Icon/IconEnum.ts';
 
 (async () => {
     const iconFilePath = await fastGlob('./src/foundation/Icon/Generated/*.tsx', { objectMode: true });
-    const iconNames = iconFilePath.map((path) => path.name.replace(/^Icon|\.tsx$/g, ''));
+    const iconNames = iconFilePath.map((path) => path.name.replace(/(^Icon)|(\.tsx$)/g, ''));
 
     const iconsEnumString = `
         export enum IconEnum {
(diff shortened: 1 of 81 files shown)


fabriciomendonca (approved):
lg2m

scripts/generateIconEnum.ts
```diff
@@ -7,7 +7,7 @@ const ICON_ENUM_OUTPUT_PATH = './src/foundation/Icon/IconEnum.ts';
 
 (async () => {
     const iconFilePath = await fastGlob('./src/foundation/Icon/Generated/*.tsx', { objectMode: true });
-    const iconNames = iconFilePath.map((path) => path.name.replace('Icon', '').replace('.tsx', ''));
+    const iconNames = iconFilePath.map((path) => path.name.replace(/^Icon|\.tsx$/g, ''));
```
fabriciomendonca, line 10: Sonarcloud is complaining about the regex, this should fix `/(^Icon)|(\.tsx$)/g`

fabriciomendonca (changes requested):

src/components/Trigger/Trigger.tsx
```diff
@@ -7,7 +7,7 @@ import { FOCUS_STYLE, FOCUS_VISIBLE_STYLE } from '@utilities/focusStyle';
 import { merge } from '@utilities/merge';
 import { Validation, validationClassMap } from '@utilities/validation';
 import React, { FC, HTMLAttributes } from 'react';
-import { IconCross, IconExclamationMarkTriangle, IconTrashBin } from '@foundation/Icon';
+import { IconCross, IconExclamationMarkTriangle, IconTrashBin } from '@foundation/Icon/Generated';
```
fabriciomendonca, line 10: Line 3 is also importing from `/Generated`

fabriciomendonca (changes requested):

fabriciomendonca (approved):

PR #1281, opened 2023-03-06
fix(Tooltip): Tooltip text was not appearing in template library

dominiquemb:


fabriciomendonca (approved):
lg2m
We had to change to `span` because there is a CSS rule being applied on the Template Editor that sets the `font-size: 0` and `line-height: 0` to `p` elements and we didn't know if it would break any other paragraph on that page if we changed that CSS selector. @imoutaharik, WDYT?

fabriciomendonca (changes requested):
Removing my approval temporarily since we will need a different solution for the tooltip in the Template Editor. 
This change is not dependent of the one in Clarify, but I believe it is better to see how they are going to handle the tooltip there since the element `z-index` cannot be changed

fabriciomendonca (approved):

PR #1288, opened 2023-03-09
chore(Tooltip): update tooltip tokens

dominiquemb:


src/components/Tooltip/BrightHeader.tsx
```diff
@@ -18,17 +18,17 @@ export enum BrightHeaderStyle {
 }
 
 export const brightHeaderBackgroundColors: Record<BrightHeaderStyle, string> = {
-    [BrightHeaderStyle.Information]: 'tw-bg-violet-60',
-    [BrightHeaderStyle.Warning]: 'tw-bg-red-60',
-    [BrightHeaderStyle.Tip]: 'tw-bg-green-60',
-    [BrightHeaderStyle.Note]: 'tw-bg-yellow-60',
+    [BrightHeaderStyle.Information]: 'tw-bg-box-selected-strong',
+    [BrightHeaderStyle.Warning]: 'tw-dark tw-bg-button-danger-background',
```
fabriciomendonca, line 22: Should we use a `button` token for non-button elements? That doesn't look right to me...

dominiquemb: That's what the design is using. I agree though, maybe for consistency we should use what the Toast is using for its red background color. Wdyt?

fabriciomendonca: I would go for the same solution used in the Toast.

fabriciomendonca (approved):

PR #1289, opened 2023-03-09
feat(RadioList): Introducing RadioList component 

selimovicz:
The design for the component can be found [here](https://www.figma.com/file/WfEbqZI0nnPWHN43MCtvnm/Fondue---Inputs?node-id=107-27120&t=FcNttRiqe7jlCkuQ-0).

### DIFF
--- a/src/components/Accordion/AccordionHeader.tsx
+++ b/src/components/Accordion/AccordionHeader.tsx
@@ -29,7 +29,7 @@ export const AccordionHeader: FC<AccordionHeaderProps> = ({
                 ])}
             >
                 {isValidElement(decorator) && (
-                    <span data-test-id="accordion-header-decorator" className="tw-shrink-0">
+                    <span data-test-id="accordion-header-decorator" className="tw-shrink-0 tw-leading-4">
                         {cloneElement(decorator as ReactElement<IconProps>, {
                             size: IconSize.Size16,
                         })}


fabriciomendonca (commented):
I tried changing the Default Value in storybook and it didn't change the selection, is that intended if I have already interacted with the list or is it a bug?

selimovicz:
> I tried changing the Default Value in storybook and it didn't change the selection, is that intended if I have already interacted with the list or is it a bug?

It was a bug, now it's fixed. Sorry about that.

src/components/RadioList/RadioButton.tsx
```diff
@@ -0,0 +1,35 @@
...
+    Weak = 'Weak',
+}
+
+export type RadioButtonProps = {
+    value: string;
+    disabled?: boolean;
+    hideLabel?: boolean;
+    emphasis?: RadioButtonEmphasis;
+    helperText?: string;
+};
```
imoutaharik, line 18: I believe the Tooltip is missing on the RadioButton, so the Required state with the little red *

imoutaharik (commented):
Thanks Zeid! This look good 🙌🏼 , I just left a few comments.
I was wondering, it would be also nice to have a Story for the RadioButton itself so we can play and have better understanding of its API. 
Selected or not
Disabled unselected, disabled selected.
required or not
Emphasis
HideLabel and with Tooltip.

WDYT?
And I would also have a quick review from @LidijaStoilova as this component needed to be implemented to fix some a11y issues in the platform. 😄

src/components/RadioList/RadioButton.tsx
LidijaStoilova: I did a quick check and it seems like the keyboard navigation is not working properly. Also a few roles, states & properties are missing, see this [W3C documentation](https://www.w3.org/WAI/ARIA/apg/patterns/radio/). Also I was wondering if we should call it RadioGroup instead of RadioList? Seems more common

imoutaharik (commented):
It looks good to me. Thanks for implementing! 🎉 

I just have a question about the tooltips and keyboard navigation. @LidijaStoilova when Tooltip are present, they should be 'tabable' and close on blur?

LidijaStoilova:
Yes, tooltips should be tabbable @imoutaharik @selimovicz  and close when you tab away or press ESC

imoutaharik (approved):
Thank you! On my side, all look good! 🙌🏼 
Everything ok for you @LidijaStoilova ?

LidijaStoilova:
> It looks good to me. Thanks for implementing! 🎉
> 
> I just have a question about the tooltips and keyboard navigation. @LidijaStoilova when Tooltip are present, they should be 'tabable' and close on blur?

I think it looks good from my side as well

PR #1291, opened 2023-03-09
feat(Tooltip): close when user enters ESC key

dominiquemb:


src/components/Tooltip/Tooltip.tsx
```diff
@@ -266,6 +258,14 @@ export const Tooltip = ({
...
     }, [isOpen]);
 
+    useEffect(() => {
+        window.addEventListener('keydown', (event) => {
+            if (isOpen && event.key === 'Escape') {
+                setIsOpen(false);
+            }
+        });
+    });
+
```
fabriciomendonca, line 268: I would put the callback into a constant and then removing the event from the window object for that callback. Otherwise, a new event will be added every time the component renders and all of them will be called when pressing the Esc key.

```typescript
useEffect(() => {
  window.addEventListener('keydown', handleCloseTooltipOnEscape);

  return () => {
    window.removeEventListener('keydown', handleCloseTooltipOnEscape);
  }
}, []);
```

fabriciomendonca (changes requested):

src/components/Tooltip/Tooltip.tsx
dominiquemb: Added

fabriciomendonca (approved):

PR #1292, opened 2023-03-09
chore(storybook): restyled storybook UI

thisguyscode:
Hey @imoutaharik - I was getting frustrated by how Storybook looks so I did some hacky overrides to Storybook's UI. Does it look acceptable to you (i.e. not tooooo hacky)?

Also, for some reason storybook previews don't render for me locally. I tried on main and beta with none of thee changes and had the same issue. Does it work for you?


thisguyscode:
Closed as I didn't consider dark mode

imoutaharik:
I see, It doesn't really work in dark mode. That's a pity I really liked the new style. 😄

PR #1293, opened 2023-03-09
feat(RichTextEditor): Add link selector

triggertoo:


src/components/RichTextEditor/utils/isValidUrl.ts
```diff
@@ -1,6 +1,10 @@
 /* (c) Copyright Frontify Ltd., all rights reserved. */
 
 export const isValidUrl = (url: string): boolean => {
+    if (url.charAt(0) === '/') {
```
celine-s, line 4: could we use startswidth '/r/'? Or does it sometimes not start with this?

src/components/RichTextEditor/Plugins/LinkPlugin/utils/getLinkNodeByProp.ts
```diff
@@ -0,0 +1,24 @@
...
+import { LinkNode } from '../../types';
+
+type SupportedNodeProps = Pick<LinkNode, 'id' | 'url'>;
+
+export const getLinkNodeByProp = (
+    nodeProp: keyof SupportedNodeProps,
+    value: string,
+    nodes: LinkNode[],
+): LinkNode | null => {
+    for (const node of nodes) {
```
celine-s, line 12: could a 'find' be used here?

celine-s (commented):
its amazing!! Thanks!! enter behaviour is not working properly on my side. Would be nice if you could test it as well.

PR #1296, opened 2023-03-10
chore(datepicker): QoL improvements 

joseazevedo88:
Add prevent open on focus optional prop and onKeyDown callback to control open/close with ref

### DIFF
--- a/src/components/DatePicker/DatePicker.stories.tsx
+++ b/src/components/DatePicker/DatePicker.stories.tsx
@@ -1,8 +1,8 @@
 /* (c) Copyright Frontify Ltd., all rights reserved. */
 
-import React, { useState } from 'react';
+import React, { useRef, useState } from 'react';
 import { Meta, StoryFn } from '@storybook/react';
-import { DatePicker, DatePickerProps } from './DatePicker';
+import { DatePicker, DatePickerProps, ReactDatePickerRef } from './DatePicker';
 import { FormControl } from '@components/FormControl';
 import { Slider } from '@components/Slider';
 import { Validation } from '@utilities/validation';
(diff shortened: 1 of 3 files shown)


src/components/DatePicker/DatePicker.stories.tsx
```diff
@@ -115,6 +116,13 @@ const TemplateDateRange: StoryFn<DatePickerProps> = () => {
             minDate={subDays(new Date(), 40)}
             maxDate={new Date()}
             variant="range"
+            ref={testRef}
```
imoutaharik, line 119: Can we change the name testRef?

imoutaharik (commented):
Looks good to me.

I just noticed something about the keyboard navigation, when the range is big (2 months for example) and you tab on the Datepicker, it's kind of hard to interact with it.
You need to use the << <  > >> buttons to find the initial date so you can finally interact with the dates.

src/components/DatePicker/DatePicker.stories.tsx
joseazevedo88: Sure, leftover

joseazevedo88:
> Looks good to me.
> 
> I just noticed something about the keyboard navigation, when the range is big (2 months for example) and you tab on the Datepicker, it's kind of hard to interact with it. You need to use the << < > >> buttons to find the initial date so you can finally interact with the dates.

Yup, added a "hacky" way of doing it in the stories, manipulating the ref, as I'll need it also for clarify. The lib is somewhat outdated and doesn't seem to be maintained anymore, some things are not up to date in terms of react standards.

imoutaharik (approved):

PR #1299, opened 2023-03-13
fix: remove @udecode/plate-core in favour of @udecode/plate

triggertoo:


src/components/RichTextEditor/serializer/nodes/link.ts
```diff
@@ -2,7 +2,7 @@
 
 import { TLinkElement } from '@components/RichTextEditor/Plugins/LinkPlugin/types';
 import { DesignTokens } from '@components/RichTextEditor/types';
-import { TElement } from '@udecode/plate-core';
+import { TElement } from '@udecode/plate';
```
SamuelAlev, line 5: Shouldn't we keep `-core` for treeshaking?

triggertoo: tree shaking is given for both packages, since both provide ESM

triggertoo: it's related to the features used, where many are not part of `@udecode/plate-core` and would bloat the package.json

bojangles-m (approved):

SamuelAlev (approved):

PR #1303, opened 2023-03-14
fix: tooltip race condition

SamCreasey:
Fixes a missing tooltip race condition caused by the setTimeout being triggered  just before the tooltip was disabled, which sets the tooltip back to 'open' even if it is disabled


fabriciomendonca (approved):
Good catch

PR #1305, opened 2023-03-14
chore(Switch)!: deprecate large size

dominiquemb:


src/components/Switch/Switch.tsx
```diff
@@ -12,13 +12,11 @@ export const SWITCH_ID = 'switch-container';
 export enum SwitchSize {
     Small = 'Small',
     Medium = 'Medium',
-    Large = 'Large',
```
fabriciomendonca, line 15: Isn't this a breaking change?

fabriciomendonca (approved):

PR #1307, opened 2023-03-14
fix(RTE): remove unwanted autoscrolling behaviour

Walkeezy:
Pass in empty function `scrollSelectionIntoView` to prevent the [default of slate](https://github.com/ianstormtaylor/slate/blob/1d8010be8500ef5c98dbd4179354db1986d5c8f4/packages/slate-react/src/components/editable.tsx#L1689), which is causing issues with unwanted page jumps when the RichTextEditor is used inside Frontify Guideline Blocks.

Ticket: https://app.clickup.com/t/85zrqj46c


celine-s (approved):
nice! 🚀

PR #1308, opened 2023-03-14
refactor(Tree)!: dnd kit

GoranRibic:
Refactor Tree Component to use dnd kit library

**BREAKING CHANGES**
- removed `baseItemPadding` prop from Tree
- changed `accepts` type of TreeItem to `string[]`; additionally, if a string inside accepts contains `-within` it is considered to be movable to an inner level of an item
- remove the option for `contentComponent` to be a function (should be a ReactNode now)
- `DropzonePosition` renamed to `CollisionPosition`
- added optional prop `direction: 'up' | 'down';` to `onDrop` handlers


src/components/Tree/TreeItem.tsx
```diff
@@ -1,311 +1,378 @@
...
     );
 
-    const handleExpandClick = useCallback(
-        (event: MouseEvent) => {
-            event.stopPropagation();
-            onExpand(id, !treeState.expandedIds.has(id));
-        },
-        [id, onExpand, treeState.expandedIds],
-    );
+    const handleExpand = (event: React.MouseEvent<HTMLButtonElement, MouseEvent>) => {
```
SamuelAlev, line 148: Same here

src/components/Tree/TreeItem.tsx
```diff
@@ -1,311 +1,378 @@
...
-                        break;
-
-                    case 'ArrowDown':
-                        event.preventDefault();
-                        if (treeItemState) {
-                            const nextItemToFocus = getNextItemToFocus(id, treeItemState, treeState);
-                            nextItemToFocus?.domElement?.focus();
-                        }
-                        break;
+        (event: React.KeyboardEvent<HTMLDivElement>) => {
```
SamuelAlev, line 155: Same here

src/components/Tree/utils/flattenChildren.ts
```diff
@@ -16,3 +16,30 @@ export const flattenChildren = (children?: ReactNode) => {
...
+        return [];
+    }
+
+    if (Array.isArray(node)) {
+        return node.flatMap(flattenTree);
+    }
+
+    if (typeof node !== 'object' || !('props' in node)) {
+        if (isValidElement(node)) {
+            return [node as React.ReactElement];
```
SamuelAlev, line 32: Same here

src/components/Tree/utils/flattenChildren.ts
```diff
@@ -16,3 +16,30 @@ export const flattenChildren = (children?: ReactNode) => {
 
     return result.filter(Boolean);
 };
+
+// ! TODO: delete this if not used
```
SamuelAlev, line 20: :)

SamuelAlev:
Animation is a bit weird, it fades in the whole subtree instead of making the height progressive
Screenshot 2023-03-14 at 17 23 45

src/components/OrderableList/OrderableList.tsx
```diff
@@ -58,7 +59,7 @@ export const OrderableList = <T extends object>({
                             key={`orderable-list-item-${index}-before`}
                             data={{
                                 targetItem: item,
-                                position: DropZonePosition.Before,
+                                position: 'before' as const,
```
SamuelAlev, line 62: Why the `as const` and not `as CollisionPosition`

src/components/Tree/Tree.tsx
```diff
@@ -10,20 +24,27 @@ import type {
...
+
 const reducer = (state: TreeState, action: TreeStateAction): TreeState => {
     switch (action.type) {
         case 'SET_SELECT': {
             const { id, isSelected } = action.payload;
 
             if (!state.items.has(id)) {
-                throw new Error(`No tree item registered with id ${id}.`);
+                console.error(`No tree item registered with id ${id}.`);
+                return state;
```
SamuelAlev, line 47: It shouldn't be the case, if so there is an error introduced

src/components/Tree/Tree.tsx
```diff
@@ -65,28 +86,32 @@ const reducer = (state: TreeState, action: TreeStateAction): TreeState => {
 
         case 'ON_DROP': {
             const { targetId, id, position } = action.payload;
+
             const targetState = state.items.get(targetId);
 
             if (!targetState) {
-                throw new Error(`No tree item registered with id ${targetId}.`);
+                console.error(`No tree item registered with id ${targetId}.`);
+                return state;
```
SamuelAlev, line 94: Same here

src/components/Tree/Tree.tsx
```diff
@@ -95,15 +120,19 @@ const reducer = (state: TreeState, action: TreeStateAction): TreeState => {
...
+                    console.error(`No tree item registered with id ${id ?? sourceState?.parentId}.`);
+                    return state;
                 }
                 const targetState = state.items.get(targetId);
                 const oldParentState = state.items.get(sourceState.parentId);
+
                 if (!targetState || !oldParentState) {
-                    throw new Error(`No tree item registered with id ${targetId ?? sourceState.parentId}.`);
+                    console.error(`No tree item registered with id ${targetId ?? sourceState.parentId}.`);
+                    return state;
```
SamuelAlev, line 135: Same here

src/components/Tree/helpers/getNextItemToFocus.ts
```diff
@@ -1,13 +1,14 @@
 /* (c) Copyright Frontify Ltd., all rights reserved. */
 
+import { UniqueIdentifier } from '@dnd-kit/core';
 import type { TreeItemState, TreeState } from '../types';
 
 export const getNextItemToFocus = (
-    id: string,
-    currentItemState: TreeItemState,
+    id: UniqueIdentifier,
+    currentItemState: TreeItemState | undefined,
```
SamuelAlev, line 8:
```suggestion
    currentItemState: TreeItemState,
```
should never be `undefined`

src/components/Tree/utils/collisionDetection.ts
```diff
@@ -0,0 +1,46 @@
...
+
+    const draggableRect = translated ?? initial;
+
+    if (!draggableRect || !droppableRect) {
+        return closestCornersCollisions;
+    }
+
+    const position = calculatePosition(draggableRect as DOMRect, droppableRect);
+
+    /* mutation - for performance reasons */
```
SamuelAlev, line 42: Really vague explanation

src/utilities/dnd/moveItems.ts
```diff
@@ -1,16 +1,17 @@
...
+const isPreviousItemWithoutSortValue = (index: number, movedItemTargetIndex: number, sort: Nullable<number>) =>
     index < movedItemTargetIndex && sort === null;
 const isMovedItem = (index: number, movedItemTargetIndex: number) => index === movedItemTargetIndex;
 const isFollowingItemWithSortLowerOrEqualToPreviousItem = (
     index: number,
     movedItemTargetIndex: number,
-    sort: NullableNumber,
-    previousSort: NullableNumber,
+    sort: Nullable<number>,
+    previousSort: Nullable<number>,
```
SamuelAlev, line 14: Let's remove the nullable on those props and give them defaults

src/utilities/dnd/moveItems.ts
```diff
@@ -1,16 +1,17 @@
 /* (c) Copyright Frontify Ltd., all rights reserved. */
 
-import { DraggableItem, DropZonePosition } from '@utilities/dnd/types';
+import { CollisionPosition } from '@components/Tree';
+import { DraggableItem } from '@utilities/dnd/types';
 
-const isFirstItemWithoutSortValue = (index: number, sort: NullableNumber) => index === 0 && sort === null;
-const isPreviousItemWithoutSortValue = (index: number, movedItemTargetIndex: number, sort: NullableNumber) =>
+const isFirstItemWithoutSortValue = (index: number, sort: Nullable<number>) => index === 0 && sort === null;
+const isPreviousItemWithoutSortValue = (index: number, movedItemTargetIndex: number, sort: Nullable<number>) =>
```
SamuelAlev, line 7: Same here

src/utilities/dnd/types.ts
```diff
@@ -1,13 +1,7 @@
...
-export enum DropZonePosition {
-    After = 'after',
-    Before = 'before',
-    Within = 'within',
-}
-
 export type DraggableItem<T = Record<string, unknown>> = T & {
     id: string;
-    sort: NullableNumber;
+    sort: Nullable<number>;
```
SamuelAlev, line 5: Same here
```suggestion
    sort: number;
```

SamuelAlev (changes requested):

src/components/OrderableList/OrderableList.tsx
GoranRibic: As it is an untyped object value, and value, in this case, for the position is 'before' so `const` is correct. If I were to give a type to the object I'm passing then it would be `CollisionPosition` and there would be no reason for me to do `as` at all.

src/components/Tree/helpers/getNextItemToFocus.ts
GoranRibic: Maybe logically it can't but technically it can be undefined as it is an item state, anyway this way we reduce cognitive complexity and lose nothing as functions were returning undefined even before this change. In some future iterations maybe we can do a proper clean code refactor.

src/components/Tree/utils/collisionDetection.ts
GoranRibic: That is because the only info that is needed is that object is being mutated so the comment's whole purpose is there to alert the reader where it happens. I can add ` /* mutation - for performance reasons (avoiding iterations) */`

src/utilities/dnd/moveItems.ts
GoranRibic: Yeah to me makes no sense to be nullable as well, for some reason this was done because BE can return a null value for sort values. But Tree Fondue should not care about it anyway. I'll remove null and if something breaks we can fix it down the road.

GoranRibic: Okay, this is used in OrderableList so we kinda a need to leave it as we will break things in Clarify where it is used. Probably we should do this change in separate PR as it is a breaking change that affects people outside the NM team. Wyt?

SamuelAlev: They are passing `number` everywhere, and if it is breaking their tests should kick in ;)

GoranRibic:
> Animation is a bit weird, it fades in the whole subtree instead of making the height progressive Screenshot 2023-03-14 at 17 23 45

I know, known issue. Change in height introduces other problems, as I already spent some time on it without success. Decided to leave it at this for now.

src/utilities/dnd/moveItems.ts
GoranRibic: Haha okay. Nullable off you go.

src/components/Tree/Tree.tsx
GoranRibic: "This is the same thing as before, but it is failing gracefully/silently. This should be the case because if we throw an error, we will break the application. And that is a no-no to the end user."

src/components/Tree/utils/collisionDetection.ts
GoranRibic: Okay, can you give a suggestion on what would it make more clear to you?

src/components/Tree/helpers/getItemPositionInParent.ts
```diff
@@ -1,12 +1,22 @@
...
+/**
+ * The getItemPositionInParent function returns the index of a given item within a list of sibling nodes, based on the item's ID and its parent ID.
+ *
+ * @param item The item whose position within the parent node is required.
+ * @param item.id The unique identifier of the item.
+ * @param item.parentId The unique identifier of the item's parent.
+ * @param nodes An array of React elements representing the nodes in the tree structure.
+ *
+ * @returns
+ * A number representing the index of the given item within the list of its sibling nodes. If the item is not found, the function will return -1.
```
SamuelAlev, line 15:
```suggestion
 * @returns {number} The index of the given item within the list of its sibling nodes. If the item is not found, the function will return -1.
```

src/components/Tree/TreeItem/TreeItem.tsx
```diff
@@ -0,0 +1,332 @@
...
+        // eslint-disable-next-line jsx-a11y/click-events-have-key-events
+        <li
+            key={id}
+            tabIndex={0}
+            role="treeitem"
+            style={liStyle}
+            id={id.toString()}
+            aria-label={label}
+            aria-level={level}
+            // layoutId={String(id)}
```
SamuelAlev, line 298:
```suggestion
```

src/components/Tree/TreeItem/TreeItem.tsx
```diff
@@ -0,0 +1,332 @@
...
+        },
+    };
+
+    const animate = {
+        x: transform?.x,
+        y: transform?.y,
+    };
+
+    return (
+        // eslint-disable-next-line jsx-a11y/click-events-have-key-events
```
SamuelAlev, line 289: Please add the key event

src/components/Tree/helpers/getMovementAnnouncements.ts
```diff
@@ -0,0 +1,96 @@
...
+    parentId: UniqueIdentifier;
+};
+
+type AnnouncementArgs = {
+    eventName: string;
+    activeId: UniqueIdentifier;
+    overId?: UniqueIdentifier;
+    treeState: TreeState;
+    currentPosition: Nullable<{ overId: UniqueIdentifier; parentId: Nullable<UniqueIdentifier> }>;
+    setCurrentPosition: React.Dispatch<
```
SamuelAlev, line 21:
```suggestion
    setCurrentPosition: Dispatch<
```

SamuelAlev (changes requested):

src/components/index.ts
```diff
@@ -32,7 +32,6 @@ export * from './MenuItem';
 export * from './Modal';
 export * from './MultiInput';
 export * from './MultiSelect';
-export * from './OrderableList';
```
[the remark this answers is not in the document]
leko91: This is to be done at a future point, probably a mistake for now, I'll revert it

src/components/Tree/Tree.cy.tsx
```diff
@@ -1,24 +1,25 @@
...
+import { UniqueIdentifier } from '@dnd-kit/core';
 
 import { Tree, TreeItem } from '@components/Tree';
-import { treeNodesMock } from '@components/Tree/utils/mocks';
+import { treeItemsMock } from '@components/Tree/utils';
 
-const TreeComponent = ({ onSelect }: { onSelect?: (id: Nullable<string>) => void }) => {
+const TreeComponent = ({ onSelect }: { onSelect?: (id: Nullable<UniqueIdentifier>) => void }) => {
     return (
-        <Tree id="treeId" onSelect={onSelect} baseItemPadding={{ top: 10, right: 20, bottom: 30, left: 40 }}>
```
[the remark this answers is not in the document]
leko91: Not 100% sure, Goran removed it. I can readd it again if we find a need for it.

src/components/Tree/helpers/getMovementAnnouncements.ts
```diff
@@ -0,0 +1,95 @@
...
+
+        const previousItem = sortedItems[overIndex - 1];
+
+        let announcement;
+        const movedVerb = eventName === 'onDragEnd' ? 'dropped' : 'moved';
+        const nestedVerb = eventName === 'onDragEnd' ? 'dropped' : 'nested';
+
+        if (!previousItem) {
+            const nextItem = sortedItems[overIndex + 1];
+            announcement = `${activeId} was ${movedVerb} before ${nextItem.id}.`;
```
SamuelAlev, line 74: Same in this file the ID should be the text content first and fallback to id if there is no text

src/components/Tree/Tree.tsx
```diff
@@ -194,87 +141,55 @@ const reducer = (state: TreeState, action: TreeStateAction): TreeState => {
...
     onExpand,
-    onDrop,
     children,
-    multiselect = true,
-    baseItemPadding = {},
+    selectedIds,
+    expandedIds,
+    draggable = false,
+    multiselect = false,
+    'data-test-id': dataTestId = 'tree',
```
SamuelAlev, line 152:
```suggestion
    'data-test-id': dataTestId = 'fondue-tree',
```
Every default `data-test-id` should be prefixed by `fondue-`, comment is also targeting the other files from the tree

src/components/Tree/Tree.tsx
```diff
@@ -284,61 +199,352 @@ export const Tree = ({
...
+    );
+
+    const announcements: TreeAnnouncements = useMemo(
+        () => ({
+            onDragStart({ active }) {
+                const nodeTitle = treeState.nodes.find((node) => node.key === active.id)?.props.contentComponent.props
+                    .title;
+
+                console.log(`Picked up ${nodeTitle || active.id}.`);
+
```
SamuelAlev, line 439:
```suggestion
```

SamuelAlev (approved):

PR #1311, opened 2023-03-15
feat(RichTextEditor): disable link- and button-button when multiple blocks are selected

triggertoo:
Disables a link-button or button-button when multiple blocks are selected in the rich text editor, since it's not yet supported by the rich text editor plugin itself.

### DIFF
--- a/src/components/RichTextEditor/Plugins/LinkPlugin/LinkButton.tsx
+++ b/src/components/RichTextEditor/Plugins/LinkPlugin/LinkButton.tsx
@@ -3,19 +3,31 @@
 import { getHotkeyByPlatform } from '@components/RichTextEditor/helpers/getHotkeyByPlatform';
 import { getTooltip } from '@components/RichTextEditor/helpers/getTooltip';
 import { IconLink16 } from '@foundation/Icon/Generated';
-import { LinkToolbarButton } from '@udecode/plate';
+import { LinkToolbarButton, isRangeInSameBlock, useEventPlateId, usePlateEditorState } from '@udecode/plate';
 import React from 'react';
-import { ButtonWrapper, IconStylingWrapper, buttonClassNames, buttonStyles } from '../helper';
+import { ButtonWrapper, IconStylingWrapper, buttonStyles, getButtonClassNames } from '../helper';
 import { PluginButtonProps } from '../types';
 
-export const LinkButton = ({ id }: PluginButtonProps) => (
-    <ButtonWrapper id={id}>
-        <LinkToolbarButton
-            tooltip={getTooltip(`Link\n${getHotkeyByPlatform('Ctrl+K')}`)}
-            icon={<IconStylingWrapper icon={<IconLink16 />} />}
-            classNames={buttonClassNames}
-            styles={buttonStyles}
-            actionHandler="onMouseDown"
-        />
-    </ButtonWrapper>
-);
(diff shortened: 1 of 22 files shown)


src/components/RichTextEditor/Plugins/LinkPlugin/LinkButton.tsx
```diff
@@ -3,19 +3,30 @@
...
+        at: editor.selection,
+    });
+
+    return (
+        <ButtonWrapper id={id}>
+            <LinkToolbarButton
+                tooltip={getTooltip(`Link\n${getHotkeyByPlatform('Ctrl+K')}`)}
+                icon={<IconStylingWrapper icon={<IconLink16 />} />}
+                classNames={{
+                    root: `${buttonClassNames.root} ${isInSameBlock ? '' : '!tw-cursor-not-allowed !tw-opacity-50'}`,
```
celine-s, line 23: maybe we could also add `!tw-cursor-not-allowed !tw-opacity-50` to the buttonClassNames such that we can reuse it also like buttonClassNames.disabled. wdyt?

src/components/RichTextEditor/Plugins/LinkPlugin/LinkButton.tsx
```diff
@@ -3,19 +3,30 @@
...
+export const LinkButton = ({ id, editorId }: PluginButtonProps) => {
+    const editor = usePlateEditorState(useEventPlateId(editorId));
+    const isInSameBlock = isRangeInSameBlock(editor, {
+        at: editor.selection,
+    });
+
+    return (
+        <ButtonWrapper id={id}>
+            <LinkToolbarButton
+                tooltip={getTooltip(`Link\n${getHotkeyByPlatform('Ctrl+K')}`)}
```
celine-s, line 20:
```suggestion
                tooltip={getTooltip(isInSameBlock ?`Link\n${getHotkeyByPlatform('Ctrl+K')}` : 'Cannot set Links to multiple paragraphs')}
```

celine-s: wdyt to adapting the tooltip msg?

src/components/RichTextEditor/Plugins/LinkPlugin/LinkButton.tsx
triggertoo: good point!

src/components/RichTextEditor/__tests__/fixtures/selectors.ts
```diff
@@ -20,3 +20,4 @@ export const FLOATING_BUTTON_EDIT = '[data-test-id=floating-button-edit]';
 export const FLOATING_BUTTON_SECONDARY = '[data-test-id=floating-button-insert-secondary]';
 export const BUTTON = '[data-test-id=button]';
 export const CHECKBOX_INPUT_ID = '[data-test-id=checkbox-input]';
+export const TOOLBAR_BUTTON = '[data-testid=ToolbarButton]';
```
celine-s, line 23: why do you need it? if so, I think the `-` is missing between test & id.?

celine-s, line 23:
```suggestion
export const TOOLBAR_BUTTON = '[data-test-id=ToolbarButton]';
```

celine-s (approved):
Nice!! 🚀

PR #1312, opened 2023-03-15
fix(Switch): adding hover state to tooltip icon

dominiquemb:


src/components/TooltipIcon/TooltipIcon.tsx
```diff
@@ -22,9 +22,15 @@ export enum TooltipIconTriggerStyle {
 }
 
 const tooltipTriggerStyleClass: Record<TooltipIconTriggerStyle, string> = {
-    [TooltipIconTriggerStyle.Danger]: 'tw-text-box-negative-strong hover:tw-text-box-negative-strong-hover',
-    [TooltipIconTriggerStyle.Warning]: 'tw-text-box-warning-strong hover:tw-text-box-warning-strong-hover',
-    [TooltipIconTriggerStyle.Primary]: 'tw-text-text-weak hover:tw-text-text-x-weak',
+    [TooltipIconTriggerStyle.Danger]: 'tw-text-box-negative-strong',
```
fabriciomendonca, line 25: Is it correct to use `box-*` tokens other than the `inverse` ones for text?

dominiquemb: You're right. I've updated these with the inverse versions.

fabriciomendonca (approved):

PR #1313, opened 2023-03-15
fix(Switch): using alias tokens for label

dominiquemb:


src/components/InputLabel/InputLabel.tsx
```diff
@@ -65,7 +63,7 @@ export const InputLabel: FC<InputLabelProps> = ({
             {required && (
                 <span
                     data-test-id="input-label-required"
-                    className="tw-h-4 tw-text-m tw-text-red-60 dark:tw-text-red-50"
+                    className="tw-h-4 tw-text-m tw-text-box-negative-strong-pressed"
```
fabriciomendonca, line 66: Same question here about the usage of `box-*` tokens without the `inverse` suffix for text

dominiquemb: After comparing with the tokens used for error text in other components, it seems that I should be using `tw-text-text-negative` instead, so I've updated it with that.

fabriciomendonca (approved):

PR #1316, opened 2023-03-17
fix(Accordion): accessibility errors

fabriciomendonca:


src/components/Accordion/AccordionHeader.tsx
```diff
@@ -1,27 +1,29 @@
...
-    as: Heading = 'div',
-}) => {
+    as: Heading = 'span',
+}: AccordionHeaderProps) => {
     const icon = <AccordionHeaderIcon isOpen={isOpen} disabled={disabled} size={size} type={type} />;
     return (
-        <div data-test-id="accordion-header" className="tw-px-8 tw-py-6">
-            <header
+        <span data-test-id="accordion-header" className="tw-block tw-px-8 tw-py-6">
+            <span
```
SamuelAlev, line 23: This is not correct HTML, `span > div` is not allowed

src/components/Accordion/AccordionHeader.tsx
```diff
@@ -1,27 +1,29 @@
...
                 data-test-id="fieldset-header"
+                role="navigation"
+                aria-label={`${children}`}
```
SamuelAlev, line 26: Not correct as the type is `ReactNode`, you would have either `undefined`, either `false`/`true`, either `[Object object]`, ...
You need to compute it

src/components/Accordion/AccordionHeader.tsx
fabriciomendonca: The children prop in `AccordionHeaderProps` is defined as string as you can see here
```
export type AccordionHeaderProps = {
    decorator?: ReactNode;
    children: string;
    disabled?: boolean;
    isOpen: boolean;
    /** @deprecated use headerComponent prop for custom header styles */
    size?: AccordionHeaderIconSize | FieldsetHeaderSize;
    /** @deprecated use headerComponent prop for custom header styles */
    bold?: boolean;
    /** @deprecated Icon type will be fixed to FieldsetHeaderType.Accordion.
     Use headerComponent prop for custom header styles */
    type?: FieldsetHeaderType;
    /** @deprecated use headerComponent prop for custom header styles */
    as?: keyof JSX.IntrinsicElements;
};
```

src/components/Accordion/AccordionHeader.tsx
fabriciomendonca: Same answer as the previous one.
The `children` prop is defined as `string` here https://github.com/Frontify/fondue/blob/beta/src/components/Accordion/types.ts#L38

src/components/Accordion/AccordionHeader.tsx
fabriciomendonca: I changed to span because on the `AccordionItem` the header is rendered inside a `button` and `button > div` is also not valid HTML

src/components/Accordion/AccordionHeader.tsx
fabriciomendonca: But it seems we're going to have this issue because of this:
https://github.com/Frontify/clarify/blob/master/application/client/Feature/Publishing/Controls/ImageGalleryControl/List/AccordionHeader/AccordionHeader.tsx#L21

src/components/Accordion/AccordionHeader.tsx
fabriciomendonca: https://github.com/Frontify/fondue/blob/beta/src/components/Accordion/Accordion.tsx#L37

src/components/Accordion/AccordionHeader.tsx
fabriciomendonca: ☝️ Actually not, because it is rendered a custom `AccordionHeader` component, not the Fondue one. But there will be the invalid HTML tree `button > div`.

fabriciomendonca: Anyway, I created a state ref for the former `Heading` (which was coming from the deprecated `as` property and will be removed from the `AccordionHeaderProps` in another PR) and added as ref to the new `span` tag which will render `children` and I'm getting the `innerText` of that element.

LidijaStoilova:
Looks good from an accessibility point of view. Thanks @fabriciomendonca!

PR #1317, opened 2023-03-17
refactor(Switch)!: add indeterminate state and change enums to union type

dominiquemb:


src/components/Switch/Switch.tsx
```diff
@@ -14,6 +14,8 @@ export enum SwitchSize {
     Medium = 'Medium',
 }
 
+export type SwitchMode = 'On' | 'Off' | 'Indeterminate';
```
SamuelAlev, line 17:
```suggestion
export type SwitchMode = 'on' | 'off' | 'indeterminate';
```

src/components/Switch/Switch.tsx
SamuelAlev, line 1: You can also change `SwitchSize` to a union type

SamuelAlev (changes requested):

src/components/Accordion/Accordion.stories.tsx
```diff
@@ -459,8 +459,8 @@ export const WithAdvancedFormControls: StoryFn<AccordionProps> = () => {
                     extra={
                         <Switch
                             label="advanced"
-                            on={showAdvancedSpacing}
-                            size={SwitchSize.Small}
+                            mode={showAdvancedSpacing ? 'on' : 'off'}
+                            size={'small'}
```
SamuelAlev, line 463:
```suggestion
                            size="small"
```
Same for the other occurences in the PR

src/components/Switch/Switch.stories.tsx
```diff
@@ -1,22 +1,28 @@
...
 import { IconExclamationMarkCircle } from '@foundation/Icon/Generated';
-import { Switch, SwitchLabelStyle, SwitchProps, SwitchSize } from './Switch';
+import { Switch, SwitchMode, SwitchProps } from './Switch';
 import { TooltipIconTriggerStyle } from '@components/TooltipIcon';
 
+const SwitchSizeTypes = ['small', 'medium'];
+
+const SwitchModeTypes = ['on', 'off', 'indeterminate'];
+
+const SwitchLabelStyleTypes = ['default', 'heading'];
```
SamuelAlev, line 13: Please type the arrays (so in the future we know if we miss a field)

src/components/Switch/Switch.stories.tsx
```diff
@@ -26,35 +32,50 @@ export default {
...
-    const [on, setOn] = useState(args.on);
-    const toggle = () => setOn(!on);
-
+    const [mode, setMode] = useState<SwitchMode>(args.mode || 'off');
     useEffect(() => {
-        setOn(args.on);
-    }, [args.on]);
+        if (args.mode !== 'on' && args.mode !== 'off' && args.mode !== 'indeterminate') {
+            return;
+        }
```
SamuelAlev, line 61: why this condition, it should be one of `SwitchModeTypes` already 🤔

src/components/Switch/Switch.tsx
```diff
@@ -69,37 +67,65 @@ export const Switch: FC<SwitchProps> = ({
...
+        const valueClasses = mode === 'indeterminate' ? 'tw-bg-base' : trueOrFalseValueClasses;
+
+        const disabledClasses = disabled && mode === 'indeterminate' ? 'tw-bg-text-disabled' : valueClasses;
+
+        return merge([
+            baseClasses,
+            mode === 'indeterminate' ? indeterminateLineClasses : dotSizeClasses[size].dimensions,
+            disabledClasses,
+        ]);
+    }, [mode, size]);
```
SamuelAlev, line 107: `React Hook useMemo has a missing dependency: 'disabled'. Either include it or remove the dependency array`

src/components/Switch/Switch.stories.tsx
```diff
@@ -26,35 +32,50 @@ export default {
...
+            control: { type: 'radio' },
+        },
+        onChange: {
+            table: { disable: false },
+            action: 'Change',
+        },
     },
     args: {
-        on: true,
+        mode: 'on',
```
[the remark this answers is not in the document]
dominiquemb: Yes this is a known issue, it's the same in Checkbox story. @fabriciomendonca do you think it's worth trying to fix this?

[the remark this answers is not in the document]
fabriciomendonca: I found a solution for this and believed it was introducing another issue on Storybook: code snippets showing the story code instead of the component usage in some cases. It happens in some components and in the first render of the page, refreshing it solves the issue. 

I will bring the topic and the solution for the controls and code snippets sync into discussion internally and if we decide to use it, I will recreate the commit with the fix: adding `@storybook/preview-api@7.0.0-beta.53` and using the `useArgs` hook instead of `useState` and `useEffect` to update the `mode` value. [Check this commit](https://github.com/Frontify/fondue/pull/1317/commits/57390d76db9649dcc8510b5bb0d3199dfbc8013e#diff-5630dc06a984553378432b84eaba3c19ee3b30756088044395a5a497e83fa07e)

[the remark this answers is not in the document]
dominiquemb: Marking this as resolved as this issue will be tackled separately.

SamuelAlev (approved):
Good work!
Accessibility wise probably not right, as the slider should now behave like a checkbox, it should use `aria-checked="mixed"` but I wil let Lidija judge.

dominiquemb:
@SamuelAlev I've sent Lidija a message about this. If this is the case, I will create a new branch and a new PR for updating the aria props.

PR #1322, opened 2023-03-20
fix: markdown transformation

bojangles-m:


bojangles-m:
> @bojangles-m frontend tests seem to fail is this a hickup?

It is just a hickup :) yes

manuelkch (approved):
Looks valid to me but i'm not that common with fondue. so up to you if you think it would make sense to let it review from someone additional or it's safe to go anyway ;)

bojangles-m:
> Looks valid to me but i'm not that common with fondue. so up to you if you think it would make sense to let it review from someone additional or it's safe to go anyway ;)

Thnx Manu, nothing special about Fondue here, it is just about transformation.

PR #1324, opened 2023-03-21
fix(RTE): improve url validation

Walkeezy:
Link & button plugins use `isValidUrl` to determine if text contains a url while the user is typing. So we can't use the `URL()` contructor, because it would also recognize `text:` as a valid url.

This causes some unwanted behaviour: https://www.loom.com/share/324b8ed6db23432ab29e2437a108ab35

Ticket: https://app.clickup.com/t/8677pga0r


src/components/RichTextEditor/utils/isValidUrl.ts
```diff
@@ -1,10 +1,3 @@
...
 
-export const isValidUrl = (url: string): boolean => {
-    try {
-        new URL(url);
-        return true;
-    } catch {
-        return false;
-    }
-};
+export const isValidUrl = (url: string): boolean => !!url.match(/^(https?:\/\/|mailto:).+/);
```
celine-s, line 3: maybe add 'tel:' as well. wdyt?

Walkeezy: added 👍🏻

celine-s (approved):
🚀

PR #1335, opened 2023-03-23
fix(RTE): quotes in serializer

Walkeezy:
Replace all double quotes to single quotes in css values, because otherwise it causes issues:

Bildschirm­foto 2023-03-23 um 10 32 58

Ticket: https://app.clickup.com/t/8677k32g3

### DIFF
--- a/src/components/RichTextEditor/Plugins/ListPlugin/ListItemContentMarkupElement.tsx
+++ b/src/components/RichTextEditor/Plugins/ListPlugin/ListItemContentMarkupElement.tsx
@@ -25,7 +25,7 @@ export const getUnderlineClassNames = (designTokens: DesignTokens, element: TEle
 export const getLicElementClassNames = (element: TElement) =>
     merge([
         getColumnBreakClasses(element),
-        element.align ? justifyClassNames[element.align as string] : 'tw-justify-start tw-flex',
+        element.align ? justifyClassNames[element.align as string] : 'tw-justify-start',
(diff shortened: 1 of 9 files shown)


celine-s (commented):
nice! good idea, also with the tests! The list plugin has now issues. 
image

replace UL_CLASSES with `"[&>li>p]:before:tw-content-['•'] [&>li>p]:before:tw-px-2 tw-list-none tw-pl-[10px] tw-mb-[10px] tw-ml-[15px] tw-list-none tw-pl-[10px] tw-mb-[10px] tw-ml-[15px]"`, same with LIST_TYPES and maybe add a test there as well.?

src/components/RichTextEditor/Plugins/ListPlugin/ListItemContentMarkupElement.tsx
```diff
@@ -25,7 +25,7 @@ export const getUnderlineClassNames = (designTokens: DesignTokens, element: TEle
 export const getLicElementClassNames = (element: TElement) =>
     merge([
         getColumnBreakClasses(element),
-        element.align ? justifyClassNames[element.align as string] : 'tw-justify-start tw-flex',
+        element.align ? justifyClassNames[element.align as string] : 'tw-justify-start',
```
Walkeezy, line 28: Removed `tw-flex`, because we have `tw-grid` on the next line anyway and there both don't go together :)

PR #1338, opened 2023-03-23
feat(Box): creating box component

dominiquemb:


src/components/Box/Box.tsx
```diff
@@ -0,0 +1,16 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+import React from 'react';
+
+interface DataProps extends React.HTMLAttributes<HTMLDivElement> {
```
imoutaharik, line 5: We don't use Interface in Fondue. You can set the type like this instead and also DataProps is a misleading name.
```
export type BoxProps = {
    'data-test-id'?: string;
} & React.HTMLAttributes<HTMLDivElement>;
```

src/components/Box/Box.tsx
```diff
@@ -0,0 +1,16 @@
...
+    'data-test-id'?: string;
+}
+
+export const Box = ({ children, ...props }: DataProps) => {
```
imoutaharik, line 9: Here at your own criteria, if you want to simplify your ternary, you can also specify data-test-id in the args.
```
export const Box = ({ children, 'data-test-id': dataTestId, ...props }: BoxProps) => {
    return (
        <div data-test-id={dataTestId ? `fondue-${dataTestId}` : 'fondue-box'} {...props}>
            {children}
        </div>
    );
};
```

src/components/Box/Box.cy.tsx
```diff
@@ -0,0 +1,35 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+import React, { FC } from 'react';
+import { Box } from './Box';
+
+const Component: FC<JSX.IntrinsicElements['div']> = (props) => {
```
imoutaharik, line 6: Is it really necessary to create a new component here as you mount the same in every test and anyway you can pass whatever to the Box component.

imoutaharik (changes requested):

src/components/Box/Box.tsx
fabriciomendonca: I would simplify it even more, like

```
export const Box = ({ children, 'data-test-id': dataTestId = 'fondue-box', ...props }: BoxProps) => {
    return (
        <div data-test-id={dataTestId} {...props}>
            {children}
        </div>
    );
};
```

src/components/Box/Box.tsx
```diff
@@ -0,0 +1,18 @@
...
+export const BOX_TEST_ID = 'fondue-box';
+
+export const Box = ({ children, 'data-test-id': dataTestId = BOX_TEST_ID, ...props }: BoxProps) => {
+    return (
+        <div data-test-id={dataTestId} {...props}>
+            {children}
+        </div>
+    );
+};
+Box.displayName = 'FondueBox';
```
joseazevedo88, line 18: What about just `Box`?

src/components/Box/Box.stories.tsx
```diff
@@ -0,0 +1,36 @@
...
+        'aria-label': {
+            type: 'string',
+        },
+    },
+    args: {
+        className: 'tw-font-bold tw-flex',
+    },
+} as Meta<Props>;
+
+export const Box: StoryFn<Props> = (args: Props) => {
```
joseazevedo88, line 29: I would add a couple more stories with styled boxes, inspiration: https://chakra-ui.com/docs/components/box

joseazevedo88 (commented):
Looking great, super happy with this initiative. Left a couple points for discussion 👌

src/components/Box/Box.stories.tsx
```diff
@@ -0,0 +1,36 @@
...
+        },
+        'data-test-id': {
+            type: 'string',
+        },
+        'aria-label': {
+            type: 'string',
+        },
+    },
+    args: {
+        className: 'tw-font-bold tw-flex',
```
SamuelAlev, line 25: Why do we allow this, Fondue is a design system not just an UI library 🤔

src/components/Box/Box.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+import React, { HTMLAttributes } from 'react';
+
+export type BoxProps = {
+    'data-test-id'?: string;
+} & HTMLAttributes<HTMLDivElement>;
+
+export const BOX_TEST_ID = 'fondue-box';
+
+export const Box = ({ children, 'data-test-id': dataTestId = BOX_TEST_ID, ...props }: BoxProps) => {
```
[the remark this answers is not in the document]
fabriciomendonca: We must provide a default value for `data-test-id` and explicitly define it since `data-*` props are defined by the user, of course other `data` props can be passed based on HTMLAttributes...

src/components/Box/Box.tsx
fabriciomendonca: All Fondue components have the `displayName` with the `Fondue` prefix to make it easier to identify them inside the React devtools.

src/components/Box/Box.stories.tsx
fabriciomendonca: On our feedback sessions we heard a lot from devs that exposing the `className` here were a must for them, also designers agreed that this is necessary since it is part of our Design System Principles.

src/components/Box/Box.tsx
```diff
@@ -0,0 +1,18 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+import React, { HTMLAttributes } from 'react';
+
+export type BoxProps = {
+    'data-test-id'?: string;
+} & HTMLAttributes<HTMLDivElement>;
```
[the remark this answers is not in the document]
fabriciomendonca: Done

joseazevedo88 (approved):

SamuelAlev (approved):

SamuelAlev:
Not happy about the `className` however

PR #1340, opened 2023-03-24
refactor(Accordion)!: delete deprecated props

fabriciomendonca:
In this PR:

- Delete deprecated props
- Use `span` for `AccordionHeaderIcon` to avoid invalid HTML tree `button > div`
- Union type instead of enum for `AccordionHeaderIconSize`
- Get rid of `React.FC` usage
- Delete unnecessary test cases
- Delete unnecessary stories
- Get rid of `FieldsetHeader` usage inside the `Accordion` component and subcomponents
- Add `FieldsetHeader` to storybook `Deprecating` category

### DIFF
--- a/src/components/Accordion/Accordion.stories.tsx
+++ b/src/components/Accordion/Accordion.stories.tsx
@@ -21,7 +21,7 @@ import { IconSize } from '@foundation/Icon/IconSize';
 import { action } from '@storybook/addon-actions';
 import { Meta, StoryFn } from '@storybook/react';
 import { merge } from '@utilities/merge';
-import React, { ReactNode, useState } from 'react';
+import React, { ReactElement, ReactNode, useState } from 'react';
 import { Color } from '../../types';
 import { EXAMPLE_IMAGES } from '../AssetInput/example-assets';
 import { Accordion as AccordionComponent, AccordionItem } from './Accordion';
@@ -123,7 +123,7 @@ export const WithDifferentAccordionItems: StoryFn<AccordionProps> = () => {
     );
 };
 
-const customHeader = ({ isOpen, children }: AccordionHeaderProps) => (
+const customHeader = ({ isOpen, children }: AccordionHeaderProps): ReactElement => (
     <div
         className={merge([
             'tw-px-6 tw-py-4 tw-bg-black-5 tw-flex tw-justify-between tw-font-bold tw-items-center',
@@ -135,11 +135,11 @@ const customHeader = ({ isOpen, children }: AccordionHeaderProps) => (
     </div>
 );
 
-const Code = ({ children }: { children?: ReactNode }) => (
+const Code = ({ children }: { children: ReactNode }): ReactElement => (
     <code className="tw-bg-black-5 tw-rounded tw-px-2 tw-text-box-negative-strong tw-text-s">{children}</code>
 );
 
-const PropsTable = ({ rows }: { rows: [ReactNode, ReactNode, ReactNode][] }) => (
+const PropsTable = ({ rows }: { rows: [ReactNode, ReactNode, ReactNode][] }): ReactElement => (
     <table className="tw-table-fixed tw-border tw-border-black-10 tw-my-4">
         <thead className="tw-bg-black-5">
             <tr className="tw-p-3 tw-border-b tw-border-b-black-10">
--- a/src/components/Accordion/types.ts
+++ b/src/components/Accordion/types.ts
@@ -1,7 +1,7 @@
 /* (c) Copyright Frontify Ltd., all rights reserved. */
 
 import { TreeState } from '@react-stately/tree';
-import { PropsWithChildren, ReactElement, ReactNode } from 'react';
+import { ReactElement, ReactNode } from 'react';
 import { Node } from '@react-types/shared';
 
 export type AccordionProps = {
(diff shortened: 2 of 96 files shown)


src/components/Accordion/AccordionHeader.tsx
```diff
@@ -4,25 +4,25 @@ import React, { ReactElement, cloneElement, isValidElement, useState } from 'rea
...
+
 export const AccordionHeader = ({
     isOpen = false,
     decorator,
     disabled = false,
     children,
-    size = AccordionHeaderIconSize.Medium,
-    bold = true,
-    type,
+    'data-test-id': testId = ACCORDION_HEADER_TEST_ID,
```
imoutaharik, line 17: I think it's better to use the same name everywhere and just name it `dataTestId`

src/components/Accordion/AccordionHeader.tsx
```diff
@@ -4,25 +4,25 @@ import React, { ReactElement, cloneElement, isValidElement, useState } from 'rea
...
 }: AccordionHeaderProps) => {
     const [headingRef, setHeadingRef] = useState<HTMLSpanElement | null>(null);
-    const icon = <AccordionHeaderIcon isOpen={isOpen} disabled={disabled} size={size} type={type} />;
+    const icon = <AccordionHeaderIcon isOpen={isOpen} disabled={disabled} />;
 
     return (
-        <span data-test-id="accordion-header" className="tw-block tw-px-8 tw-py-6">
+        <span data-test-id={testId} className="tw-block tw-px-8 tw-py-6">
```
imoutaharik, line 23: Please provide a default with the fondue prefix.

imoutaharik (changes requested):

src/components/Accordion/Accordion.stories.tsx
```diff
@@ -132,23 +123,23 @@ export const WithDifferentAccordionItems: StoryFn<AccordionProps> = () => {
...
         ])}
     >
         {children}
-        <AccordionHeaderIcon isOpen={isOpen} type={FieldsetHeaderType.Accordion} size={AccordionHeaderIconSize.Small} />
+        <AccordionHeaderIcon isOpen={isOpen} size="small" />
     </div>
 );
 
-const Code: FC = ({ children }) => (
+const Code = ({ children }: PropsWithChildren<unknown>) => (
```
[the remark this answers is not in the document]
fabriciomendonca: You are right, `({ children }: { children?: ReactNode })` will be way better :) Good catch

src/components/Accordion/AccordionHeader.tsx
fabriciomendonca: Perfect. Will change it and we keep it consistent

src/components/Accordion/AccordionHeader.tsx
fabriciomendonca: The default value comes from the `ACCORDION_HEADER_TEST_ID` constant value `fondue-accordion-header` on line 10 :)

src/components/Accordion/types.ts
```diff
@@ -17,43 +17,30 @@ export type AriaAccordionItemProps = {
     header: Omit<AccordionHeaderProps, 'isOpen'> & { active?: boolean; onClick?: () => void };
     padding?: boolean;
     divider?: boolean;
-    headerComponent?: FC<AccordionHeaderProps>;
+    headerComponent?: (props: AccordionHeaderProps) => JSX.Element | null;
```
SamuelAlev, line 20:
```suggestion
    headerComponent?: (props: AccordionHeaderProps) => ReactNode | null;
```
So you can have `ReactElement` inside and even `Nullable<ReactElement>`

src/components/Accordion/types.ts
```diff
@@ -17,43 +17,30 @@ export type AriaAccordionItemProps = {
...
+    'data-test-id'?: string;
 };
 
 export type AccordionItemProps = PropsWithChildren<{
     header: Omit<AccordionHeaderProps, 'isOpen'> & { active?: boolean; onClick?: () => void };
     padding?: boolean;
     divider?: boolean;
-    headerComponent?: FC<AccordionHeaderProps>;
+    headerComponent?: (props: AccordionHeaderProps) => JSX.Element | null;
```
SamuelAlev, line 28: Same

SamuelAlev (changes requested):

src/components/Accordion/types.ts
fabriciomendonca: With JSX.Element too, or am I wrong? I will change it and also bring back the type you added on Clarify for the `Header` component

fabriciomendonca: I was researching about this and `JSX.Element` would be the best solution here since `ReactNode` represents anything that `React` can render, including `string`, `boolean`, `number` and so on... But we must provide a component as the `headerComponent` prop and `React` will type it like this:
```
declare global {
  namespace JSX {
    interface Element extends React.ReactElement<any, any> { }
  }
}
```

src/components/Accordion/types.ts
fabriciomendonca: Same answer

src/components/Accordion/types.ts
fabriciomendonca: @imoutaharik and I were talking about this and we believe the way to go is using ReactElement also to keep things consistent in Fondue since we use it in other places.

SamuelAlev: ReactElement doesn't allow TextNode, where ReactNode does.

SamuelAlev: You can go for it if you feel like 👍

fabriciomendonca: Went for it since it was already used on Jonas PR too

imoutaharik (approved):

SamuelAlev (approved):

PR #1344, opened 2023-03-27
feat(Container): creating container component

dominiquemb:


src/components/Container/Container.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+import React from 'react';
+
+export type PaddingType = 0 | 4 | 8 | 12 | 16 | 20 | 24 | 28 | 32 | 36 | 40;
+export type ContainerProps = {
+    children?: React.ReactNode;
+    minWidth?: number;
+    maxWidth?: number;
+    padding?: PaddingType;
+    dataTestId?: string;
```
imoutaharik, line 11: "data-test-id" instead of camelCase

src/components/Container/Container.stories.tsx
```diff
@@ -0,0 +1,28 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+import React from 'react';
+import { Meta, StoryFn } from '@storybook/react';
+import { Container as ContainerComponent, ContainerProps, PaddingType } from './Container';
+
+const acceptablePaddingInput: PaddingType[] = [0, 4, 8, 12, 16, 20, 24, 28, 32, 36, 40];
+
+export default {
+    title: 'Components/Container',
```
imoutaharik, line 10: Storybook category should be `Layout` and not `Components`

src/components/Container/Container.tsx
```diff
@@ -0,0 +1,30 @@
...
+export const Container = ({ children, dataTestId, minWidth, maxWidth, padding = 0 }: ContainerProps) => {
+    const acceptablePaddingInput: PaddingType[] = [0, 4, 8, 12, 16, 20, 24, 28, 32, 36, 40];
+
+    return (
+        <div
+            data-test-id={dataTestId ? dataTestId : 'fondue-container'}
+            style={{
+                maxWidth,
+                minWidth,
+                padding: acceptablePaddingInput.includes(padding) ? padding : 0,
```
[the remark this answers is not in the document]
dominiquemb: Dynamic classes don't easily work with tailwind unfortunately. In the screenshot is what I tried.
Screenshot 2023-03-28 at 15 16 47

src/components/Container/Container.tsx
```diff
@@ -0,0 +1,45 @@
...
+export type PaddingType = 0 | 4 | 8 | 12 | 16 | 20 | 24 | 28 | 32 | 36 | 40;
+export type ContainerProps = {
+    children?: React.ReactNode;
+    minWidth?: number;
+    maxWidth?: number;
+    padding?: PaddingType;
+    ['data-test-id']?: string;
+};
+
+export const Container = ({ children, minWidth, maxWidth, padding = 0, ...props }: ContainerProps) => {
```
imoutaharik, line 14:
```suggestion
export const Container = ({ children, minWidth, maxWidth, padding = 0, 'data-test-id': dataTestId }: ContainerProps) => {
```

src/components/Container/Container.stories.tsx
```diff
@@ -0,0 +1,29 @@
...
+            options: Object.values(acceptablePaddingInput),
+            control: { type: 'select' },
+        },
+        children: { table: { disable: false } },
+    },
+    args: {
+        minWidth: 50,
+        maxWidth: 100,
+        padding: 12,
+        ['data-test-id']: 'custom-test-id',
```
imoutaharik, line 24: No need for the `[]`

src/components/Container/Container.stories.tsx
```diff
@@ -0,0 +1,29 @@
...
+        children: <div>This is a container</div>,
```
imoutaharik, line 25: Just an idea, in the example, can you even add a background color to the content so it's easier to see the padding changes.

imoutaharik (changes requested):

src/utilities/dimensions.ts
```diff
@@ -0,0 +1,33 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+export type SpacingValues = 0 | 4 | 8 | 12 | 16 | 20 | 24 | 28 | 32 | 36 | 40;
+export type DimensionUnities = '%' | 'px' | 'rem' | 'em' | 'vh' | 'vw';
```
SamuelAlev, line 4: Same

src/utilities/tokens.ts
```diff
@@ -0,0 +1,53 @@
...
+    name: PalleteName;
+    attributes: {
+        'tailwind-name': string;
+    };
+};
+
+export type TokenPalletes = Record<PalleteName | string, TokenPallete[]>;
+
+export const TOKEN_STATES = ['hover', 'pressed', 'inverse', 'inverse-hover', 'inverse-pressed'];
+export const TOKEN_STATES_REGEX = new RegExp(`(${TOKEN_STATES.join('|')})$`, 'i');
```
SamuelAlev, line 18: That looks overkill, what is the intention?

SamuelAlev (changes requested):

src/utilities/tokens.ts
fabriciomendonca: The intention here is to get only the "base" token name from the list of tokens, specially when using box tokens. That base name will be used as a prefix for the states one. The regex will be used to filter those and retrieve a list with only the base ones. 
Example:
```
// The first three ones should be applied to bg
box-neutral - added to the list
box-neutral-hover - ignored
box-neutral-pressed - ignored
// inverse ones should be applied to color
box-neutral-inverse - ignored
box-neutral-inverse-hover - ignored
box-neutral-inverse-pressed - ignored
```

src/utilities/tokens.ts
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    return palettes
+        .filter((palette) => !TOKEN_STATES_REGEX.test(palette.attributes['tailwind-name']))
+        .map((palette) => `${palette.attributes['tailwind-name']}`);
+};
+
+export const BOX_ALIAS_TOKENS_PREFIX = getAliasTokensTailwindNamesByCategory('box', true);
+export const BOX_BG_ALIAS_TOKENS = BOX_ALIAS_TOKENS_PREFIX.map((name) => `tw-bg-${name}`);
+export const BOX_TEXT_ALIAS_TOKENS = BOX_ALIAS_TOKENS_PREFIX.map((name) => `tw-text-${name}-inverse`);
```
SamuelAlev, line 53: Not allowed https://tailwindcss.com/docs/content-configuration#dynamic-class-names

fabriciomendonca: Done the manually created array. It is being used only in the stories for now.

SamuelAlev: >It is being used only in the stories for now.
Move it to the stories otherwise it will be added to the styles.css in build

PR #1346, opened 2023-03-30
chore(typescript): Replace React.FC and React.PropsWithChildren

jonasbuechel:
# Get types ready for React 18
@types/react at version 18 does not have implicit children property anymore so we decided to replace React.FC and Reac.PropsWithChildren with simple type definitions.

## Replacing React.FC
**Before:** 
`const MyComponent: FC<MyComponentProps> = ({ someProperty }) => ()`

**After:** (children only in use where needed)
`const MyComponent = ({ someProperty, children }: MyComponentProps): ReactElement => ()`

## Replacing PropsWithChildren
**Before:**
`export type MyComponentProps = PropsWithChildren<{
    emphasis?: Emphasis;
    size?: Size;
}>;`

**After:**
```
export type MyComponentProps = {
    emphasis?: Emphasis;
    size?: Size;
    children?: ReactNode;
};

```

### DIFF
--- a/scripts/templates/component.ts
+++ b/scripts/templates/component.ts
@@ -8,12 +8,11 @@ export const setup = (componentName: string): ComponentFileBuilderResponse => ({
 
 import React, { ReactNode } from "react";
 
-export type ${componentName}Props = { foo: string; children?: ReactNode; };
+export type ${componentName}Props = { children?: ReactNode; };
 
-export const ${componentName} = ({ foo, children }: ${componentName}Props) => {
+export const ${componentName} = ({ children }: ${componentName}Props) => {
     return (
         <div data-test-id="${toKebabCase(componentName)}">
-            <div>{foo}</div>
             {children}
         </div>
     );
--- a/scripts/templates/component.ts
+++ b/scripts/templates/component.ts
@@ -6,9 +6,9 @@ import { toKebabCase } from '../transforms';
 export const setup = (componentName: string): ComponentFileBuilderResponse => ({
     content: `/* (c) Copyright Frontify Ltd., all rights reserved. */
 
-import React, { PropsWithChildren } from "react";
+import React, { ReactNode } from "react";
 
-export type ${componentName}Props = PropsWithChildren<{ foo: string }>;
+export type ${componentName}Props = { foo: string; children?: ReactNode; };
 
 export const ${componentName} = ({ foo, children }: ${componentName}Props) => {
     return (
--- a/src/components/Accordion/types.ts
+++ b/src/components/Accordion/types.ts
@@ -1,7 +1,7 @@
 /* (c) Copyright Frontify Ltd., all rights reserved. */
 
 import { TreeState } from '@react-stately/tree';
-import { FC, PropsWithChildren, ReactNode } from 'react';
+import { PropsWithChildren, ReactElement, ReactNode } from 'react';
 import { Node } from '@react-types/shared';
 import { FieldsetHeaderSize, FieldsetHeaderType } from '@components/FieldsetHeader';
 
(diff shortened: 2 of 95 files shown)


src/components/Breadcrumbs/Breadcrumbs.cy.tsx
```diff
@@ -20,16 +20,19 @@ const BREADCRUMB_ITEMS_MIXED_ELEMENTS = [
     { label: 'item with link', link: '/some-third-link' },
 ];
 
-const ChangingBreadcrumbs: FC<BreadcrumbsProps> = () => {
-    const [items, setItems] = useState(BREADCRUMB_ITEMS);
+const ChangingBreadcrumbs = ({ items }: BreadcrumbsProps): ReactElement => {
```
SamuelAlev, line 23: Why the change?

SamuelAlev (approved):
Small details comments, feel free to merge

src/components/Accordion/types.ts
```diff
@@ -17,14 +17,14 @@ export type AriaAccordionItemProps = {
...
-    headerComponent?: (props: AccordionHeaderProps) => ReactElement;
+    headerComponent?: (props: AccordionHeaderProps) => ReactElement | null;
 };
 
 export type AccordionItemProps = PropsWithChildren<{
     header: Omit<AccordionHeaderProps, 'isOpen'> & { active?: boolean; onClick?: () => void };
     padding?: boolean;
     divider?: boolean;
-    headerComponent?: (props: AccordionHeaderProps) => ReactElement;
+    headerComponent?: (props: AccordionHeaderProps) => ReactElement | null;
```
SamuelAlev, line 27: Not needed anymore

SamuelAlev (changes requested):

src/components/Accordion/types.ts
```diff
@@ -1,7 +1,7 @@
 /* (c) Copyright Frontify Ltd., all rights reserved. */
 
 import { TreeState } from '@react-stately/tree';
-import { FC, PropsWithChildren, ReactNode } from 'react';
+import { PropsWithChildren, ReactElement, ReactNode } from 'react';
```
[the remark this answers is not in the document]
jonasbuechel: removed then all with https://github.com/Frontify/fondue/commit/55f5fcd1e2ede8d450979426b93b15aa789f4b45

src/components/Breadcrumbs/Breadcrumbs.cy.tsx
jonasbuechel: Needed to correctly type the function and now needed to replace the already used prop `items`, which was in use but also passed BREADCRUM_ITEMS 😸

src/components/Accordion/types.ts
SamuelAlev: You can update the Clarify code to comply, Fondue shouldn't adapt to Clarify (this is important)

fabriciomendonca: This is the same discussion we are having here: https://github.com/Frontify/fondue/pull/1340
I'm using `JSX.Element` which is the default inferred type by Typescript when you create a component without `React.FC`

fabriciomendonca: Changes on Clarify in this PR https://github.com/Frontify/clarify/pull/14480

jonasbuechel: Fair point, requires those changes to be released first
https://github.com/Frontify/clarify/pull/14562

jonasbuechel: removed them 👍

imoutaharik (commented):
Thanks for taking care! 🥇 
I did a smoke test and didn't find anything breaking in Clarify.

Just left some small comments.

scripts/templates/component.ts
```diff
@@ -6,9 +6,9 @@ import { toKebabCase } from '../transforms';
 export const setup = (componentName: string): ComponentFileBuilderResponse => ({
     content: `/* (c) Copyright Frontify Ltd., all rights reserved. */
 
-import React, { PropsWithChildren } from "react";
+import React, { ReactNode } from "react";
 
-export type ${componentName}Props = PropsWithChildren<{ foo: string }>;
+export type ${componentName}Props = { foo: string; children?: ReactNode; };
```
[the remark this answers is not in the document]
jonasbuechel: done 👍

src/components/MultiInput/MultiInput.tsx
```diff
@@ -10,9 +10,10 @@ export enum MultiInputLayout {
 export type MultiInputProps = {
     layout: MultiInputLayout;
     spanLastItem?: boolean;
+    children: ReactNode;
```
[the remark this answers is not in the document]
jonasbuechel: Yes, the component won't have any output without children, right?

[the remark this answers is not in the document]
imoutaharik: Yes, that's right. This component needs children for the logic inside.

imoutaharik (approved):

PR #1351, opened 2023-04-04
fix(DatePicker): single date selection styles for disabled past and future dates

fabriciomendonca:


joewoodson (approved):
Looks good for an interim fix until a solution with new styling for the valid past dates scenario is found 👍🏽

PR #1352, opened 2023-04-04
feat(Grid): creating Grid layout component

dominiquemb:


src/components/Grid/Grid.tsx
```diff
@@ -0,0 +1,76 @@
...
+            data-test-id={dataTestId}
+            className={merge([bg, color])}
+            style={{
+                margin,
+                padding,
+                width,
+            }}
+            as={ContainerElement}
+        >
+            <div data-test-id="fondue-grid-inner-wrapper" className="tw-w-full tw-h-full tw-inline-flex tw-flex-wrap">
```
fabriciomendonca, line 69: The `Grid` component is meant to provide an interface to CSS Grid Layout. We must use `tw-grid` instead of Flex Box.

fabriciomendonca, line 69: Why to have so many nested elements? This will lead to a really complex HTML tree lie `div > div > div > child` but it should be `div > child`

src/components/Grid/Grid.stories.tsx
```diff
@@ -0,0 +1,70 @@
...
+        'positive',
+        'negative',
+        'warning',
+        'neutral',
+    ];
+    return (
+        <Grid {...args}>
+            {contentClasses.map((contentClass, i) => {
+                const key = `content-key-${contentClass}-${i}`;
+                return <div key={key} className={`tw-w-full tw-h-full tw-bg-box-${contentClass}-strong`}></div>;
```
fabriciomendonca, line 66: We can't use dynamic Tailwind class names like `tw-bg-box-${contentClass}-strong`, the `contentClasses` array must have the full class names.

src/components/Grid/Grid.cy.tsx
```diff
@@ -0,0 +1,108 @@
...
+const GRID_ID = '[data-test-id=fondue-grid]';
+const GRID_INNER_WRAPPER_ID = '[data-test-id=fondue-grid-inner-wrapper]';
+const GRID_ITEM_ID = '[data-test-id=fondue-grid-item]';
+const GRID_ITEM_CONTENT_ID = '[data-test-id=fondue-grid-item-content]';
+const CONTENT_STRING = 'Test';
+
+const GridComponent = (props: GridProps): ReactElement => {
+    return (
+        <Grid {...props}>
+            <div>{CONTENT_STRING}</div>
```
fabriciomendonca, line 15: We can use the `Box` component here as well.

fabriciomendonca (changes requested):

src/components/Grid/Grid.stories.tsx
```diff
@@ -0,0 +1,72 @@
...
+        spacingX: 4,
+        spacingY: 4,
+        column: 4,
+        'data-test-id': 'custom-test-id',
+        padding: 4,
+        margin: 4,
+        width: '100%',
+        rowHeight: '100px',
+        maxHeight: undefined,
+        boxColorToken: 'box-neutral',
```
fabriciomendonca, line 42: This must be removed

src/components/Grid/Grid.tsx
```diff
@@ -0,0 +1,93 @@
...
+    'data-test-id': dataTestId = CONTAINER_TEST_ID,
+    width = '100%',
+    rowHeight,
+    bg,
+    color,
+    margin = 0,
+    padding = 0,
+    as: ContainerElement = 'div',
+}: GridProps) => {
+    const spacingYClassName = spacingMap[spacingY] ? `tw-gap-y-${spacingMap[spacingY]}` : `tw-gap-y-${spacingMap[4]}`;
```
fabriciomendonca, line 66: This won't work since Tailwind does not accept dynamic class names. You should add the whole string to the map. Same for padding and margin

src/components/Grid/Grid.tsx
```diff
@@ -0,0 +1,93 @@
...
+    margin?: SpacingValue;
+    bg?: string;
+    color?: string;
+    'data-test-id'?: string;
+    as?: ContainerHTMLElement;
+};
+
+export const CONTAINER_TEST_ID = 'fondue-grid';
+
+const spacingMap = {
```
fabriciomendonca, line 26: You can create a map constant for gap, gap-x and gap-y constants in the `@utilities/dimentions` module, like the ones created for padding and margin.

src/components/Grid/Grid.tsx
```diff
@@ -0,0 +1,93 @@
...
+        childrenLength++;
+        return child;
+    });
+
+    return (
+        <Box
+            data-test-id={dataTestId}
+            className={merge(['tw-grid', bg, color, spacingXClassName, spacingYClassName])}
+            style={{
+                margin,
```
fabriciomendonca, line 80: Applying the `padding` and `margin` directly here will convert them to pixels but we must use the exported maps from the `@utilities/dimensions` module.
Please have a look at the `Container` component. https://github.com/Frontify/fondue/blob/beta/src/components/Container/Container.tsx

fabriciomendonca (changes requested):

fabriciomendonca (changes requested):

src/components/Grid/Grid.stories.tsx
```diff
@@ -0,0 +1,72 @@
...
+        'tw-bg-box-positive-strong',
+        'tw-bg-box-negative-strong',
+        'tw-bg-box-warning-strong',
+        'tw-bg-box-neutral-strong',
+    ];
+    return (
+        <Grid {...args}>
+            {contentClasses.map((contentClass, i) => {
+                const key = `content-key-${contentClass}-${i}`;
+                return <Box key={key} className={merge(['tw-w-full tw-h-full', contentClass])}></Box>;
```
fabriciomendonca, line 68: Please, add some content to the children so we can see the `color` property taking effect

src/components/Grid/Grid.stories.tsx
```diff
@@ -0,0 +1,72 @@
...
+        margin: 4,
+        width: '100%',
+        rowHeight: '100px',
+        maxHeight: undefined,
+        boxColorToken: 'box-neutral',
+    },
+} as Meta<GridProps>;
+
+export const Default: StoryFn<GridProps> = (args: GridProps) => {
+    const contentClasses = [
```
fabriciomendonca, line 47: This can be moved to outside the component

fabriciomendonca (changes requested):

src/components/Grid/Grid.tsx
```diff
@@ -0,0 +1,93 @@
...
+                padding,
+                width,
+                gridTemplateColumns: `repeat(${column}, ${(1 / column) * 100}%)`,
```
fabriciomendonca, line 83: This is not working properly since the grid container is getting bigger than the parent element due to the gap values which should be considered when calculating the column width.
For checking it, go to the Storybook:

- Set padding and margin to 0
- Set column value 1
- You will see how it should fit inside the parent element
- Change the amount of columns
- You will see the grid container getting bigger and bigger because of the added gap

fabriciomendonca (changes requested):

src/components/Grid/Grid.tsx
dominiquemb: Good catch!

fabriciomendonca (changes requested):

src/components/Grid/Grid.tsx
```diff
@@ -0,0 +1,83 @@
...
+    GapXMapping,
+    GapYMapping,
+    MARGIN_VALUES_MAP,
+    PADDING_VALUES_MAP,
+    SPACING_VALUES,
+    SpacingValue,
+} from '@utilities/dimensions';
+import { ContainerHTMLElement } from 'src/types/elements';
+import { merge } from '@utilities/merge';
+import { Box } from '../Box/Box';
```
fabriciomendonca, line 15:
```suggestion
import { Box } from '@components/Box';
```

src/components/Grid/Grid.tsx
```diff
@@ -0,0 +1,83 @@
...
+    rowHeight?: `${number}${DimensionUnity}` | 'auto';
+    padding?: SpacingValue;
+    margin?: SpacingValue;
+    bg?: string;
+    color?: string;
+    'data-test-id'?: string;
+    as?: ContainerHTMLElement;
+};
+
+export const CONTAINER_TEST_ID = 'fondue-grid';
```
fabriciomendonca, line 32: Maybe renaming this to `GRID_TEST_ID`?

src/utilities/dimensions.ts
```diff
@@ -31,3 +31,55 @@ export const MARGIN_VALUES_MAP: Record<SpacingValue, string> = {
     36: 'tw-m-9',
     40: 'tw-m-10',
 };
+
+export const GapXMapping = {
```
fabriciomendonca, line 35:
```suggestion
export const GAP_X_MAP = {
```

src/utilities/dimensions.ts
```diff
@@ -31,3 +31,55 @@ export const MARGIN_VALUES_MAP: Record<SpacingValue, string> = {
...
+    60: 'tw-gap-x-15',
+    64: 'tw-gap-x-16',
+    68: 'tw-gap-x-17',
+    72: 'tw-gap-x-18',
+    76: 'tw-gap-x-19',
+    80: 'tw-gap-x-20',
+    84: 'tw-gap-x-21',
+};
+
+export const GapYMapping = {
```
fabriciomendonca, line 61:
```suggestion
export const GAP_Y_MAP = {
```

fabriciomendonca (changes requested):

src/components/Grid/Grid.stories.tsx
```diff
@@ -0,0 +1,80 @@
...
+            control: { type: 'select' },
+        },
+        padding: {
+            options: SPACING_VALUES,
+            control: { type: 'select' },
+        },
+        margin: {
+            options: SPACING_VALUES,
+            control: { type: 'select' },
+        },
```
fabriciomendonca, line 32: Let's add the `as` and `maxHeight` props to the `argTypes`
```
as: {
    options: ['div', 'span'],
    control: {
        type: 'select',
    },
},
maxHeight: {
    type: 'string',
},
```

fabriciomendonca (changes requested):

fabriciomendonca (approved):

PR #1349, opened 2023-04-03
chore: Upgrade to React 18

imoutaharik:
@jonasbuechel @SamuelAlev Whenever you want to give green light, the PR is ready before I go on Holidays.

PR #1355, opened 2023-04-05
chore: update deps

SamuelAlev:
Update majors of `react`, `react-dom`, `glob` and `storybook`


imoutaharik:
For typecheck issues  (it's fine to use any on the Flyout, we will deprecate it anyway.)
-> https://github.com/Frontify/fondue/pull/1349

fabriciomendonca (approved):
Checked the components and they look the same as the ones in beta.
The popperjs flip was not working properly only in the `Menu` component which is already fixed in this PR.
So it looks good to me.

PR #1359, opened 2023-04-06
fix: update Breadcrumb spacing to match Figma design and refactored CurrentBreadcrumb/BreadcrumbItem

SamCreasey:
Updated spacing for the breadcrumb and removed unnecessary logic separating the types to simplify the component

### DIFF
--- a/src/components/Breadcrumbs/CurrentBreadcrumbItem.tsx
+++ b/src/components/Breadcrumbs/CurrentBreadcrumbItem.tsx
@@ -44,7 +44,7 @@ export const CurrentBreadcrumbItem = ({
 
     return (
         <li
-            className="tw-w-full tw-inline-flex tw-align-middle tw-gap-x-1 tw-text-m tw-text-black dark:tw-text-white tw-items-center"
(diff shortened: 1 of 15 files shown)


src/components/Breadcrumbs/CurrentBreadcrumbItem.tsx
```diff
@@ -32,71 +22,42 @@ export const CurrentBreadcrumbItem = ({
...
-                                bold && 'tw-font-bold',
-                                isFocusVisible && FOCUS_STYLE,
-                            ])}
-                        >
-                            {label}
-                        </span>
-                    </ItemWithBadges>
-                </>
-            )}
+            <Element ref={ref} {...props} href={link} className={classNames}>
```
fabriciomendonca, line 52: Shouldn't we remove the `href` prop here since it is already part of the `{...props}`? Otherwise we're gonna have `<span href="...">` and `<button href="...">`

SamCreasey: Good catch! Missed this prop

PR #1372, opened 2023-04-17
fix(Date Picker): fixed arrow position when dropdown is open

bojangles-m:
If the Reference field is smaller and positioned right in the browser then the Arrow should be also positioned right.
This PR fixes it.

### DIFF
--- a/src/components/DatePicker/DatePicker.tsx
+++ b/src/components/DatePicker/DatePicker.tsx
@@ -167,6 +167,22 @@ export const DatePicker = forwardRef<ReactDatePicker<never, boolean>, DatePicker
                             />
                         </div>
                     )}
+                    popperModifiers={[
+                        {
+                            name: 'arrow',
+                            options: {
+                                padding: ({ popper, reference, placement }) => {
+                                    const side = placement.includes('end') ? 'left' : 'right';
+                                    const padding = {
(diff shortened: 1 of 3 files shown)


src/components/DatePicker/DatePicker.tsx
```diff
@@ -167,6 +167,22 @@ export const DatePicker = forwardRef<ReactDatePicker<never, boolean>, DatePicker
...
                     )}
+                    popperModifiers={[
+                        {
+                            name: 'arrow',
+                            options: {
+                                padding: ({ popper, reference, placement }) => {
+                                    const side = placement.includes('end') ? 'left' : 'right';
+                                    const padding = {
+                                        [`${side}`]:
+                                            Math.max(popper.width, Math.min(popper.width, reference.width)) - 40,
```
fabriciomendonca, line 178: I would extract this magic number `40` into a constant so we can keep track of why it is there. WDYT?

fabriciomendonca (changes requested):

src/components/DatePicker/DatePicker.tsx
bojangles-m: Updated

fabriciomendonca (approved):
lg2m

PR #1373, opened 2023-04-17
chore(Stack)!: deprecating old Stack component

dominiquemb:


src/components/Card/Card.stories.tsx
```diff
@@ -5,7 +5,7 @@ import { Meta, StoryFn } from '@storybook/react';
 import { Card, CardProps } from './Card';
 import { SegmentedControls } from '@components/SegmentedControls/SegmentedControls';
 import { Divider } from '@components/Divider/Divider';
-import { Stack } from '@layout/Stack';
+import { LegacyStack } from '../../layout/LegacyStack';
```
fabriciomendonca, line 8: We should use the path alias here for consistency `@layout/LegacyStack`

src/layout/LegacyStack/LegacyStack.tsx
```diff
@@ -50,25 +50,25 @@ const justifyMap: Record<StackJustify, string> = {
...
     direction = 'row',
     wrap = false,
     justify = 'start',
     align = 'stretch',
-}: StackProps): ReactElement => {
+}: LegacyStackProps): ReactElement => {
     return (
         <div
-            data-test-id="stack"
+            data-test-id="legacy-stack"
```
fabriciomendonca, line 71: We should use the `fondue-` prefix here and also add `data-test-id` as a prop like we are doing on other components now.

fabriciomendonca (changes requested):

fabriciomendonca (commented):
It looks good. Will wait the Clarify PR (https://github.com/Frontify/clarify/pull/14930) to be approved so I can approve this one.

fabriciomendonca (approved):

PR #1377, opened 2023-04-20
chore: update `semantic-release` and `storybook`

SamuelAlev:
Add missing packages
Update `semantic-release` and `storybook`


package.json
```diff
@@ -116,8 +120,8 @@
         "react-dom": "18.2.0",
         "rimraf": "4.4.1",
         "rollup": "3.20.2",
-        "semantic-release": "19.0.5",
-        "storybook": "7.0.4",
+        "semantic-release": "^21.0.1",
```
fabriciomendonca, line 123: Maybe we don't need this since we are using `pnpx` for running `semantic-release`

PR #1381, opened 2023-04-24
feat(icons): Generate icons and updates to svg structures

mikeyrayvon:
Generates IconCamera and IconFolderArrowRight


src/foundation/Icon/Generated/IconBell12.tsx
```diff
@@ -11,21 +11,12 @@ function IconBell12(props: GeneratedIconProps): React.ReactElement<GeneratedIcon
             name="IconBell12"
             {...props}
         >
-            <g fill="none">
```
imoutaharik, line 14: Mmmm, Why a lot of size 12 icons got updated? That `<g fill="none"` may cause an issue in Clarify with a possibility of the icons going blank.

mikeyrayvon: No clue! I just pulled the branch and ran the generate icons command...

src/foundation/Icon/Generated/IconBell12.tsx
```diff
@@ -11,21 +11,12 @@ function IconBell12(props: GeneratedIconProps): React.ReactElement<GeneratedIcon
...
-                <g clipPath="url(#prefix__a)">
-                    <path
-                        fill="currentColor"
```
mikeyrayvon, line 17: But it looks like the fill on the path was (and will still be) `currentColor`. At least in this case

src/foundation/Icon/Generated/IconBell12.tsx
```diff
@@ -11,21 +11,12 @@ function IconBell12(props: GeneratedIconProps): React.ReactElement<GeneratedIcon
...
-                    />
-                </g>
-                <defs>
-                    <clipPath id="prefix__a">
-                        <path fill="#fff" d="M0 0h12v12H0z" />
-                    </clipPath>
-                </defs>
-            </g>
+            <path
+                fill="currentColor"
```
mikeyrayvon, line 15: See here

PR #1386, opened 2023-04-26
fix: aria-level must be >= 1

anxobotana:
Current `aria-level` values start from `0` when those values must be always `>= 1`

_The value for aria-level is an integer greater than or equal to 1._


anxobotana:
closing because this can/must be solve from clarify

PR #1391, opened 2023-04-27
fix: tree item dragged text color

anxobotana:


oliverschwendener:
To me it looks like this is not fixed yet, see my screen recording:

https://user-images.githubusercontent.com/15727229/234848845-711dcc1d-1451-496e-93be-381b931acbe4.mov

oliverschwendener (approved):

PR #1392, opened 2023-04-27
fix(RTE): Incomplete string escaping or encoding

imoutaharik:
image

Let me know if you think that may cause issues.


src/components/RichTextEditor/serializer/utils/serializeLeafToHtml.ts
```diff
@@ -12,7 +12,7 @@ import escapeHtml from 'escape-html';
 
 export const serializeLeafToHtml = (node: TText): string => {
     let string = escapeHtml(node.text);
-    string = string.replace('\n', '<br />');
+    string = string.replace(/\n/g, '<br />');
```
SamuelAlev, line 15:
```suggestion
    string = string.replaceAll(/\n/g, '<br />');
```
So ESLint doesn't complains after merging https://github.com/Frontify/fondue/pull/1394

triggertoo (approved):
lgtm 👍

PR #1396, opened 2023-04-28
feat(Divider)!: add vertical prop to Divider

dominiquemb:


src/components/Divider/Divider.tsx
```diff
@@ -2,6 +2,8 @@
 
 import React, { ReactElement } from 'react';
 
+export type DividerVertical = true | false;
```
imoutaharik, line 5: Do we really need that type?

src/components/Divider/Divider.tsx
```diff
@@ -19,6 +21,7 @@ export type DividerProps = {
     style?: DividerStyle;
     height?: DividerHeight | string;
     color?: string;
+    vertical?: DividerVertical;
```
imoutaharik, line 24: That could just be `boolean`

src/components/Divider/Divider.tsx
```diff
@@ -29,16 +32,37 @@ const styleMap = {
...
+    return (
+        <>
+            {vertical && (
+                <div
+                    className="tw-flex tw-self-stretch tw-mt-0 tw-mb-0 tw-items-center tw-justify-center"
+                    style={{ width: height }}
+                    data-test-id="divider"
+                >
+                    <div
+                        className={`tw-w-px ${verticalClassNames}`}
```
imoutaharik, line 50: `tw-w-px` is already present in `verticalClassNames` so you can just pass the variable -> `className={verticalClassNames}`

imoutaharik (changes requested):

src/components/Divider/Divider.stories.tsx
```diff
@@ -35,7 +36,30 @@ export default {
...
+            },
+        },
     },
 } as Meta;
 
-export const Divider: StoryFn<DividerProps> = (args: DividerProps) => <DividerComponent {...args} />;
+export const Divider: StoryFn<DividerProps> = (args: DividerProps) => {
+    if (args.vertical) {
+        return (
+            <div className="tw-flex tw-flex-row">
```
fabriciomendonca, line 50: We could use `Flex` component in the stories, WDYT?

src/components/Divider/Divider.tsx
```diff
@@ -29,16 +32,37 @@ const styleMap = {
...
-        />
-    </div>
-);
+    color = '#CCC',
+}: DividerProps): ReactElement => {
+    const verticalClassNames = `tw-w-px tw-h-full tw-border-r tw-m-0 ${styleMap[style]}`;
+    return (
+        <>
+            {vertical && (
+                <div
```
fabriciomendonca, line 44: I believe we're gonna need to use `aria-hidden="true"` on both cases, since `Divider` is a `div` with no content. It would be nice to confirm that with @LidijaStoilova before merging this...

src/components/Divider/Divider.stories.tsx
```diff
@@ -35,7 +37,30 @@ export default {
...
+        vertical: {
+            control: {
+                type: 'boolean',
+            },
+        },
     },
 } as Meta;
 
-export const Divider: StoryFn<DividerProps> = (args: DividerProps) => <DividerComponent {...args} />;
+export const Divider: StoryFn<DividerProps> = (args: DividerProps) => {
```
fabriciomendonca, line 48: There is a problem when you first load the story page: the `vertical` prop is false but it is rendering the divider in the vertical position... Idk if this a problem on the story or the component itself.
https://deploy-preview-1396--beta-fondue-components.netlify.app/?path=/story/components-divider--divider

imoutaharik (changes requested):
There is a problem with the story when you first load it. Divider is shown as vertical even if it's false in the controls.

src/components/Divider/Divider.cy.tsx
```diff
@@ -44,24 +44,44 @@ describe('Divider component', () => {
...
+
+    it('allows the divider to be vertical', () => {
+        cy.mount(<Divider vertical={true} />);
+
+        cy.get(`div${DIVIDER_SELECTOR}`).should('exist');
+        cy.get(DIVIDER_LINE_SELECTOR).should('exist').and('have.class', 'tw-border-r');
+    });
+
+    it('should render with custom data-test-id', () => {
+        cy.mount(<Divider data-test-id="custom-container-test-id" />);
```
fabriciomendonca, line 82: Maybe we could change the custom id to `custom-divider-test-id`

src/components/Divider/Divider.stories.tsx
```diff
@@ -35,7 +38,30 @@ export default {
...
```
fabriciomendonca, line 49: It would be nice to have a story with the vertical divider between elements with different amount of content to see its behaviour on this type of use case. WDYT?

PR #1398, opened 2023-05-01
feat(new): DialogFooter and DialogHeader

imoutaharik:


src/components/DialogFooter/DialogFooter.stories.tsx
```diff
@@ -0,0 +1,40 @@
...
+    },
+} as Meta<DialogFooterProps>;
+
+export const Default: StoryFn<DialogFooterProps> = () => {
+    return (
+        <DialogFooter
+            buttons={[
+                {
+                    children: 'Cancel',
+                    onClick: () => console.log('Cancel'),
```
fabriciomendonca, line 26: We could use Storybook actions for the callbacks, WDYT?

src/components/DialogFooter/DialogFooter.stories.tsx
```diff
@@ -0,0 +1,40 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+import React from 'react';
+import { Meta, StoryFn } from '@storybook/react';
+import { DialogFooter, DialogFooterProps } from './DialogFooter';
+import { ButtonEmphasis, ButtonStyle } from '@components/Button';
+import { IconCheckMark } from '@foundation/Icon';
```
fabriciomendonca, line 7: Shouldn't we import from `@foundation/Icon/Generated`? https://github.com/Frontify/fondue/search?q=%40foundation%2FIcon%2FGenerated

src/components/DialogFooter/DialogFooter.tsx
```diff
@@ -0,0 +1,27 @@
...
+    buttons: ButtonProps[];
+    'data-test-id'?: string;
+};
+
+export const DialogFooter = ({
+    buttons,
+    'data-test-id': dataTestId = 'fondue-dialog-footer',
+}: DialogFooterProps): ReactElement => {
+    return (
+        <div
```
fabriciomendonca, line 16: We could use `Box` here, WDYT?

src/components/DialogHeader/DialogHeader.cy.tsx
```diff
@@ -0,0 +1,76 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+import React from 'react';
+import { DialogHeader, DialogHeaderSize } from './DialogHeader';
+import { IconIcon } from '@foundation/Icon';
```
fabriciomendonca, line 5: Same thing about importing from `/Generated`...

src/components/DialogHeader/DialogHeader.tsx
```diff
@@ -0,0 +1,70 @@
...
+    title: string;
+    size?: DialogHeaderSize;
+    collapseBottom?: boolean;
+    onClose?: () => void;
+    icon?: ReactElement;
+    badge?: ReactElement;
+    'data-test-id'?: string;
+};
+
+export enum DialogHeaderSize {
```
fabriciomendonca, line 18: We were changing these enums to literal union types in other components... Maybe we should change here too?

src/components/DialogHeader/DialogHeader.tsx
```diff
@@ -0,0 +1,70 @@
...
+    return (
+        <div
+            data-test-id={dataTestId}
+            className={merge([
+                'tw-px-6 tw-flex tw-items-center tw-border-b tw-border-b-line',
+                size === DialogHeaderSize.Large ? 'tw-py-6' : 'tw-py-4',
+                collapseBottom && 'tw-pb-0 tw-border-none',
+            ])}
+        >
+            <div className="tw-flex tw-items-center" data-test-id={`${dataTestId}-title`}>
```
fabriciomendonca, line 41: Maybe using the `Flex` component?

src/components/DialogHeader/DialogHeader.tsx
```diff
@@ -0,0 +1,70 @@
...
+                {icon && (
+                    <span data-test-id={`${dataTestId}-icon`} className="tw-mr-1.5">
+                        {icon}
+                    </span>
+                )}
+                <Text weight="x-strong" size={size === DialogHeaderSize.Large ? 'large' : 'medium'}>
+                    {title}
+                </Text>
+                {badge && (
+                    <span data-test-id={`${dataTestId}-badge`} className="tw-ml-1.5">
```
fabriciomendonca, line 51: `Badge` component is rendering a `div` by default. Using it inside a `span` will lead to invalid markup. Should we change this here?

src/components/DialogHeader/DialogHeader.tsx
```diff
@@ -0,0 +1,70 @@
...
+import { IconCross } from '@foundation/Icon/Generated';
+import { Text } from '@typography/Text';
+
+export type DialogHeaderProps = {
+    title: string;
+    size?: DialogHeaderSize;
+    collapseBottom?: boolean;
+    onClose?: () => void;
+    icon?: ReactElement;
+    badge?: ReactElement;
```
fabriciomendonca, line 14: Maybe we can set badge and icon props to be `BadgeProps`, `IconProps`. Being a `ReactElement` means we can pass something different than a `Badge` and `Icon` for them...

src/components/DialogHeader/DialogHeader.tsx
imoutaharik: Changed to div

src/components/DialogHeader/DialogHeader.tsx
imoutaharik: Good point, I set a type to the ReactElement. Like that we don't have Icon: ReactElement and Badge: Badgeprops. 
A react Element needs to be passed for both properties. It's more consistent.

src/components/DialogHeader/DialogHeader.tsx
imoutaharik: I discussed that with Jonas, Tomorrow I'm challenging it with the Fronti-FE community.

src/components/DialogHeader/DialogHeader.stories.tsx
```diff
@@ -0,0 +1,41 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+import React from 'react';
+import { Meta, StoryFn } from '@storybook/react';
+import { DialogHeader as DialogHeaderComponent, DialogHeaderProps, DialogHeaderSize } from './DialogHeader';
+import { Badge, BadgeStyle } from '@components/Badge';
+import { IconIcon } from '@foundation/Icon';
```
fabriciomendonca, line 7: Shouldn't this be also imported from `/Generated`?

fabriciomendonca (approved):

PR #1400, opened 2023-05-02
feat(Stack): creating new Stack component

dominiquemb:


src/components/Stack/Stack.stories.tsx
```diff
@@ -0,0 +1,158 @@
...
+
+const Code = ({ children }: { children: ReactNode }): ReactElement => (
+    <code className="tw-bg-box-neutral tw-rounded tw-px-2 tw-text-box-neutral-inverse tw-text-s">{children}</code>
+);
+
+export const Stack: StoryFn<STACK_PROPS> = (args: STACK_PROPS) => {
+    const isVertical = args.direction === 'row' || args.direction === 'row-reverse';
+
+    return (
+        <StackComponent {...args} divider={<Divider vertical={isVertical} />}>
```
imoutaharik, line 88: The Divider is enforced while it's an optional property. I would create different stories:
- Stack (No Divider)
- Stack with horizontal Divider
- Stack with vertical Divider

imoutaharik (changes requested):

src/components/Stack/Stack.stories.tsx
```diff
@@ -0,0 +1,174 @@
...
+            This story shows only Box Alias Tokens in controls, but you can pass any color token class to define bg and
+            color properties.
+        </p>
+    </StackComponent>
+);
+StackWithBoxAliasToken.args = {
+    bg: 'tw-bg-box-neutral',
+};
+
+export const StackAsSpanElement: StoryFn<STACK_PROPS> = (args: STACK_PROPS) => (
```
imoutaharik, line 137: That story doesn't seem to be very useful. As there is only one item within the Stack component, property like alignment or justify for example seems useless.

imoutaharik (changes requested):

imoutaharik (changes requested):

src/components/Stack/Stack.stories.tsx
dominiquemb: @imoutaharik I think I forgot to push the changes. The span story is gone now.

imoutaharik (commented):
Thanks for making the changes! 🎉  This looks good to me.
But there is just one tiny detail before we can merge... don't forget to export the components in the `index.ts` file at the root of `components` folder. 😉

imoutaharik (changes requested):
Last comment is still not addressed.

imoutaharik (approved):

PR #1402, opened 2023-05-03
feat: add `parentType` in the drop callback

anxobotana:
Adds the possibility to get the `parentType` of the zone we drop in

### DIFF
--- a/src/components/Tree/TreeItem/TreeItem.tsx
+++ b/src/components/Tree/TreeItem/TreeItem.tsx
@@ -107,8 +107,8 @@ export const TreeItem = memo(
                     return;
                 }
 
-                if (isActive && over && canDrop) {
-                    onDrop?.({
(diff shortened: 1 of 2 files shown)


src/components/Tree/TreeItem/TreeItem.tsx
```diff
@@ -108,7 +108,12 @@ export const TreeItem = memo(
                 }
 
                 if (isActive && over && canDrop) {
-                    onDrop?.({ id: active.id, parentId: projection.parentId, sort: projection.position });
+                    onDrop?.({
```
fabriciomendonca, line 111: Why not to add `onDrop` to the `if`? Like 
```typescript
if (isActive && over && canDrop && onDrop) {
  onDrop(...);
}
```

SamuelAlev (approved):

PR #1405, opened 2023-05-04
fix: move/delete/append actions on tree 

SamuelAlev:


src/components/Tree/types.ts
```diff
@@ -90,7 +90,7 @@ export type TreeState = {
     expandedIds: Set<string>;
     selectionMode: 'single' | 'multiselect';
     overlay?: Overlay;
-    nodes: ReactElement[];
+    nodes: ReactElement<InternalTreeItemProps>[];
```
SamuelAlev, line 93: Let's keep it like that and fix the type issues by checking the data :)

src/components/Tree/Tree.tsx
```diff
@@ -129,16 +139,13 @@ const reducer = produce((draft: TreeState, action: TreeStateAction) => {
                     return;
                 }
 
+                const offset = draft.nodes.findLastIndex((node) => node.props.parentId === id) - sliceIndex;
```
anxobotana, line 142: Note that when the item is deleted `node.props.parentId` still contains the old data, so I think this won't exclude it from the list.
That's why I added the filter part:
```
((node.props.parentId === id && newNodeChildrenIds.includes(node.props.id)) || node.props.parentId !== id)
```

anxobotana: nvm... I see what you did... it might work 😄

src/components/Tree/Tree.tsx
```diff
@@ -122,28 +121,24 @@ const reducer = produce((draft: TreeState, action: TreeStateAction) => {
...
+                if (parentNodeIndex === -1) {
+                    console.error(`Element with ID "${parentId}" not found.`);
                     return;
                 }
 
-                const currentChildrenIds = getNodeChildrenIds(draft.nodes, id);
+                const currentChildrenIds = getNodeChildrenIds(draft.nodes, parentId);
 
-                const newChildrenIds = children.map((node) => node.props.id as string);
+                const newChildrenIds = children.map((node) => node.props.id);
```
GoranRibic, line 135: Fyi this will have type any[] if we remove 'as string'.

GoranRibic (approved):

src/components/Tree/Tree.tsx
SamuelAlev: Not anymore
Screenshot 2023-05-08 at 09 19 47

PR #1406, opened 2023-05-04
fix: add `aria-label` on drag handle

anxobotana:
Fixes a couple of accessibility issues:
- the aria-description and language (`<div>`) were added into the `<ul>`, they are added outside the `<ul>` now
- the drag button has no aria-label, it will have now a `aria-labeledby` controlled from clarify

https://app.clickup.com/t/8676zm491

### DIFF
--- a/src/components/Tree/TreeItem/DragHandle.tsx
+++ b/src/components/Tree/TreeItem/DragHandle.tsx
@@ -15,7 +15,7 @@ export const DragHandle = forwardRef(
     ({ active, className, ...props }: DragHandleProps, ref: React.ForwardedRef<HTMLButtonElement>) => {
         return (
             <button
-                aria-labelledby="DraggableLabel"
+                aria-label="Draggable item"
                 {...props}
                 ref={ref}
                 className={merge([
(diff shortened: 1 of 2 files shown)


src/components/Tree/TreeItem/DragHandle.tsx
```diff
@@ -15,6 +15,7 @@ export const DragHandle = forwardRef(
     ({ active, className, ...props }: DragHandleProps, ref: React.ForwardedRef<HTMLButtonElement>) => {
         return (
             <button
+                aria-labelledby="DraggableLabel"
```
SamuelAlev, line 18: We don't set this ID anywhere in Fondue, why not using static `aria-label`? (translations is a topic for the Fondue team and out-of-scope)

anxobotana: exactly because of that, there is no translations yet in fondue. The label can be controlled from clarify depending on the context, language and needs

SamuelAlev: We should rely on:
```
onDragStart({ active }) {
    return `Picked up ${getActiveTitle(active) || active.id}.`;
},
```

Translations for later as none of the components have translations yet.

anxobotana: one thing is the aria-description and a different is the aria-label. Accessibility tests complains about the missing label... I can add a static label `Draggable Item` and leave with that for now

anxobotana: done

anxobotana: closing clarify PR https://github.com/Frontify/clarify/pull/15090 since it's not needed anymore

SamuelAlev (approved):

PR #1407, opened 2023-05-05
fix(Rich Text Editor): updated type export

bojangles-m:
Added test because the method was failing and it was failing because of incorrectly imported types.

### DIFF
--- a/src/components/RichTextEditor/serializer/serializeToHtml.spec.ts
+++ b/src/components/RichTextEditor/serializer/serializeToHtml.spec.ts
@@ -34,6 +34,21 @@ describe('serializeNodesToHtml()', () => {
         expect(serialized).to.contain('column-gap:normal');
     });
 
+    it('should serialize the html text', () => {
+        const node = [
+            {
+                type: 'p',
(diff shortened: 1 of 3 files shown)


src/components/RichTextEditor/serializer/serializeToHtml.spec.ts
```diff
@@ -34,6 +34,21 @@ describe('serializeNodesToHtml()', () => {
...
+                type: 'p',
+                children: [
+                    {
+                        text: "<img src=x onerror='alert(123)'/>\\n",
+                    },
+                ],
+            },
+        ];
+        const serialized = serializeNodesToHtml(node);
+        expect(serialized).to.be.a('string');
```
fabriciomendonca: Wouldn't it be good to also check if the returned string is as expected, instead of only the type of it? Like on the `serializes an empty line to a linebreak` test case?

fabriciomendonca (approved):
Just left a suggestion for the test `should serialize the html text` but it is not mandatory. lg2m

PR #1412, opened 2023-05-08
fix(Tree): grab handler visibility

anxobotana:
On tree, the grab handler should be only visible on hover, even when item is active.
Currently it's always visible if the item is active.

https://app.clickup.com/t/8677xj8ev


imoutaharik (commented):
I still see the handler when the item is active:
image

anxobotana:
> I still see the handler when the item is active: image

@imoutaharik 
The issue is because the focus stays on the item. Clicking anywhere else will remove the focus from the item and only on :hover the item will be visible. We cannot remove the focus-within pseudo-class for accessibility, removing it the dragger won't be visible browsing with keyboard.

imoutaharik:
> > I still see the handler when the item is active: image
> 
> @imoutaharik The issue is because the focus stays on the item. Clicking anywhere else will remove the focus from the item and only on :hover the item will be visible. We cannot remove the focus-within pseudo-class for accessibility, removing it the dragger won't be visible browsing with keyboard.

I see, I understand the technical constraint here. Make sure The designer is aware of it because the purpose of the task is not really fulfilled. It would be nice to know if Design should be actually updated in this case.

anxobotana:
> I see, I understand the technical constraint here. Make sure The designer is aware of it because the purpose of the task is not really fulfilled. It would be nice to know if Design should be actually updated in this case. If items are draggable, should the handler always be visible after all. I found it strange to have it visible on active item and not on other draggable items while looking the same than non-draggable (unless you hover) - it's confusing. WDYT?
> 

@imoutaharik agreed. IMHO, probably having the handler always visible when the item is draggable looks more cohesive and coherent. But It's not my call 😅

anxobotana:
Due the limitation/conflict :focus/:hover Product Design has decided to leave it as is for now. 
So i will close the PR for now

PR #1414, opened 2023-05-09
refactor(Textarea): new textarea component

fabriciomendonca:


src/components/Textarea/Textarea.tsx
```diff
@@ -29,92 +24,11 @@ export type TextareaProps = {
...
-                    ])}
-                    data-test-id="error-state-exclamation-mark-icon"
-                >
-                    <IconExclamationMarkTriangle />
-                </span>
-            )}
-        </div>
-    );
+export const Textarea = ({ ...props }: TextareaProps): ReactElement => {
+    return <TextareaField {...props} />;
```
fabriciomendonca, line 32: The new `Textarea` input has new props like `label`, `helpText`, `errorText`. So we are going to split it into `TextareaField` (exported and with stories, this is the current `Textarea` component) and `TextareaInput` (internal, for avoiding breaking changes). The new `Textarea` will verify if the `label` property is provided (required for rendering the `TextareaInput`) and decide which of them should be rendered.

src/components/TextareaField/TextareaField.cy.tsx
```diff
@@ -0,0 +1,132 @@
...
+import React from 'react';
+import { TextareaField } from './TextareaField';
+
+const DEFAULT_TEXT = 'A new text area';
+const PLACEHOLDER = 'This is the placeholder';
+const DECORATOR_TEXT = 'And also a decorator';
+const DECORATOR = <span>{DECORATOR_TEXT}</span>;
+const INPUT_TEXT = 'some text';
+const ROW_HEIGHT = 16;
+const TEXTAREA_ID = '[data-test-id=textarea]';
```
fabriciomendonca, line 13: I was thinking, should we expose and change the test id? This will lead to breaking changes but (hopefully) only on test files in `Clarify`/`web-app` project since devs shouldn't rely on `data-test-id` for the component features... WDYT?

fabriciomendonca: BTW, I exported the `data-test-id` prop but kept the default value `textarea` to avoid breaking changes

src/internal/inputs/InputValidation.cy.tsx
```diff
@@ -0,0 +1,42 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+import React from 'react';
+import { InputValidation } from './InputValidation';
+import { Validation } from '@utilities/validation';
+
+const INPUT_VALIDATION_LOADING_ID = '[data-test-id="fondue-input-valdiation-loading-state"]';
```
imoutaharik, line 7: There is a small typo here in the data-test-id

PR #1415, opened 2023-05-10
fix: active tree item colors on dragging

anxobotana:
Regression:
bugs #1391 and  #1395 are back after TreeItem refactoring.

When an active tree items is dragged or an inactive item is dragged and goes over an active item, colors are off (see description from original PRs).

### DIFF
--- a/src/components/Tree/TreeItem/TreeItem.tsx
+++ b/src/components/Tree/TreeItem/TreeItem.tsx
@@ -198,11 +198,12 @@ export const TreeItem = memo(
                     FOCUS_VISIBLE_STYLE,
                     !isActive && !isSelected && 'focus-within:tw-bg-box-neutral',
                     'tw-outline-none tw-ring-inset tw-group tw-px-2.5 tw-no-underline tw-leading-5 tw-h-10',
-                    isSelected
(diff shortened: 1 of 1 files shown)


src/components/Tree/TreeItem/TreeItem.tsx
```diff
@@ -198,11 +198,12 @@ export const TreeItem = memo(
                     FOCUS_VISIBLE_STYLE,
                     !isActive && !isSelected && 'focus-within:tw-bg-box-neutral',
                     'tw-outline-none tw-ring-inset tw-group tw-px-2.5 tw-no-underline tw-leading-5 tw-h-10',
-                    isSelected
+                    isSelected && !transform?.y
```
GoranRibic, line 201: why use transform.y? Can't you use the isActive flag?

anxobotana: No, because we want the active item to have `tw-text`text` color (not white) when an not active item is dragged over displacing the active one.

Otherwhise you'll have this: 
Screenshot 2023-04-28 at 14 34 17

instead of:
Screenshot 2023-04-28 at 14 40 22

PR #1416, opened 2023-05-10
feat(Tree): cursor change can/cannot drop item

anxobotana:
Accessibility improvement: Cursor changes when an TreeItem can or cannot be dropped in tree

### DIFF
--- a/src/components/Tree/TreeItem/TreeItem.tsx
+++ b/src/components/Tree/TreeItem/TreeItem.tsx
@@ -1,6 +1,6 @@
 /* (c) Copyright Frontify Ltd., all rights reserved. */
 
-import React, { Children, MouseEvent, memo, useCallback, useEffect, useMemo } from 'react';
+import React, { Children, MouseEvent, memo, useCallback, useMemo } from 'react';
 import { AnimateLayoutChanges, useSortable } from '@dnd-kit/sortable';
 import { useDndContext, useDndMonitor } from '@dnd-kit/core';
 import { CSS } from '@dnd-kit/utilities';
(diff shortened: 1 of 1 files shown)


src/components/Tree/TreeItem/TreeItem.tsx
```diff
@@ -91,6 +91,12 @@ export const TreeItem = memo(
                 (typeof over?.data?.current?.accepts === 'string' &&
                     over.data.current.accepts?.split(', ').includes(active.data.current.type)));
 
+        useEffect(() => {
```
GoranRibic, line 94: This is not the place to fix this. You have onDragStart handler where you can set cursor to grabbing

anxobotana: it is not about dragging, it's about `no-drag`, when the item is over a not allowed place to be dropped.


https://github.com/Frontify/fondue/assets/357211/5466329a-0833-42eb-bebd-f90a15e3fee3

GoranRibic: Ah okay,  still no need for duplicating logic, you have the onDragOver method on useDndMonitor.

anxobotana: Ok, i moved it to the `onDragMove` method. `onDragOver` cannot be used for this because it ONLY tracks when the draggable moves from one item position to another and not when the draggable gets moved at the different levels.

see video (using `onDragOver`):

https://github.com/Frontify/fondue/assets/357211/33ee5e23-3c1f-4603-8f50-dcb5cbf6bfbc

GoranRibic (approved):

PR #1420, opened 2023-05-15
Add hugWidth prop and add padding for each item

joseazevedo88:
image

### DIFF
--- a/src/components/SegmentedControls/SegmentedControls.tsx
+++ b/src/components/SegmentedControls/SegmentedControls.tsx
@@ -31,6 +31,7 @@ export type SegmentedControlsProps = {
     onChange: (id: string) => void;
     ariaLabel?: string;
     disabled?: boolean;
+    hugWidth?: boolean;
 };
 
 const isIconItem = (item: TextOrNumberItem | IconItem): item is IconItem => (item as IconItem).icon !== undefined;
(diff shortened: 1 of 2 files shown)


src/components/SegmentedControls/SegmentedControls.tsx
```diff
@@ -131,13 +133,17 @@ export const SegmentedControls = ({
...
     const selectedIndex = items.findIndex((item) => item.id === radioGroupState.selectedValue);
+    const width = hugWidth ? '' : 'tw-w-full';
 
     return (
         <div className="tw-flex">
             <fieldset
                 {...radioGroupProps}
                 data-test-id="fondue-segmented-controls"
-                className="tw-relative tw-h-9 tw-w-full tw-grid tw-grid-flow-col tw-auto-cols-fr tw-justify-evenly tw-p-0 tw-border tw-border-black-20 tw-m-0 tw-bg-black-0 tw-rounded tw-font-sans tw-text-s tw-select-none"
+                className={merge([
```
imoutaharik, line 143: Thanks for handling, I think something more can be done here as it doesn't hug enough. We still have some extra space. It should hug `All` a bit more.
Design:
image
When Inspecting:
Screenshot 2023-05-15 at 16 54 03
We need to get rid of that purple space generated by flexbox when hugwidth is true.

src/components/SegmentedControls/Slider.stories.tsx
```diff
@@ -32,8 +32,18 @@ Text.args = {
...
-export const Number = SegmentedControlsTemplate.bind({});
-Number.args = {
+export const TextWithHugWidth = SegmentedControlsTemplate.bind({});
+TextWithHugWidth.args = {
+    items: [
+        { id: 'a', value: 'All' },
+        { id: 'b', value: 'Top 10' },
+        { id: 'c', value: 'Bottom 10' },
+    ],
+    hugWidth: true,
```
imoutaharik, line 42: Screenshot 2023-05-15 at 16 55 48
Would be nice to add that property to argTypes so it doesn't appear as a textInput but a boolean toggle in Storybook controls.

imoutaharik (changes requested):
Thanks for taking care so quickly!! I just left a few comments and as all properties need to be tested. I must say that a cypress test is also missing. 😃

imoutaharik (approved):
Thanks! It looks much better now and the animation look good too. 😃

PR #1421, opened 2023-05-16
feat: add new emphasis and size to match new design

mnoleto:
https://www.figma.com/file/sNs31nygQIXSfM55RbvS8B/Bulk-Select-Assets?type=design&node-id=875-236497&t=fHKW9xCmM5Knat0Q-4

- Add a checkbox size to be able to have a checkbox of 32px width and height
- Add an emphasis to change the checkbox style (border width, border color) and remove the border when is checked.

### DIFF
--- a/src/components/Checkbox/Checkbox.tsx
+++ b/src/components/Checkbox/Checkbox.tsx
@@ -5,6 +5,7 @@ import { InputLabel, InputLabelTooltipProps } from '@components/InputLabel/Input
 import React, { ForwardRefRenderFunction, HTMLAttributes, ReactNode, forwardRef, useEffect, useState } from 'react';
 
 import { FOCUS_STYLE } from '@utilities/focusStyle';
+import { IconSize } from '@foundation/Icon';
 import { merge } from '@utilities/merge';
 import { mergeProps } from '@react-aria/utils';
 import { useCheckbox } from '@react-aria/checkbox';
@@ -121,7 +122,9 @@ const CheckboxComponent: ForwardRefRenderFunction<HTMLInputElement, CheckboxProp
     };
 
     const stateMap: Record<CheckboxState, ReactNode> = {
-        [CheckboxState.Checked]: <IconCheckMark />,
+        [CheckboxState.Checked]: (
+            <IconCheckMark size={size === CheckboxSize.XLarge ? IconSize.Size24 : IconSize.Size16} />
+        ),
         [CheckboxState.Mixed]: <IconMinus />,
         [CheckboxState.Unchecked]: null,
     };
--- a/src/components/Checkbox/Checkbox.tsx
+++ b/src/components/Checkbox/Checkbox.tsx
@@ -2,8 +2,7 @@
 
 import { IconCheckMark, IconMinus } from '@foundation/Icon/Generated';
 import { InputLabel, InputLabelTooltipProps } from '@components/InputLabel/InputLabel';
-import React, { ForwardRefRenderFunction, HTMLAttributes, forwardRef } from 'react';
-import { useEffect, useState } from 'react';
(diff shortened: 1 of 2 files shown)


src/components/Checkbox/Checkbox.tsx
```diff
@@ -112,6 +115,21 @@ const CheckboxComponent: ForwardRefRenderFunction<HTMLInputElement, CheckboxProp
         inputRef,
     );
 
+    const getCheckboxSize = () => {
```
imoutaharik, line 118: Here instead of creating a function with a switch you could use a variable with Record ->
```
const sizeClassesMap: Record<CheckboxSize, string> = {
    [CheckboxSize.Default]: 'tw-h-4 tw-w-4',
    [CheckboxSize.Large]: 'tw-h-5 tw-w-5',
    [CheckboxSize.XLarge]: 'tw-h-8 tw-w-8',
}
```

src/components/Checkbox/Checkbox.tsx
```diff
@@ -140,8 +158,8 @@ const CheckboxComponent: ForwardRefRenderFunction<HTMLInputElement, CheckboxProp
                                 data-test-id="checkbox-icon-box"
                                 aria-hidden="true"
                                 className={merge([
-                                    'tw-leading-3  tw-p-2 tw-relative tw-flex tw-items-center tw-justify-center tw-rounded tw-border tw-shrink-0 tw-flex-none',
-                                    size === 'Default' ? 'tw-h-4 tw-w-4' : 'tw-h-5 tw-w-5',
+                                    'tw-leading-3 tw-p-2 tw-relative tw-flex tw-items-center tw-justify-center tw-rounded tw-shrink-0 tw-flex-none',
+                                    getCheckboxSize(),
```
imoutaharik, line 162: and here `sizeClassesMap[size]`

src/components/Checkbox/Checkbox.tsx
```diff
@@ -151,14 +169,22 @@ const CheckboxComponent: ForwardRefRenderFunction<HTMLInputElement, CheckboxProp
                                                   'tw-bg-box-disabled tw-text-box-disabled-inverse tw-border-line-strong',
                                           ])
                                         : merge([
+                                              !isCheckedOrMixed(state) && 'tw-bg-base hover:tw-bg-box-neutral',
```
imoutaharik, line 172: The cognitive complexity have increased here as well. It would be nice to simplify now that you added one more `emphasis` option.

imoutaharik (changes requested):

imoutaharik:
@mnoleto  Also I think the Checkmark icon should be bigger on `XLarge`
How it is now:
Screenshot 2023-05-16 at 11 44 41
In Figma:
Screenshot 2023-05-16 at 11 45 52

src/components/Checkbox/Checkbox.tsx
```diff
@@ -105,23 +108,62 @@ const CheckboxComponent: ForwardRefRenderFunction<HTMLInputElement, CheckboxProp
...
             isIndeterminate: state === CheckboxState.Mixed,
-            'aria-label': ariaLabel || label,
+            'aria-label': ariaLabel,
             value,
         },
         toggleState,
         inputRef,
     );
 
+    const emphasisCheckedClassesMap: Record<CheckboxEmphasis, string> = {
```
imoutaharik, line 118: All those tailwind classes mapping can be moved outside of the component so we can keep it as clean as possible.

imoutaharik (changes requested):
Thanks for addressing so quickly, I just left one last comment and we're almost good to ship.

imoutaharik (approved):
🙌🏼  Thanks

PR #1425, opened 2023-05-17
refactor(MultiSelect): remove Weak emphasis

fabriciomendonca:
This PR removes the Weak emphasis from the `MultiSelect` introduced by this PR https://github.com/Frontify/fondue/pull/1077/files

I checked the `web-app` and there was no usage of this emphasis. On the other hand, the changes to the `Trigger` and `Dropdown` components are being used in other contexts. So there will be no changes to those components in order to avoid breaking changes in the `web-app`

### DIFF
--- a/src/components/MultiSelect/MultiSelect.stories.tsx
+++ b/src/components/MultiSelect/MultiSelect.stories.tsx
@@ -5,6 +5,7 @@ import React, { useState } from 'react';
 import { FormControl } from '@components/FormControl';
 import { Validation } from '@utilities/validation';
 import { MultiSelect as MultiSelectComponent, MultiSelectProps, MultiSelectSize, MultiSelectType } from './MultiSelect';
+import { TriggerEmphasis } from '@components/Trigger';
 
 export default {
     title: 'Components/Multi Select',
@@ -42,6 +43,10 @@ export default {
             options: Object.keys(MultiSelectType),
             control: { type: 'select' },
         },
+        emphasis: {
+            options: Object.keys(TriggerEmphasis),
+            control: { type: 'select' },
+        },
         size: {
             options: Object.keys(MultiSelectSize),
             control: { type: 'select' },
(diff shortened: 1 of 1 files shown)


src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -265,26 +220,14 @@ export const MultiSelect = ({
...
-                            />
-                        )}
                     </div>
                 </div>
             </Trigger>
 
             {open &&
                 heightIsReady &&
-                emphasis === MultiSelectEmphasis.Default &&
                 createPortal(
```
imoutaharik, line 231: We can use the new `Portal` component to wrap this div.

imoutaharik (commented):
Thanks, Regarding the `Trigger` and `Dropdown` the weak emphasis should stay because it's needed in some cases. The MultiSelect should also have the weak emphasis available on the Trigger as per Design:
Screenshot 2023-05-17 at 12 42 12
What doesn't make sense and needs to be removed is the change of Design within the overlay when changing emphasis (Change needed only on the MultiSelect) -> (toggling checkboxes vs Avatar)

src/components/MultiSelect/MultiSelect.tsx
fabriciomendonca: Done

src/components/MultiSelect/MultiSelect.stories.tsx
```diff
@@ -50,10 +42,6 @@ export default {
             options: Object.keys(MultiSelectType),
             control: { type: 'select' },
         },
-        emphasis: {
```
imoutaharik, line 53: This should stay and change the Trigger Style. We should be able to control that in Storybook. Also it would be nice to add a Story for weak Trigger.

imoutaharik (changes requested):

PR #1427, opened 2023-05-17
feat(new)!: DialogBody and  InlineDialog

imoutaharik:


src/utilities/dialogs/Overlay.tsx
```diff
@@ -0,0 +1,79 @@
...
+    return (
+        <Popper open={open} placement={placement} offset={offset} flip={flip} enablePortal={enablePortal}>
+            {Children.map(children, (child) => {
+                if (isValidElement(child) && typeof child.type === 'function') {
+                    const { name } = child.type;
+
+                    if (name === Trigger.name) {
+                        return (
+                            <Popper.Reference>
+                                <div className="tw-w-fit" id={id}>
```
SamCreasey, line 48: is this necessary? I think this div could cause issues in the future that we've had in the past where children components can no longer determine their width based on the parent component

src/components/InlineDialog/InlineDialog.tsx
```diff
@@ -0,0 +1,48 @@
...
+import { BaseDialogProps, Modality, OverlayProps } from '../../types/dialog';
+import { PopperPlacement } from '@components/Popper/types';
+import { Trigger } from '@utilities/dialogs/Trigger';
+import { Content } from '@utilities/dialogs/Content';
+import { Overlay } from '@utilities/dialogs/Overlay';
+
+export type InlineDialogProps = OverlayProps & BaseDialogProps;
+
+Trigger.displayName = 'FondueInlineDialogTrigger';
+Content.displayName = 'FondueInlineDialogContent';
```
SamCreasey, line 13: These should be removed from this file

SamCreasey (changes requested):

src/components/DialogHeader/DialogHeader.tsx
```diff
@@ -1,28 +1,12 @@
 /* (c) Copyright Frontify Ltd., all rights reserved. */
```
SamCreasey, line 2: DialogueHeader close button should either be a Fondue component or have correct focus styling

src/components/Popper/Popper.stories.tsx
```diff
@@ -4,29 +4,20 @@ import React, { useState } from 'react';
...
+import { POPPER_STORY_ARGS, PopperPlacement, PopperProps } from '@components/Popper/types';
 import { Box } from '@components/Box';
 
 export default {
     title: 'Utilities/Popper',
     component: Popper,
     tags: ['autodocs'],
     args: {
-        placement: 'bottom-start',
+        placement: PopperPlacement.BottomStart,
```
SamuelAlev, line 15: As Popper component isn't used that much (yet), I would go union type route instead of enum

src/components/Popper/Popper.tsx
```diff
@@ -43,11 +42,12 @@ export const Popper = ({
     open,
     placement = PopperPlacement.BottomStart,
     offset = [0, 8],
-    flip = false,
+    flip = true,
     enablePortal = false,
```
SamuelAlev, line 46: Should be true by default

src/components/Popper/Popper.tsx
```diff
@@ -73,35 +73,27 @@ export const Popper = ({
...
+                        return <div ref={setReferenceElement}>{child}</div>;
                     }
 
                     if (name === Content.name) {
                         return enablePortal ? (
                             <Portal>
                                 <div
                                     ref={setPopperElement}
-                                    style={popperInstance.styles.popper}
+                                    style={{ zIndex: 9999, ...popperInstance.styles.popper }}
```
SamuelAlev, line 84: Quite a bad for scaling, please create variables for the different cases you have.
Examples:

Best read: [Chakra UI](https://chakra-ui.com/community/recipes/z-index)
Alternatives:
- [MUI](https://mui.com/material-ui/customization/z-index/)
- [Bootstrap](https://getbootstrap.com/docs/5.0/layout/z-index/)
- [CoreUI](https://coreui.io/docs/layout/z-index/)

src/components/Popover/Popover.tsx
```diff
@@ -28,46 +17,21 @@ export const Popover = ({
     flip = false,
     enablePortal = false,
```
SamuelAlev, line 18: Should be true by default

src/hooks/useFocusTrap.ts
```diff
@@ -0,0 +1,77 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
```
SamuelAlev, line 1: Please add some tests

src/utilities/dialogs/Overlay.tsx
```diff
@@ -0,0 +1,116 @@
...
+    );
+
+    const maxContentHeight = autoHeight ? maxAutoHeight : maxHeight;
+    const handleClosingInteraction = () => {
+        if (open && modality !== Modality.BlockingModal) {
+            if (handleClose) {
+                handleClose();
+            }
+        }
+    };
```
SamuelAlev, line 49: Add useCallback

src/components/DialogBody/DialogBody.tsx
```diff
@@ -0,0 +1,18 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+import React, { ReactElement } from 'react';
+import { DialogBodyProps } from '../../types/dialog';
+
+export const DialogBody = ({
```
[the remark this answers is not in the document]
imoutaharik: Yes, Reusability in other components. It's totally optional to use those in any dialogs. They are available as independent components as they may be in use anywhere else you want to use the Header (for example) with same styling without the close button. It's pretty flexible that way.
Fo example, it's not recommended to use Flyouts in small devices and present the content in a drawer instead for responsive design. So it's more composable.

src/components/Popover/Popover.tsx
imoutaharik: I set it to false by default as the use of React.Portal creates an accessibility issue (violation in Level A). So, Portal should be enabled only when the content really needs to pop out, if it's within a container with overflow hidden or above a canvas for example. Decision consciously made by the dev.

src/hooks/useFocusTrap.ts
imoutaharik: I added the tests for focusTrap directly in InlineDialog.cy.tsx The whole section `Modality` test it.
Seems like none of the hooks have an independent test file 🤔 , we test them in use in the component.
You feel like it's not enough?

src/components/Popper/Popper.tsx
imoutaharik: Refactored! Thanks for pointing out. :)

src/components/Popper/Popper.tsx
imoutaharik: Replied on second comment. :P

src/components/InlineDialog/InlineDialog.stories.tsx
```diff
@@ -0,0 +1,287 @@
...
+                        icon={<IconDotsVertical16 />}
+                        onClick={() => setIsOpen(!isOpen)}
+                    ></Button>
+                </InlineDialog.Trigger>
+                <InlineDialog.Content>
+                    <DialogBody>
+                        <Box className="tw-p-4">
+                            <Dropdown
+                                onChange={(id) => console.log(id)}
+                                activeItemId={'1'}
```
SamCreasey, line 99:
```suggestion
                                activeItemId='1'
```

src/utilities/dialogs/Content.tsx
```diff
@@ -0,0 +1,7 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+import React, { ReactElement, ReactNode } from 'react';
+
+// eslint-disable-next-line react/jsx-no-useless-fragment
```
SamCreasey, line 5: This is different sytax to the Trigger but they do the same thing. Maybe just use the same as the Trigger and remove the es-lint rule

SamCreasey (changes requested):

src/utilities/dialogs/Overlay.tsx
```diff
@@ -0,0 +1,120 @@
...
+    const maxContentHeight = autoHeight ? maxAutoHeight : maxHeight;
+    const handleClosingInteraction = useCallback(() => {
+        if (open && modality !== Modality.BlockingModal) {
+            if (handleClose) {
+                handleClose();
+            }
+        }
+    }, [handleClose, modality, open]);
+
+    useFocusTrap(ref.current, open, modality === Modality.NonModal);
```
SamCreasey, line 53: In the popover story with interactive content the focus is not contained within the flyout. Is this intentional? It seems like a case when you would have the focus captured.

SamCreasey (changes requested):

src/utilities/dialogs/Content.tsx
imoutaharik: It's not quite the same. in the Trigger, children is one ReactElement but Content receive a ReactNode. You would get an error. I don't want people to wrap their content in a container to use the component so I wrap it in a Fragment. I already checked that and disabling the es-lint rule here is the cleanest and best solution to keep the ease of use.

src/components/InlineDialog/InlineDialog.stories.tsx
imoutaharik: Good catch, I didn't even noticed.

src/utilities/dialogs/Overlay.tsx
imoutaharik: Intentional, for now.

src/components/Popover/Popover.tsx
SamuelAlev: This is not true, many accessible libraries are doing it so layering is easier (and zIndex issues are gone).

src/hooks/useFocusTrap.ts
SamuelAlev: `src/hooks/useFocusTrap.ts`, the hook isn't component specific.

SamCreasey:
Screenshot 2023-06-20 at 17 36 24
When you open an Inline dialogue the tooltip is immediately visible

src/components/InlineDialog/InlineDialog.stories.tsx
```diff
@@ -0,0 +1,302 @@
...
+                modality={args.modality}
+                placement={args.placement}
+                flip={args.flip}
+                offset={args.offset}
+                enablePortal={args.enablePortal}
+                darkUnderlay={args.darkUnderlay}
+                autoHeight={args.autoHeight}
+            >
+                <InlineDialog.Trigger>
+                    <Button
```
SamCreasey, line 90: Focus is not restored to the trigger when the dialogue is closed with escape

src/components/InlineDialog/InlineDialog.stories.tsx
```diff
@@ -0,0 +1,302 @@
...
+                                        content="Just some Information"
+                                    />
+                                    <p>Information</p>
+                                </Flex>
+                            </Box>
+                            <Dropdown
+                                onChange={(id) => console.log(id)}
+                                activeItemId="1"
+                                menuBlocks={[
+                                    {
```
SamCreasey, line 116: All the menu items are disabled or show an incorrect pointer. Not sure if this is an issue with this implementation or if it's on the component itself?

SamCreasey: It seems that the pointer is based on the content underneath the dropdown menu somehow

SamCreasey:
Screenshot 2023-06-20 at 17 47 12
With `placement=top` the `flip` argument doesn't seem to do anything

imoutaharik:
> Screenshot 2023-06-20 at 17 47 12 With `placement=top` the `flip` argument doesn't seem to do anything

It works for me. any specific story or combination of props? I just tried the flip in every story and couldn't reproduce.

src/components/Popover/Popover.tsx
imoutaharik: And yet we got issues reported linked to the use of Portal. I can't find in any docs a cause of it, maybe the focus sequence was not right... topic I will tackle here properly with time. I can set default to true but will also add a comment in the docs to disable it if any a11y issue pops out. 👍🏼

src/hooks/useFocusTrap.ts
imoutaharik: Will add.

src/components/InlineDialog/InlineDialog.stories.tsx
imoutaharik: It's an issue when nested overlay components are using Portals apparently. I removed it from dropdown as it seems to not be necessary. But this next quarter, I'm going to heavily test interactions of overlays within overlays and the dropdown will also be replaced by the Select. So it's "To be continued..." topic.

SamCreasey (approved):

PR #1431, opened 2023-05-22
feat(segmentedControls): remove px in simple item

celine-s:
px was added [here](https://github.com/Frontify/fondue/pull/1420/files#:~:text=w%2Dfull%20tw%2D-,px,-%2D4%20tw%2Dpy)

issue when the segment is small:
image

now:
image

### DIFF
--- a/src/components/SegmentedControls/SegmentedControls.tsx
+++ b/src/components/SegmentedControls/SegmentedControls.tsx
@@ -95,7 +95,7 @@ const SegmentedControlsItem = forwardRef<HTMLDivElement, SegmentedControlsItemPr
                 onClick={handleMockLabelClick}
                 data-test-id={getSegmentedControlsItemTestId()}
                 className={merge([
-                    'tw-relative tw-w-full tw-px-4 tw-py-2 tw-inline-flex tw-justify-center tw-items-center tw-font-sans tw-font-normal tw-h-full tw-text-center',
+                    'tw-relative tw-w-full tw-py-2 tw-inline-flex tw-justify-center tw-items-center tw-font-sans tw-font-normal tw-h-full tw-text-center',
                     isActive && !disabled ? 'tw-text-text' : 'tw-text-text-weak',
                     !disabled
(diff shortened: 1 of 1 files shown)


joseazevedo88 (changes requested):
The horizontal padding is needed in order to prevent the element to hit the border. Here you can try to add `flex-shrink: 0` to the child element. Let me know if it works, otherwise we need another solution

celine-s:
> The horizontal padding is needed in order to prevent the element to hit the border. Here you can try to add `flex-shrink: 0` to the child element. Let me know if it works, otherwise we need another solution

Ahhh sure, makes sense!

Shrink-0 doesn't work, as the size is controlled from the outside and the segmentControl Item should be smaller. See here on [marketplace](https://product.frontify.com/marketplace/explore). How about a compromise sth like `hugWidth ? tw-px-4 : tw-px-2`?

joseazevedo88:
It's not a padding issue, as you can see in this screenshot, there's hardcoded width on the parent container. Just removing that will allow the container to grow. Also hardcoded widths are very rarely a good idea 😄 

image

image

joseazevedo88:
The designer can't do that, I'll explain why: `SegmentedControls` has a design with 16px of horizontal padding on each side. In this case the element inside is an icon with 16px. That gives us 16*3*3 = 144px + borders. So the designer can't expect to have a component and constraint the width like that. If you want I can explain to the designer why it's not correct by principle to hardcode widths and even ore so in this case. Also please refer to the design of the component here: https://www.figma.com/file/WfEbqZI0nnPWHN43MCtvnm/Fondue---Inputs?type=design&node-id=1420-90714&t=TcMf8FJ0tLIqraNe-4

joseazevedo88:
In my opinion, the only correct solution here would be to design a small variant of the `SegmentedControls` variant with less padding, if the designer think the current padding is to big for that specific usage

celine-s:
Good point. Thanks for clarification.

PR #1436, opened 2023-05-23
fix: tooltip paddings

anxobotana:
fixing paddings accordingly to figma design


imoutaharik:
@anxobotana The new Tooltip Figma Design doesn't have the action buttons anymore but it's still in use.
Changing the padding makes the buttons too close to the edge. It would be nice to add that extra padding on that element.
Screenshot 2023-05-23 at 14 43 28

src/components/Tooltip/Tooltip.tsx
```diff
@@ -57,13 +57,13 @@ export type TooltipProps = {
  * This is a temporary workaround because for some yet unknown reasons `tailwindcss` in clarify purges the `tw-pb-3.5` and `tw-pt-3.5` class.
```
SamuelAlev, line 57: Can be removed

anxobotana:
> @anxobotana The new Tooltip Figma Design doesn't have the action buttons anymore but it's still in use. Changing the padding makes the buttons too close to the edge. It would be nice to add that extra padding on that element. Screenshot 2023-05-23 at 14 43 28

@imoutaharik in fact the padding size was applied wrong, it has been fixed now, for the buttons and other cases applying the larger padding when these extra elements are present

Screenshot 2023-05-23 at 15 26 59

SamuelAlev (approved):
👍

imoutaharik (approved):

PR #1440, opened 2023-05-24
feat(SegmentedControls): add padding variant

imoutaharik:


celine-s (approved):
nice, thank you for tackling this!

src/components/SegmentedControls/SegmentedControls.cy.tsx
```diff
@@ -116,4 +118,16 @@ describe('SegmentedControls Component', () => {
             });
         });
     });
+
+    it('renders with a default padding', () => {
+        cy.mount(<Component items={TEXT_ITEMS} padding="default" />);
```
celine-s, line 123: You could event remove padding here, as default doesn't need to be passed.

src/components/SegmentedControls/SegmentedControls.tsx
```diff
@@ -95,7 +99,8 @@ const SegmentedControlsItem = forwardRef<HTMLDivElement, SegmentedControlsItemPr
                 onClick={handleMockLabelClick}
                 data-test-id={getSegmentedControlsItemTestId()}
                 className={merge([
-                    'tw-relative tw-w-full tw-px-4 tw-py-2 tw-inline-flex tw-justify-center tw-items-center tw-font-sans tw-font-normal tw-h-full tw-text-center',
+                    'tw-relative tw-w-full tw-py-2 tw-inline-flex tw-justify-center tw-items-center tw-font-sans tw-font-normal tw-h-full tw-text-center',
+                    padding === 'default' ? 'tw-px-4' : 'tw-px-2',
```
celine-s, line 103:
```suggestion
                    padding === 'small' ? 'tw-px-2' : 'tw-px-4',
```

celine-s: this way we wouldn't need the padding to be defined at all

src/components/SegmentedControls/SegmentedControls.tsx
```diff
@@ -133,6 +138,7 @@ export const SegmentedControls = ({
     ariaLabel = 'SegmentedControls',
     disabled = false,
     hugWidth = false,
+    padding = 'default',
```
joseazevedo88, line 141: I would use the concept of `variant`, instead of `padding`. That way, if we need to add more props to the small variant we can do so with a single prop. Would also do `small | medium` for sizes

imoutaharik: I agree with the name of the sizes. Not so sure about changing the name to `variant`, I get your point to generalize in case there will be future changes but those changes are not planned at all and even that property wasn't really supposed to be. I'm just thinking that having variant=small would create more expectations then just to have the padding changed. while consuming I would expect more to happen.

joseazevedo88: I see your point. But: 

>  those changes are not planned at all

> even that property wasn't really supposed to be.

Do you see my point? We will never know what will and won't be needed, that's why I suggested a variant. Also, I don't think it's a matter of expectations, there should be a design for the different variants and it's up to the designers to use them correctly in figma and it's up to the devs to consume them correctly. 

Having said that,I don't want to block this, so feel free to merge 😄 👌

joseazevedo88 (approved):

PR #1441, opened 2023-05-25
chore: plate update [#867842pht]

bojangles-m:


imoutaharik (commented):
thanks for taking care of this @bojangles-m ! 🎉 
It now works as in Beta.

I just tested functionality from Storybook and compared against beta for now.

CheckboxList finally keeps the TextStyle.
I'll add a task to discuss the expected behaviour of those plugins as orderedList and BulletList clear the formatting but that's for future improvements.
I see you fixed the HotKey issue with the columnBreak plugin but that plugin still causes issues in some stories -> readonly, and toolbar Top. We get a typeError.
`Cannot read properties of null (reading 'breakAfterColumn')`

bojangles-m:
> thanks for taking care of this @bojangles-m ! 🎉 It now works as in Beta.
> 
> I just tested functionality from Storybook and compared against beta for now.
> 
> CheckboxList finally keeps the TextStyle. I'll add a task to discuss the expected behaviour of those plugins as orderedList and BulletList clear the formatting but that's for future improvements. I see you fixed the HotKey issue with the columnBreak plugin but that plugin still causes issues in some stories -> readonly, and toolbar Top. We get a typeError. `Cannot read properties of null (reading 'breakAfterColumn')`

I am sorry but I dont understand the issue you are pointing, could you be more specific how you get to the error or add Loom vide which clearly shows what is the problem, please?

imoutaharik (approved):

PR #1450, opened 2023-06-01
feat(RichTextEditor): improve serializer

ragi96:
- improved code style `serializeNodeToHtmlRecursive`
- pass plugin instead of styles (styles are already within plugins, and if you have special plugins you need to pass them also to `parseRawValue` to get the real nodes)
- also fixes the edge case in which the list could be unstyled


src/components/RichTextEditor/Plugins/ListPlugin/ListItemMarkupElement.tsx
```diff
@@ -24,8 +25,9 @@ export class ListItemMarkupElement extends MarkupElement {
 }
 
 export const getLiStyles = (element: TElement, styles: Record<string, CSSProperties>): CSSProperties => {
+    console.log(getDeepestTextStyle(element));
```
Walkeezy, line 28:
```suggestion
```

Walkeezy (approved):

PR #1451, opened 2023-06-01
fix(Tag): adding tests for sizes, adding tokens, safari icon misalignment

dominiquemb:


src/components/Tag/Tag.tsx
```diff
@@ -63,7 +63,7 @@ export const Tag = ({ type, label, onClick, size = TagSize.Medium, 'data-test-id
         <button
             data-test-id={dataTestId}
             className={merge([
-                'tw-inline-flex tw-items-center tw-border tw-border-solid tw-rounded-full tw-text-xs tw-transition-colors tw-group  tw-break-word',
+                'tw-inline-flex tw-items-center tw-rounded-full tw-text-xs tw-transition-colors tw-group  tw-break-word',
                 size === TagSize.Small ? 'tw-px-[6px] tw-py-[2px]' : 'tw-px-2.5 tw-py-1',
```
imoutaharik, line 67: I know it was not added by you but it caught my attention when I saw your test. 
it would be nice if you could modify that little part and  change those value as `tw-px-1.5 tw-py-0.5` instead of `tw-px-[6px] tw-py-[2px]`

src/components/Tag/Tag.tsx
```diff
@@ -63,7 +63,7 @@ export const Tag = ({ type, label, onClick, size = TagSize.Medium, 'data-test-id
         <button
             data-test-id={dataTestId}
             className={merge([
-                'tw-inline-flex tw-items-center tw-border tw-border-solid tw-rounded-full tw-text-xs tw-transition-colors tw-group  tw-break-word',
+                'tw-inline-flex tw-items-center tw-rounded-full tw-text-xs tw-transition-colors tw-group  tw-break-word',
```
imoutaharik, line 66: When you click on the tag, the blue outline appears. We should disable it adding that class as well `tw-outline-none`

dominiquemb: Added

src/components/Tag/Tag.tsx
dominiquemb: Fixed

PR #1456, opened 2023-06-06
fix(Card): accessibility, tests, adding tokens

dominiquemb:


src/components/Card/Card.tsx
```diff
@@ -30,16 +31,45 @@ export const Card = ({
 
     const clickableProps = !!onClick ? mergeProps(buttonProps, focusProps) : {};
 
+    React.useEffect(() => {
```
imoutaharik, line 34: This seems a bit over-engineered, I believe we can simplify by using the useButton hook on line 26.
Try this:
On line 26:
```
const { buttonProps } = useButton(
        {
            elementType: 'div',
            onPress: (event) => onClick && onClick(event),
            onPressChange: (isPressed) => {
                setKeyDownTrue(isPressed);
            },
        },
        ref,
    );

```

And remove all useEffect for keyUp and KeyDown.

src/components/Card/Card.tsx
```diff
@@ -30,16 +31,45 @@ export const Card = ({
...
         <div
             {...clickableProps}
             data-test-id={dataTestId}
             ref={ref}
             className={merge([
-                'tw-w-full tw-outline-none tw-bg-white tw-border tw-border-black-10 tw-rounded tw-overflow-hidden',
-                hoverable ? 'hover:tw-border-black' : '',
-                onClick ? 'tw-cursor-pointer' : 'tw-cursor-default',
-                isFocusVisible && FOCUS_STYLE,
+                'tw-w-full tw-outline-none tw-bg-white tw-border tw-border-line tw-rounded tw-overflow-hidden',
```
imoutaharik, line 66: when the border width changes, it affect the content of the div and the text seems to move. IT may be worth it to use outline instead of border.
Also the Hover style doesn't appear with onClick so you need to add it `hoverable || onClick` and also on you ternary condition for onClick -> I believe we don't need the active style if there is no onClick.
Try like this:
```
className={merge([
                'tw-w-full tw-bg-white tw-outline tw-outline-1 tw-outline-line tw-outline-offset--1 tw-rounded tw-overflow-hidden',
                hoverable || onClick
                    ? 'tw-cursor-pointer hover:tw-outline-line-xx-strong active:tw-outline-line-xx-strong active:tw-outline-2 active:tw-outline-offset--2'
                    : 'tw-cursor-default',
                isFocusVisible && FOCUS_STYLE,
                keyDownTrue && onClick ? 'tw-outline-line-xx-strong tw-outline-2 tw-outline-offset--2' : '',
            ])}
```

imoutaharik (changes requested):

src/components/Card/Card.tsx
dominiquemb: If I apply these changes, the onClick styles stop appearing completely.  Let's talk about this on the call tomorrow.

src/components/Card/Card.tsx
dominiquemb: Yep this works. Updated.

src/components/Card/Card.tsx
dominiquemb: The updates we discussed have been applied & cypress tests updated.

imoutaharik (approved):

PR #1457, opened 2023-06-06
fix: stabilize rte debounce

SamCreasey:
The debounce function was being re-rendered every time which meant that the timeout was never cleared if the component rerendered, causing a memory leak and triggering the callback unintentionally


src/hooks/useDebounce.ts
```diff
@@ -0,0 +1,16 @@
...
+import { useCallback, useRef } from 'react';
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export const useDebounce = <T extends (...args: any[]) => void>(func: T, timeout = 10) => {
+    const timer = useRef<number | undefined>();
+
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const debouncedFunc = useCallback((...args: any[]) => {
+        clearTimeout(timer.current);
+        timer.current = window.setTimeout(() => func.apply(this, args), timeout);
```
fulopdaniel, line 12: `window` is omittable I think

src/utilities/index.ts
```diff
@@ -10,7 +10,6 @@ export * from './focusStyle';
 export * from './generateRandomId';
 export * from './merge';
 export * from './number';
-export * from './useDebounce';
```
fulopdaniel, line 13: can utilities/debounce be removed as well? Also can we change all implementations that use utilities/debounce use the new usedebounce hook?

src/hooks/useDebounce.ts
```diff
@@ -0,0 +1,16 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+import { useCallback, useRef } from 'react';
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export const useDebounce = <T extends (...args: any[]) => void>(func: T, timeout = 10) => {
+    const timer = useRef<number | undefined>();
+
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const debouncedFunc = useCallback((...args: any[]) => {
```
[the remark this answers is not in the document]
SamCreasey: fixed

src/hooks/useDebounce.ts
SamCreasey: fixed

src/utilities/index.ts
SamCreasey: not yet

fulopdaniel (approved):

PR #1463, opened 2023-06-09
feat(breadcrumbs): add optional vertical gap

fulopdaniel:


celine-s (approved):
lgtm. 
Should someone from the fondue team also have a look.?

src/components/Breadcrumbs/Breadcrumbs.cy.tsx
```diff
@@ -95,6 +96,16 @@ describe('Breadcrumb component', () => {
         cy.get(BREADCRUMB_ITEM_ID).eq(2).find('a').should('exist');
     });
 
+    it('should render gap class appropriately', () => {
+        for (const gap of Object.entries(BreadcrumbGap)) {
```
imoutaharik, line 100: Using for loop in cypress makes the tests flaky, we try to avoid it.

imoutaharik (changes requested):
Thanks for adding the option! 🎉 
I just made a comment for the tests.

imoutaharik (approved):
Thanks!

PR #1465, opened 2023-06-12
fix(AddBlockButton): updating tokens, adding props, adding tests

dominiquemb:


src/components/AddBlockButton/AddBlockButton.tsx
```diff
@@ -39,12 +49,16 @@ export const AddBlockButton = ({
             className={merge([
                 'tw-group tw-leading-none tw-rounded-sm tw-outline-none',
                 isFocusVisible && FOCUS_STYLE,
-                orientation === AddBlockButtonDirection.Vertical ? 'tw-rotate-90' : '',
+                orientation === AddBlockButtonDirection.Vertical ? 'tw-rotate-90' : 'tw-rotate-0',
```
imoutaharik, line 52: Here I would just add maybe `!direction && orientation ...`
So if direction is defined, orientation gets overwritten.

dominiquemb: Done

src/components/AddBlockButton/AddBlockButton.stories.tsx
```diff
@@ -14,12 +15,18 @@ export default {
     tags: ['autodocs'],
     args: {
         title: 'Add Block Button',
+        direction: ButtonDirection.Right,
         orientation: AddBlockButtonDirection.Horizontal,
```
imoutaharik, line 19: You can remove the orientation from the Storybook args.
We deprecate it so no need to document it. We just don't remove from the component yet.

src/components/AddBlockButton/AddBlockButton.stories.tsx
```diff
@@ -14,12 +15,18 @@ export default {
...
+
     argTypes: {
         onClick: {
             action: 'Click',
         },
+        direction: {
+            options: Object.values(ButtonDirection),
+            control: 'select',
+        },
         orientation: {
```
imoutaharik, line 30: To be removed here too :)

src/components/AddBlockButton/AddBlockButton.tsx
```diff
@@ -14,23 +14,51 @@ export enum AddBlockButtonDirection {
...
-    orientation = AddBlockButtonDirection.Horizontal,
+    direction,
+    orientation,
     'data-test-id': dataTestId = 'add-block-button',
 }: AddBlockButtonProps): ReactElement => {
     const { isFocusVisible, focusProps } = useFocusRing();
     const ref = useRef<HTMLButtonElement | null>(null);
     const { buttonProps } = useButton({ onPress: () => onClick() }, ref);
 
+    let rotatePosition;
```
imoutaharik, line 44: This seems to be a bit overkill. Too much if condition here and also on line 70.
We can simplify this by using just the right conditions and mapping through the type using Record.

Remove this whole block, we don't need it and outside the component after line 22, I would just add those:
```
const orientationMap: Record<AddBlockButtonDirection, string> = {
    [AddBlockButtonDirection.Horizontal]: 'tw-rotate-0',
    [AddBlockButtonDirection.Vertical]: 'tw-rotate-90',
};

const directionMap: Record<ButtonDirection, string> = {
    [ButtonDirection.Left]: 'tw-rotate-180',
    [ButtonDirection.Top]: 'tw-rotate-[270deg]',
    [ButtonDirection.Bottom]: 'tw-rotate-90',
    [ButtonDirection.Right]: 'tw-rotate-0',
};
```

src/components/AddBlockButton/AddBlockButton.tsx
```diff
@@ -39,12 +67,17 @@ export const AddBlockButton = ({
             className={merge([
```
imoutaharik, line 67: Here I would simplify using the mapping. We still need to add a default so I would add `tw-rotate-0` in the first line of styling. 

```
className={merge([
                'tw-group tw-leading-none tw-rounded-sm tw-outline-none tw-rotate-0',
                isFocusVisible && FOCUS_STYLE,
                orientation && !direction && orientationMap[orientation],
                direction && directionMap[direction],
            ])}
```

imoutaharik (changes requested):

src/components/AddBlockButton/AddBlockButton.tsx
dominiquemb: Added

src/components/AddBlockButton/AddBlockButton.tsx
imoutaharik: if you give a default value for direction, this line `orientation && !direction && orientationMap[orientation],` may become useless 🤔

imoutaharik (approved):

PR #1466, opened 2023-06-13
feat(BrandColorPicker): fix visibility of icons by adding size to segmentControl

celine-s:
[Before](https://beta-fondue-components.frontify.com/?path=/story/components-color-picker--with-brand-colors):
image

Now:
image

### DIFF
--- a/src/components/ColorPicker/BrandColorPicker.tsx
+++ b/src/components/ColorPicker/BrandColorPicker.tsx
@@ -63,6 +63,7 @@ export const BrandColorPicker = ({ palettes: defaultPalettes = [], currentColor,
                 </div>
                 <div className="tw-w-[72px]">
                     <SegmentedControls
+                        size="small"
                         items={views}
                         activeItemId={view}
                         onChange={(colorView) => setView(colorView as BrandColorView)}


src/components/ColorPicker/BrandColorPicker.tsx
```diff
@@ -63,6 +63,7 @@ export const BrandColorPicker = ({ palettes: defaultPalettes = [], currentColor,
                 </div>
                 <div className="tw-w-[72px]">
                     <SegmentedControls
+                        size="small"
```
joseazevedo88, line 66: Not related to this PR but I find it counterintuitive that a size prop controls the padding. Also that 72px shouldn't be needed, right? This component should take its own space, and the input should fill the rest with `flex: 1`. 

Either way, not super important, just some suggestions for possibly improving the component if there's time. I'll approve 👌

joseazevedo88 (approved):

src/components/ColorPicker/BrandColorPicker.tsx
imoutaharik: @joseazevedo88 It's not pretty right now I agree but the `size` property controls only the padding for now but it may change, that's why we finally decided not to call it `padding`. I have tasks created to discuss different properties once we have a full time Designer joining. (Not until August)

celine-s: Agreed. Was considering to remove it, but the size then would slightly change. Shall I anyway do it or check with a designer or just leave it? @joseazevedo88

celine-s: couldn't agree more. I'll just check it.

celine-s: Couldn't agree more! Will leave it for the moment like this, as the width is needed like this. Will keep it in the back of my head if I can think of a good compromise solution.
Thanks for sharing your thoughts on it!

PR #1468, opened 2023-06-14
fix(Stack): default spacing, padding and margin to 0

imoutaharik:
Better now than later. The new component is still not widely used.


getflourish:
Love it! :) This makes using the Stack more predictable and will cause less unintentional spacing that just happened to be there by default.

dominiquemb (approved):

PR #1469, opened 2023-06-14
chore: Update fondue-tokens to latest version

imoutaharik:


thisguyscode:
@imoutaharik Not sure I'm the right person to review this as it's more about your deployment process that then actual token changes.

thisguyscode:
Theme doesn't seem to affect colors for me. Am I doing something wrong?

Screenshot 2023-06-14 at 16 47 43

thisguyscode:
Ah shit, I also did something wrong on my side. I didn't realise that the 'button' component tokens were using brand tokens (rather than 'box' alias tokens). So that would need to change too... I'll get round to that soon.

Screenshot 2023-06-14 at 16 50 50

imoutaharik:
> Theme doesn't seem to affect colors for me. Am I doing something wrong?
> 
> Screenshot 2023-06-14 at 16 47 43

It does kind of, either on Fondue, beta-fondue or deploy review; it just only noticeable on `line, shadow and some colors in button`

imoutaharik:
> Ah shit, I also did something wrong on my side. I didn't realise that the 'button' component tokens were using brand tokens (rather than 'box' alias tokens). So that would need to change too... I'll get round to that soon.
> 
> Screenshot 2023-06-14 at 16 50 50

But nothing that could break right now, right? I need to update here and web-app because there is an issue of colors in the platform. We have `--box-selected-color: #f1f3f9;` instead of `#F0EAFA`.

thisguyscode:
No I don’t think it’s anything breaking but as I said the aren’t affecting
the tokens view as one would expect so hard to verify

dominiquemb (approved):

PR #1470, opened 2023-06-14
fix(Checklist): added tests, storybook args

dominiquemb:


src/components/Checklist/Checklist.cy.tsx
```diff
@@ -23,22 +23,42 @@ const CHECKBOXES = [
...
+    ariaLabel?: string;
+    direction?: ChecklistDirection;
+    columns?: Columns;
+    activeValues?: string[];
+    setActiveValues?: () => void;
+    'data-test-id'?: string;
+}): ReactElement => {
+    const [activeBoxes, setActiveBoxes] = useState<string[]>(activeValues ? activeValues : []);
 
     return direction === ChecklistDirection.Horizontal ? (
```
imoutaharik, line 46: The Ternary condition with the ChecklistDirection seems unnecessary.
The component could be simplified like this:
```
const [activeBoxes, setActiveBoxes] = useState<string[]>(activeValues ? activeValues : []);

    return (
        <Checklist
            ariaLabel={ariaLabel}
            data-test-id={dataTestId}
            activeValues={activeBoxes}
            setActiveValues={setActiveValues ?? setActiveBoxes}
            direction={direction ?? ChecklistDirection.Vertical}
            checkboxes={CHECKBOXES}
            columns={columns}
        />
    );
 ```

File: src/components/Checklist/Checklist.stories.tsx [hunk 2]
```diff
@@ -20,6 +20,9 @@ export default {
             control: { type: 'radio' },
         },
     },
+    args: {
```
imoutaharik, line 23: if you add the args here, there is no need to add it to every story unless you're changing the value.

src/components/Checklist/Checklist.stories.tsx
```diff
@@ -49,18 +52,22 @@ const COLUMN_CHECKBOXES = Array.from({ length: 8 })
         tooltip: { content: `Random Tooltip ${index}` },
     }));
 
-export const Checklist: StoryFn<ChecklistProps> = (args: ChecklistProps) => {
+export const Checklist = (args: ChecklistProps) => {
```
imoutaharik, line 55: We should keep the `StoryFn`, it sems the CheklistProps were developed in a wrong way.
The missing props could be added this way:
`export const Checklist: StoryFn<ChecklistProps & ChecklistBase> = (args: ChecklistProps & ChecklistBase)`

And add the `export` in CheckList.tsx for the `ChecklistBase` type.

src/components/Checklist/Checklist.stories.tsx
```diff
@@ -49,18 +52,22 @@ const COLUMN_CHECKBOXES = Array.from({ length: 8 })
...
     const [activeBoxes, setActiveBoxes] = useState<string[]>([]);
 
     return (
         <ChecklistComponent
             {...args}
+            ariaLabel={typeof args.ariaLabel === 'string' ? args.ariaLabel : 'checklist'}
```
imoutaharik, line 61: You can remove this, it's part of the args already.

src/components/Checklist/Checklist.tsx
```diff
@@ -103,6 +103,8 @@ export const Checklist = ({
 
     const columns = ('columns' in props && props.columns) || 1;
 
+    console.log(ariaLabel);
```
imoutaharik, line 106: Can be removed

imoutaharik (changes requested):

src/components/Checklist/Checklist.tsx
dominiquemb: Removed

src/components/Checklist/Checklist.stories.tsx
dominiquemb: Removed

src/components/Checklist/Checklist.stories.tsx
dominiquemb: I tried following your suggestion but was still getting lint errors related to the ChecklistVertical | ChecklistHorizontal type, so I re-structured the ChecklistProps instead. Have a look.

src/components/Checklist/Checklist.stories.tsx
dominiquemb: Removed unnecessary args.

src/components/Checklist/Checklist.stories.tsx
```diff
@@ -55,6 +58,7 @@ export const Checklist: StoryFn<ChecklistProps> = (args: ChecklistProps) => {
     return (
         <ChecklistComponent
             {...args}
+            direction={ChecklistDirection.Horizontal}
```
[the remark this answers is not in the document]
dominiquemb: Fixed

imoutaharik (approved):

PR #1471, opened 2023-06-15
fix: Tree bugs and optimization

anxobotana:
Some cleaning and optimization of the Tree component code.
Optimization on how the reducer catches node changes for node with no `contentComponent` props, like some wrappers. This ensures re-render when needed.

### DIFF
--- a/src/components/Tree/Tree.tsx
+++ b/src/components/Tree/Tree.tsx
@@ -121,9 +121,7 @@ const reducer = produce((draft: TreeState, action: TreeStateAction) => {
             {
                 const { id: parentId, children } = action.payload;
 
-                const parentNodeIndex = findIndexById(draft.rootNodes, parentId);
-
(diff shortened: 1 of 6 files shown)


src/components/Tree/Tree.tsx
```diff
@@ -143,6 +141,7 @@ const reducer = produce((draft: TreeState, action: TreeStateAction) => {
                     return;
                 }
 
+                const parentNodeIndex = findIndexById(draft.rootNodes, parentId);
```
fabriciomendonca, line 144: Why to change the way it was done before? With this change, the `findIndexById` will be called twice instead of once. Was there any bug related to that?

anxobotana: no reason, it was left from a experiment... i will revert it 
nice catch!

fabriciomendonca (approved):

PR #1479, opened 2023-06-19
feat(RichTextEditor): add aria-label

celine-s:
[ticket](https://app.clickup.com/t/86782x7u0)


celine-s:
as discussed with @LidijaStoilova we decided to leave it like this

PR #1480, opened 2023-06-20
fix: tree component improvements

anxobotana:
Fixes extra re-renderings, adds logic to expand on hover, and optimizes some logic

### DIFF
--- a/src/components/Tree/Tree.tsx
+++ b/src/components/Tree/Tree.tsx
@@ -227,7 +227,7 @@ export const Tree = memo(
         );
 
         const [treeState, updateTreeState] = useReducer(reducer, initialState);
-        const [isPending, startTransition] = useTransition();
+        const [, startTransition] = useTransition();
 
         const [offset, setOffset] = useState(0);
         const [overId, setOverId] = useState<Nullable<string>>(null);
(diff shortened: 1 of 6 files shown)


src/components/Tree/Tree.tsx
```diff
@@ -101,7 +101,12 @@ const reducer = produce((draft: TreeState, action: TreeStateAction) => {
 
         case 'SET_PROJECTION':
             {
+                if (isEqual(draft.projection, action.payload)) {
+                    return;
+                }
+
                 draft.projection = action.payload;
+                return;
```
fabriciomendonca, line 109: Do we need the added returns in the end? Wouldn't the `break` be enough for the `switch`?

src/components/Tree/Tree.tsx
```diff
@@ -343,7 +352,9 @@ export const Tree = memo(
         };
 
         const handleDragMove = ({ delta }: DragMoveEvent) => {
-            setOffset(delta.x);
+            if (delta.x % 10 === 0) {
```
fabriciomendonca, line 355: Why 10? Maybe extracting it into a constant would explain better the magic number, wouldn't it?

fabriciomendonca (approved):

src/components/Tree/Tree.tsx
anxobotana: no we don't need the `return` it's just a formality, _the immer way_ (https://immerjs.github.io/immer/return)

src/components/Tree/Tree.tsx
anxobotana: yes, or making it relative to the tree padding. The point is reducing the times we refresh the offset which was triggering useless re-renderings

anxobotana: updated!

fabriciomendonca (approved):

PR #1484, opened 2023-06-22
fix(MultiSelect): hide close buttons on tags in MultiSelect when disabled

dominiquemb:


src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -208,7 +208,7 @@ export const MultiSelect = ({
                                         type={getTagType()}
                                         label={key.toString()}
                                         size={size === MultiSelectSize.Small ? TagSize.Small : TagSize.Medium}
-                                        onClick={() => toggleSelection(key)}
+                                        onClick={disabled ? undefined : () => toggleSelection(key)}
```
imoutaharik, line 211: I think this line doesn't need to be changed. Everything was already disabled as the trigger as `pointer-events-none` when disabled. I believe that it was just the presence of the `X` that was misleading it could be interactive.

dominiquemb: @imoutaharik Basically the presence of the onClick triggers the X to be present.  I added the ternary statement to make the onClick undefined if the component is disabled.

PR #1488, opened 2023-06-26
fix(useToggleOverlay): export useToggleOverlay

julianiff:
I think the useToggleOverlay is not yet exported from the hooks folder.


imoutaharik (approved):
LOL, I always forget.

PR #1492, opened 2023-06-29
feat(RTE): allow tabbing to next or prev element

fulopdaniel:
Previously on TAB the RTE blurred and removed focus. However TAB should also focus the next element in the focus tree, so in this PR I am implementing that.

### DIFF
--- a/src/components/RichTextEditor/RichTextEditor.tsx
+++ b/src/components/RichTextEditor/RichTextEditor.tsx
@@ -7,7 +7,7 @@ import { RenderPlaceholderProps } from 'slate-react';
 import { ContentReplacement } from './ContentReplacement';
 import { RichTextEditorProvider } from './context/RichTextEditorContext';
 import { Position } from './EditorPositioningWrapper';
-import { forceToBlurActiveElement } from './helpers';
+import { forceToFocusNextElement } from './helpers';
 import { useEditorState } from './hooks';
 import { GAP_DEFAULT, KEY_ELEMENT_BREAK_AFTER_COLUMN, PluginComposer, defaultPlugins } from './Plugins';
 import { PaddingSizes, TreeOfNodes } from './types';
@@ -83,8 +83,7 @@ export const RichTextEditor = ({
         },
         onKeyDown: (event) => {
             if (event.code === 'Tab') {
-                // Forcing a blur event because of accessibility
-                forceToBlurActiveElement();
+                forceToFocusNextElement(event, !event.shiftKey);
             }
         },
         scrollSelectionIntoView: () => {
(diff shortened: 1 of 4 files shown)


src/components/RichTextEditor/helpers/forceToFocusElement.ts
```diff
@@ -0,0 +1,18 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+export const forceToFocusNextElement = (event: React.KeyboardEvent, forwards: boolean) => {
+    if (!document.activeElement) {
+        return;
+    }
+    event.preventDefault();
+    const focusableElementSelectors =
+        'button, [href], input, select, textarea, [role="textbox"], [tabindex]:not([tabindex="-1"]';
```
ragi96, line 9:
```suggestion
        ''a[href], area[href], input:not([disabled]), select:not([disabled]), textarea:not([disabled]), button:not([disabled]), iframe, object, embed, [tabindex="0"], [contenteditable]';
```

What do you think to adjust them a bit more, so disabled stuff is not focusable

imoutaharik: On the focusTrap hook I'm working on, the list of tabbable elements is like this:
export const TABBABLE_ELEMENTS = [
    'input:not([disabled]):not([type=hidden])',
    'select:not([disabled])',
    'textarea:not([disabled])',
    'button:not([disabled])',
    'a[href]',
    'area[href]',
    'summary',
    'iframe',
    'object',
    'embed',
    'audio[controls]',
    'video[controls]',
    '[contenteditable]',
    '[tabindex]:not([tabindex="-1"]):not([disabled])',
].join(':not([hidden]):not([tabindex="-1"]),');

fulopdaniel: Thanks, great list. Implemented it with some tweaks.

imoutaharik (approved):

src/components/RichTextEditor/RichTextEditor.tsx
```diff
@@ -83,8 +83,7 @@ export const RichTextEditor = ({
         },
         onKeyDown: (event) => {
             if (event.code === 'Tab') {
-                // Forcing a blur event because of accessibility
-                forceToBlurActiveElement();
+                forceToFocusNextElement(event, !event.shiftKey);
```
celine-s, line 86: Is it now still possible to leave the rte again? I now cannot tab outside of the RTE anymore.

src/components/RichTextEditor/helpers/forceToFocusElement.ts
ragi96: @imoutaharik thanks for that improved list.

celine-s (approved):
nice

PR #1494, opened 2023-06-30
refactor(TreeItem): remove focus-within css

fabriciomendonca:
As discussed with Pete, there shouldn't be a background change in the `focus-within` state

### DIFF
--- a/src/components/Tree/TreeItem/TreeItem.tsx
+++ b/src/components/Tree/TreeItem/TreeItem.tsx
@@ -1,6 +1,6 @@
 /* (c) Copyright Frontify Ltd., all rights reserved. */
 
-import React, { Children, MouseEvent, memo, useCallback, useMemo } from 'react';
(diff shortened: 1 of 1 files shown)


src/components/Tree/TreeItem/TreeItem.tsx
```diff
@@ -307,6 +306,10 @@ export const TreeItem = memo(
             </li>
         );
     },
+    (prevProps, nextProps) => {
+        console.log(prevProps, nextProps);
+        return false;
+    },
```
anxobotana, line 312: do we need this?

fabriciomendonca: Oh no. Sorry. Will remove it. Good catch

fabriciomendonca: Done

anxobotana (approved):
LGTM

PR #1495, opened 2023-06-30
fix: updated raw parser [TASK-7579]

bojangles-m:
The clickUp task: https://app.clickup.com/t/2523021/TASK-7579

What was braking? when passing text with html  <br /> the RTE was breaking because it creates Nodes without proper parent tag like Paragraph and looked liked this:
```
[{ "text": "\n" }]
```
This made RTE unresponsive and not capable of editing it anymore.

Solution: Wrapping such a text into paragraph parent which is then parsed in correct object for RTE.

Also refactored a bit of test and added new ones for this case.


imoutaharik (commented):
LGTM!

ragi96 (approved):
Thanks

PR #1501, opened 2023-07-05
feat(RTE): toolbar is stuck open

fulopdaniel:
**New approach:**

On focus out or click out, we check if the target element is part of the RTE or the RTE toolbar or an overlay (modal, e.g. internal link chooser). If not, we force close the toolbar and any modal that is open (e.g. floating link modal).

Reason for this approach:

Simply closing things in the onBlur handler does not work, because e.g. on link chooser click the onBlur method is called, and in that case we do not want to actually close anything.

**Other noteworthy workarounds:**

1. The toolbar gets its position by fetching the bounding box of the selected text area. There can be cases where the selected text area is empty, so the toolbar defaults to x:0, y:0 position, unintendedly jumping to the top left of the screen. [This is the workaround for that.](https://github.com/Frontify/fondue/pull/1501/files#diff-133fab4247d254669b20f98a5137b03821c16a1eb694f3d6021b8823339ddc10R28-R44)

2. On click outside all modals should close. External plugins have their own store (e.g. button plugin in guideline-blocks) that open or close the floating modals. To be able to close those modals as well, I added an optional "hideExternalFloatingModals" prop, so consumers can control what should happen with those modals on outside click.


src/components/RichTextEditor/RichTextEditor.tsx
```diff
@@ -61,7 +63,6 @@ export const RichTextEditor = ({
     const breakAfterPlugin = plugins.plugins.find((plugin) => plugin.key === KEY_ELEMENT_BREAK_AFTER_COLUMN);
     const columns = breakAfterPlugin?.options?.columns ?? 1;
     const columnGap = breakAfterPlugin?.options?.gap ?? GAP_DEFAULT;
-
```
bojangles-m, line 64: Put just space back here please.

bojangles-m (approved):
Good job, LGTM

imoutaharik (commented):
The code looks good to me but I have questions about the UX changes.
Is that fast animation when the RTE appears a new feature? Also, when you choose the link plugin, it opens the flyout but the RTE disappear, is that intended?
Here's a loom to illustrate: https://www.loom.com/share/eb322cb5d3eb4daf8fb37244469623c7?sid=b229ad1e-7369-466b-84b6-b4a99c6b659c

Also some tests are failing and there are not flaky.

fulopdaniel:
> The code looks good to me but I have questions about the UX changes. Is that fast animation when the RTE appears a new feature? Also, when you choose the link plugin, it opens the flyout but the RTE disappear, is that intended? Here's a loom to illustrate: https://www.loom.com/share/eb322cb5d3eb4daf8fb37244469623c7?sid=b229ad1e-7369-466b-84b6-b4a99c6b659c
> 
> Also some tests are failing and there are not flaky.

Good catches, let me fix those issues.

bojangles-m:
Great job, one small issue I found.
Screenshot 2023-07-10 at 09 56 58

bojangles-m (commented):
Small issue with the overlapping of plugins.

src/components/RichTextEditor/BlurObserver.ts
```diff
@@ -0,0 +1,42 @@
...
+        if (isFloatingModalOpen) {
+            floatingLinkActions.hide();
+        }
+    }, [editorRef, isFloatingModalOpen]);
+
+    useEffect(() => {
+        const handleInteractWithOutside = (event: FocusEvent | MouseEvent) => {
+            const richTextEditorSelector = `[data-editor-id='${editorRef.id}']`;
+            const toolbarSelector = "[data-selector='toolbar-floating']";
+            const richTextEditorWrapper = document.querySelector<HTMLDivElement>(richTextEditorSelector);
```
ragi96, line 21: What do you think about using a `ref` for the `richTextEditorWrapper` and `toolbarWrapper`? So we just have to query the DOM once

fulopdaniel: Hmmm.. toolbar does not always exist, and same goes for modals, so we can't really just query them once. We could do it for the RTE wrapper, but at that point I would keep it consistent. What do you think?

src/components/RichTextEditor/EditorPositioningWrapper/ToolbarWrapper/ToolbarWrapperPositioningFloating.tsx
```diff
@@ -24,14 +25,33 @@ export const ToolbarWrapperPositioningFloating = ({
...
+            const selection = window.getSelection();
+            const selectionFrom = selection?.getRangeAt(0).startOffset;
+            const selectionTo = selection?.getRangeAt(0).endOffset;
+            const hasActiveSelection = !!selection && selectionFrom !== selectionTo;
+
+            if (!hasActiveSelection && cleanupFunction.current) {
+                cleanupFunction.current();
+            }
+        });
+    }, []);
```
ragi96, line 39:
```suggestion
    useEffect(() => {
        const handleSelectionChange  = () => {
            const selection = window.getSelection();
            const selectionFrom = selection?.getRangeAt(0).startOffset;
            const selectionTo = selection?.getRangeAt(0).endOffset;
            const hasActiveSelection = !!selection && selectionFrom !== selectionTo;

            if (!hasActiveSelection && cleanupFunction.current) {
                cleanupFunction.current();
            }
        }
        document.addEventListener('selectionchange', handleSelectionChange);
         
        return () => {
            document.removeEventListener('selectionchange', handleSelectionChange);
        }
    }, []);
```

Cleanup function was missing for removing the event listener

src/components/RichTextEditor/RichTextEditor.tsx
```diff
@@ -32,6 +33,7 @@ export type RichTextEditorProps = {
     border?: boolean;
     updateValueOnChange?: boolean; // Only set to true when you are sure that performance isn't an issue
     toolbarWidth?: number;
+    hideExternalFloatingModals?: (editorId: string) => void;
```
ragi96, line 36: Add a test and story with this new prop

src/components/RichTextEditor/context/RichTextEditorContext.tsx
```diff
@@ -9,12 +9,14 @@ export type RichTextEditorContextProps = {
     styles: Record<string, CSSProperties>;
     position: Position;
     wrapperClassNames: string;
+    editorId: string;
```
ragi96, line 12: I don't think the editorId is needed to be stored within the context

fulopdaniel: In PlateWrapper.tsx I need to add the editor id to the wrapper tag. This seems the most straightforward way, unless you have something in mind.

src/components/RichTextEditor/RichTextEditor.tsx
fulopdaniel: I will, once this solution has been agreed upon.

src/components/RichTextEditor/BlurObserver.ts
ragi96: True, didn't think about that its completely removed from the DOM

ragi96 (commented):
Great, just the missing test and the code smells by sonar

PR #1502, opened 2023-07-05
fix: Tree clean empty items to prevent phantom items

anxobotana:
When removing all children from a parent, the last child is kept in the UI (as a phantom item). Ensure we unregister them correctly


fabriciomendonca (approved):
Works as expected

PR #1503, opened 2023-07-06
feat!: Dialog Subcomponents improvements

imoutaharik:


src/types/dialog.ts
```diff
@@ -24,11 +28,15 @@ export type DialogHeaderProps = {
 export type DialogBodyProps = {
     children: ReactNode;
     maxHeight?: number | string;
```
selimovicz, line 30: Is there a reason why we accept `maxHeight` as `number | string`, then to use it like `style={{maxHeight}}`. Should we maybe force user to use some predefined sizes, by strongly typing this property.

src/types/dialog.ts
```diff
@@ -24,11 +28,15 @@ export type DialogHeaderProps = {
...
+    padding?: DialogSize;
     'data-test-id'?: string;
 };
 
 export type DialogFooterProps = {
-    buttons: ButtonProps[];
+    children?: ReactNode;
+    actionButtons: [ButtonProps] | [ButtonProps, ButtonProps];
```
selimovicz, line 37: What if user in some case wants to have more that 2 actions buttons, he will not be able to specify those. I know that scenario is less likely to happen, but still. I would do simply:
```suggestion
    actionButtons: ButtonProps[]
```

selimovicz (approved):
All good here. Awesome work Ilham. I left two comments there, if you have time to address those. Thanks.

src/types/dialog.ts
imoutaharik: Not yet. For now it will be flexible as we would need a designer to decide the predefined sizes we should have in the platform.

PR #1504, opened 2023-07-06
fix(Tree): ignore keydown if element is not a tree item

fabriciomendonca:
The keydown callback was running for a `TreeItem` children when it is not supposed to, so we will return early if the active item role is not `treeitem`


anxobotana (approved):
LGTM

PR #1517, opened 2023-07-12
fix(Trigger): Caret not focusable on Dropdown and MultiSelect.

imoutaharik:
This makes the caret on the Dropdown and MultiSelect not focusable.


LidijaStoilova (approved):
Keyboard Accessibility is working correctly

PR #1523, opened 2023-07-20
fix(Rich Text Editor): test for missing mapper in HTML serializer

bojangles-m:


src/components/RichTextEditor/serializer/utils/serializeNodeToHtmlRecursive.ts
```diff
@@ -22,6 +22,7 @@ import {
     TDescendant,
     TElement,
     isText,
+    // ELEMENT_IMAGE,
```
imoutaharik, line 25: If it's not in use, It's better to remove it.

imoutaharik (approved):

PR #1524, opened 2023-07-24
feat(icons): add unique clip path id to all icons

fulopdaniel:
Some icons have clipPath builtin e.g. `IconChartPie12` or `IconDoAndDont20` . Respectively they both have a clip path that clips a 12x12 and a 20x20 area. However the IDs of these clip paths are the same prefix__a  which results in conflicting.

Bonus: Fix RTE focus issue.

### DIFF
--- a/src/components/RichTextEditor/helpers/forceToFocusElement.ts
+++ b/src/components/RichTextEditor/helpers/forceToFocusElement.ts
@@ -13,7 +13,7 @@ const TABBABLE_ELEMENTS = [
     'embed',
     'audio[controls]',
     'video[controls]',
-    '[contenteditable]',
+    '[contenteditable="true"]',
     '[tabindex]',
 ].join(':not([hidden]):not([tabindex="-1"]):not([disabled]),');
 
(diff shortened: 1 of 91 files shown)


imoutaharik (approved):

src/components/RichTextEditor/helpers/forceToFocusElement.ts
```diff
@@ -13,7 +13,7 @@ const TABBABLE_ELEMENTS = [
     'embed',
     'audio[controls]',
     'video[controls]',
-    '[contenteditable]',
+    '[contenteditable="true"]',
```
ragi96, line 16: Not related to that PR

fulopdaniel: I know, but I intentionally commited this as this will be needed for the Do&Dont as well. Would you prefer to open a separate PR for this?

ragi96: Yes would be cool if you open another PR for that so we have it clean

ragi96 (approved):

PR #1528, opened 2023-07-27
chore: fondue tokens updated

bojangles-m:


imoutaharik (commented):
@bojangles-m This change is creating an accessibility issue on dark mode.
Elements must have sufficient color contrast.

thisguyscode:
@bojangles-m & @imoutaharik - this one's my bad. I gave @bojangles-m the wrong instructions.

`text-weak` = `grey.40`

'text-x-weak' = `grey.50` (which doesn't need to pass accessibility requirements because it's reserved for non-essential text)

Sorry things are a bit of a rush before holidays.

PR #1530, opened 2023-08-01
feat(RTE): add dir to serializer

fulopdaniel:
https://app.clickup.com/t/865cruy5y


ragi96 (approved):
Great seems working on all elements 

PS: don't forget to add it also for the blocks serializer

imoutaharik (approved):

PR #1543, opened 2023-08-09
refactor: Tree in `Flyout` and `InlineDialog` support

Kenny806:
This ensures that Tree in compatible with the new `InlineDialog`.


anxobotana (approved):
LGTM, but it seems that we need to revisit how the background behaves in a flyout with draggable items.

src/components/Tree/Tree.stories.tsx
```diff
@@ -327,3 +328,20 @@ export const WithExpandOnSelect = ({ ...args }: TreeProps) => {
...
     );
 };
+
+export const InsideFlyout = ({ ...args }: TreeProps) => {
+    const [isOpen, setIsOpen] = useState(false);
+
+    return (
+        <div style={{ width: 800 }}>
+            <Flyout
+                trigger={<button type={'button'}>{'Open Tree'}</button>}
```
imoutaharik, line 338: It's fine for me to add this story but can we style a bit so it doesn't look like bare minimum. I would use a Fondue Button here. 
`trigger={<Button>Open Tree</Button>}`

src/components/Tree/Tree.stories.tsx
```diff
@@ -327,3 +328,20 @@ export const WithExpandOnSelect = ({ ...args }: TreeProps) => {
         </div>
     );
 };
+
+export const InsideFlyout = ({ ...args }: TreeProps) => {
+    const [isOpen, setIsOpen] = useState(false);
+
+    return (
+        <div style={{ width: 800 }}>
```
imoutaharik, line 336: Just a detail here but you could use our Layout component instead of div -> either `Box` or `Container`

anxobotana:
@imoutaharik I will take the PR and fix among those things you commented, the background issues when in a flyout.

src/components/Tree/Tree.stories.tsx
```diff
@@ -327,3 +333,50 @@ export const WithExpandOnSelect = ({ ...args }: TreeProps) => {
...
+                    <DialogBody>
+                        <TreeView
+                            {...args}
+                            expandedIds={expandedIds}
+                            onExpand={handleItemExpand}
+                            onShrink={handleItemShrink}
+                        >
+                            {treeItemsMock.map(renderTreeItemLabel)}
+                        </TreeView>
+                    </DialogBody>{' '}
```
imoutaharik, line 377: what's the {'  '} for?

anxobotana: nothing... just missed it when cleaning :)  good catch!
cleaned now!

imoutaharik (approved):

PR #1544, opened 2023-08-11
fix(RTE): serializer line breaks

ragi96:
[Task](https://app.clickup.com/t/861n2wyq1)


src/components/RichTextEditor/serializer/utils/serializeLeafToHtml.ts
```diff
@@ -12,7 +12,8 @@ import escapeHtml from 'escape-html';
 
 export const serializeLeafToHtml = (node: TText): string => {
     let string = escapeHtml(node.text);
-    string = string.replaceAll('\n', '<br />');
+    string === '' && (string = '&#xFEFF;');
```
fulopdaniel, line 15: Question: why is this needed?

fulopdaniel: also nit, bit I was a bit confused at first when looking at this line, maybe
```suggestion
    if (string === '') {
        string = '&#xFEFF;'
    }
```

src/components/RichTextEditor/serializer/serializeToHtml.ts
```diff
@@ -46,14 +46,9 @@ export const serializeNodesToHtml = (
 
     let html = '';
     for (let i = 0, len = nodes.length; i < len; i++) {
-        const node = nodes[i];
-        if (isEmptyNode(node)) {
-            html += '<br />';
```
fulopdaniel, line 51: Is it fine to remove this? I figured you might need to add the empty character here as well?

ragi96: The problem with the solution we had in place was that you are gonna loose information of empty lines, as they were just `<br/>`, but the line-height of the selected text-style was removed as it didn't know about that. So now we are parsing all empty lines as correct leafs, which will end up in correct spacing.

src/components/RichTextEditor/serializer/utils/serializeLeafToHtml.ts
ragi96: `Question: why is this needed?`

It's needed for hard breaks, so we don't get empty h1 or other tags that are not displayed. So we ensure that the tags are always rendered.

fulopdaniel (approved):

PR #1547, opened 2023-08-16
fix: dropdown caret positioning

SamuelAlev:
Fix alignment
Before:
Screenshot 2023-08-16 at 20 27 30

After:
Screenshot 2023-08-16 at 20 34 17

### DIFF
--- a/src/components/Trigger/Trigger.tsx
+++ b/src/components/Trigger/Trigger.tsx
@@ -113,28 +113,27 @@ export const Trigger = ({
                         <IconTrashBin size={IconSize.Size12} />
                     </button>
                 )}
-                <div aria-hidden={true}>
-                    <button
-                        {...buttonProps}
-                        type="button"
-                        tabIndex={-1}
-                        disabled={disabled}
-                        className={merge([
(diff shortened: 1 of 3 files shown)


bojangles-m (approved):

src/components/Trigger/Trigger.tsx
```diff
@@ -113,28 +113,27 @@ export const Trigger = ({
...
-                                      FOCUS_VISIBLE_STYLE,
-                                  ]),
-                        ])}
-                    >
-                        <div className={merge(['tw-transition-transform', isOpen && 'tw-rotate-180'])}>
-                            <IconCaretDown size={IconSize.Size16} />
-                        </div>
-                    </button>
-                </div>
+                <button
```
imoutaharik, line 116: Did you check the Axe devTools? I remember I did that change and when reviewing with Lidija, it was apparently necessary to have the aria-hidden on a parent level and not the button.

SamuelAlev: Yup, nothing new appeared

PR #1548, opened 2023-08-17
demo: Fondue animations

SamCreasey:


imoutaharik:
@SamCreasey Any objections if we close that PR for now? We will look into it and may reopen when we get to the topic.

imoutaharik:
Closing for now as animations is not part of the current timeline.

PR #1553, opened 2023-08-30
ci: fix deprecation notice

SamuelAlev:
Screenshot 2023-08-30 at 14 15 58

Also:
- Cleaned the prefixes for pr title validation
- Removed automerge on renovate

### DIFF
--- a/.github/semantic.yml
+++ b/.github/semantic.yml
@@ -1,18 +1,14 @@
 titleOnly: true
 types:
-  - feat
-  - fix
-  - docs
-  - dx
-  - refactor
-  - perf
-  - test
-  - workflow
-  - build
-  - ci
-  - chore
-  - types
-  - wip
-  - release
-  - deps
-  - style
+    - feat
+    - fix
+    - docs
+    - refactor
+    - perf
+    - test
+    - ci
+    - chore
+    - release
+    - deps
+    - style
+    - ci
--- a/.github/semantic.yml
+++ b/.github/semantic.yml
@@ -8,7 +8,5 @@ types:
     - test
     - ci
     - chore
-    - release
-    - deps
     - style
     - ci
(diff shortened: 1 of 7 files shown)


.github/workflows/continuous-deployment.yml
```diff
@@ -35,7 +35,7 @@ jobs:
 
             - name: Get pnpm store directory
               id: pnpm-cache
-              run: echo "::set-output name=pnpm_cache_dir::$(pnpm store path)"
+              run: echo "pnpm_cache_dir=$(pnpm store path)" >> $GITHUB_OUTPUT
```
ryancarville, line 38: Doesn't the `$GITHUB_OUTPUT` need to be wrapped in `"`? Looks like this one and `$GITHUB_ENV` use them.  
See doc here:
https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-output-parameter

SamuelAlev: Works the same, the data is still passed https://github.com/Frontify/fondue/actions/runs/6025904700/job/16347675219#step:6:3

ryancarville (approved):
LGTM 👍

.github/semantic.yml
```diff
@@ -1,18 +1,14 @@
...
+    - docs
+    - refactor
+    - perf
+    - test
+    - ci
+    - chore
+    - release
+    - deps
+    - style
+    - ci
```
imoutaharik, line 14: @SamuelAlev `ci` prefix is not recognized for the PR title validation. We could add the same types on `pull-request-title.yml`

SamuelAlev: Fixed

imoutaharik (approved):

PR #1569, opened 2023-09-11
chore: rename OrderableList to LegacyOrderableList

anxobotana:
In order to replace `react-dnd` we need to replace OrderableList, the only component using it.
First step use LegacyOrderableList, this should work the same way as OrderableList

### DIFF
--- a/src/components/LegacyOrderableList/LegacyOrderableList.cy.tsx
+++ b/src/components/LegacyOrderableList/LegacyOrderableList.cy.tsx
@@ -122,7 +122,7 @@ describe('OrderableList Component', () => {
         cy.get(DRAGGABLE_ITEM).each(($el) => {
             const dataTransfer = new DataTransfer();
             cy.wrap($el)
-                .should('have.text', LegacyLegacyItemDragState.Idle)
+                .should('have.text', LegacyItemDragState.Idle)
                 .trigger('dragstart', { dataTransfer })
                 .trigger('drag');
             cy.wrap($el).should('have.text', LegacyItemDragState.Dragging);
--- a/src/components/LegacyOrderableList/LegacyOrderableList.cy.tsx
+++ b/src/components/LegacyOrderableList/LegacyOrderableList.cy.tsx
@@ -1,8 +1,8 @@
 /* (c) Copyright Frontify Ltd., all rights reserved. */
 
 import {
-    ItemDragState,
     LegacyDragProperties,
+    LegacyItemDragState,
     LegacyOrderableList,
     LegacyOrderableListItem,
     LegacyOrderableListProps,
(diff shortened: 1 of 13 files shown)


src/components/LegacyOrderableList/LegacyOrderableList.cy.tsx
```diff
@@ -0,0 +1,167 @@
...
+        const renderWithDragState = (_item: TestItem, { componentDragState }: LegacyDragProperties) => (
+            <div style={{ height: `${ITEM_HEIGHT}px` }}>{componentDragState}</div>
+        );
+
+        cy.mount(<OrderableListWithDefaultProps renderContent={renderWithDragState} />);
+
+        cy.get(DRAGGABLE_ITEM).each(($el) => {
+            const dataTransfer = new DataTransfer();
+            cy.wrap($el)
+                .should('have.text', LegacyLegacyItemDragState.Idle)
```
imoutaharik, line 125: Got a Legacy too many here.

src/components/LegacyOrderableList/LegacyOrderableList.stories.tsx
```diff
@@ -0,0 +1,63 @@
...
+import { Meta, StoryFn } from '@storybook/react';
+import { useState } from 'react';
+import { LegacyOrderableList as OrderableListComponent } from './LegacyOrderableList';
+import { LegacyOrderableListItem } from './types';
+import { LegacyOrderableListProps } from '.';
+import { chain } from '@react-aria/utils';
+import { renderContent, storyItems } from '@components/LegacyOrderableList/utils';
+
+export default {
+    title: 'Components/Legacy Orderable List',
```
imoutaharik, line 12: You could move this to `Deprecated` category in Storybook -> `Deprecated/Legacy OrderableList`

src/components/index.ts
```diff
@@ -32,6 +32,7 @@ export * from './FrontifyPattern';
 export * from './Grid';
 export * from './InlineDialog';
 export * from './InputLabel';
+export * from './LegacyOrderableList';
```
imoutaharik, line 35: Did you just copy OrderableList and created the Legacy?
Are you keeping the old one as `OrderableList` for later refactoring ?

imoutaharik (commented):
There is only 3 instances of that component in web-app that could be renamed pretty quickly. But before refactoring the whole component we should make sure there is no impact in other repositories that uses Fondue like `guidelines-block`.

src/components/index.ts
anxobotana: Yup, the idea is having a new `OrderableList` using `Tree` component.
In fact i have already the new component almost ready

imoutaharik (approved):

anxobotana:
@imoutaharik can you merge this PR? 
SonarCloud complains about duplication are intended, LegacyOrderableList is a copy of OrderableList, a follow up PR will implement the new one

PR #1574, opened 2023-09-14
chore(datepicker): reset the css for children container

joseazevedo88:
Updating the datepicker package led to introducing breaking styles in the datepicker children container


joewoodson (approved):
LGTM

PR #1581, opened 2023-09-15
feat: added focusOnMount and onEnterPressed props to textarea

thomas-tosch:
I added the following props:
- focusOnMount
- onEnterPressed

They behave the same way as in the input component.

I was requiring those props to integrate the textarea into an ongoing feature.

### DIFF
--- a/src/components/Textarea/Textarea.stories.tsx
+++ b/src/components/Textarea/Textarea.stories.tsx
@@ -16,6 +16,7 @@ export default {
         resizeable: true,
         selectable: false,
         validation: Validation.Default,
+        focusOnMount: false,
     },
     argTypes: {
         value: { type: 'string' },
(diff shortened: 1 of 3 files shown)


src/components/Textarea/Textarea.tsx
```diff
@@ -47,9 +49,25 @@ export const Textarea = ({
...
     selectable = false,
+    focusOnMount,
+    onEnterPressed,
 }: TextareaProps): ReactElement => {
     const Component = autosize ? TextareaAutosize : 'textarea';
 
+    const textareaElement = useRef<HTMLTextAreaElement | null>(null);
+
+    useEffect(() => {
+        setTimeout(() => {
```
imoutaharik, line 60: Why do we need the setTimeout? It seems unnecessary to me, especially with a delay set to 0.

src/components/Textarea/Textarea.stories.tsx
```diff
@@ -31,6 +32,8 @@ export default {
         },
         minRows: { type: 'number' },
         maxRows: { type: 'number' },
+        onEnterPressed: { action: 'onEnterPressed' },
+        focusOnMount: { type: 'boolean' },
```
imoutaharik, line 36: Please also add a story for that.

imoutaharik (changes requested):

imoutaharik (approved):

PR #1588, opened 2023-09-21
fix: apply text disabled text color for the tab label

mnoleto:
TabItem disabled state
- `text-weak` is overriding the `text-disabled` and we can't see the right color

Screenshot 2023-09-21 at 10 24 39

https://www.figma.com/file/MmTUiRZXSDXtewPb7JYoOg/Metadata-(Permissions-%26-Dependencies)?type=design&node-id=5307-48196&mode=design&t=3c0p2euhnY4rKNTf-0

### DIFF
--- a/src/components/Tabs/Tabs.tsx
+++ b/src/components/Tabs/Tabs.tsx
@@ -229,9 +229,8 @@ export const Tabs = ({
                                 id={`${tab.id}-btn`}
                                 className={merge([
                                     'tw-group tw-relative tw-mx-0 tw-py-4 tw-px-2 tw-w-max tw-cursor-pointer tw-flex tw-items-center tw-justify-center tw-whitespace-nowrap',
-                                    tab.disabled && 'tw-text-text-disabled',
-                                    !tab.disabled && 'hover:tw-text-text',
-                                    tab.id === activeItemId ? 'tw-font-medium tw-text-text' : 'tw-text-text-weak',
+                                    tab.id === activeItemId && 'tw-font-medium tw-text-text',
+                                    tab.disabled ? 'tw-text-text-disabled' : 'tw-text-text-weak hover:tw-text-text',
                                     size === TabSize.Small ? 'tw-text-sm' : 'tw-text-md',
                                 ])}
                                 key={tab.id}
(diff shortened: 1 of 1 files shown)


src/components/Tabs/Tabs.tsx
```diff
@@ -296,8 +296,8 @@ export const Tabs = ({
                                     return (
                                         <button
                                             className={merge([
-                                                'tw-flex tw-items-center tw-w-full tw-mb-3 tw-text-left tw-text-text-weak',
-                                                tab.disabled && 'tw-text-text-disabled',
+                                                'tw-flex tw-items-center tw-w-full tw-mb-3 tw-text-left',
```
imoutaharik, line 299: @mnoleto You are making the change on the overflow menu here. I think the issue is on line `232` to`234`
```
tab.disabled && 'tw-text-text-disabled',
!tab.disabled && 'hover:tw-text-text',
tab.id === activeItemId ? 'tw-font-medium tw-text-text' : 'tw-text-text-weak',

File: src/components/Tabs/Tabs.tsx [hunk 2]
```diff
@@ -229,9 +229,8 @@ export const Tabs = ({
                                 id={`${tab.id}-btn`}
                                 className={merge([
                                     'tw-group tw-relative tw-mx-0 tw-py-4 tw-px-2 tw-w-max tw-cursor-pointer tw-flex tw-items-center tw-justify-center tw-whitespace-nowrap',
-                                    tab.disabled && 'tw-text-text-disabled',
-                                    !tab.disabled && 'hover:tw-text-text',
-                                    tab.id === activeItemId ? 'tw-font-medium tw-text-text' : 'tw-text-text-weak',
+                                    tab.id === activeItemId && 'tw-font-medium tw-text-text',
```
imoutaharik, line 232: @mnoleto  tab.id line should go after. Because the tab.disabled ternary overwrites the `tw-text-text` class and change it to `tw-text-text-weak`.

mnoleto: Can we have a tab disabled and active at the same time?

imoutaharik: No you can't. but on line 233 it checks if disabled and if not disabled, it uses text-weak and it overwrites the active state with text-text styling.

mnoleto: Done!

PR #1590, opened 2023-09-21
fix: table header shouldn't overlay other content

Kenny806:
Table header had z-index: 10 in order to avoid being overlaid by table content in mobile view (when the header is sticky). This high z-index value causes it to overlay other elements on the page, such as a sticky header.  The solution should be much less likely to conflict with other elements.
Before:
Snímek obrazovky 2023-09-21 v 14 10 25

After:
Snímek obrazovky 2023-09-21 v 14 06 52

### DIFF
--- a/src/components/Table/Table.tsx
+++ b/src/components/Table/Table.tsx
@@ -162,7 +162,7 @@ export const Table = ({
                         </TableHeaderRow>
                     ))}
                 </thead>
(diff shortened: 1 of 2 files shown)


joseazevedo88 (approved):
LGTM, wdyt about having `sm:tw-z-auto` which is the default?

PR #1595, opened 2023-09-26
chore: add box color token mapping to text component

ryancarville:
Add `boxColor token` mapping to text component for optional accessibility usage.


src/typography/Text/Text.tsx
```diff
@@ -42,12 +59,32 @@ const colorMap: Record<TextColor, string> = {
     interactive: 'tw-text-text-interactive',
 };
 
+/** @description optional background box color used when accessibility contrast is needed */
+const boxColorMap: Record<BoxColor, string> = {
+    default: '',
+    base: 'tw-bg-base',
```
imoutaharik, line 65: @ryancarville We don't have a box color `base` or even `base-alt`, for those background color, we already have the TextColor mapping which is right.  Also the tailwind class is wrong. We are not setting the background color on the Text component but the right Text color depending if it's using a Box token as container.

So it should be:
`neutral: 'tw-text-box-neutral-inverse'`

src/typography/Text/Text.tsx
```diff
@@ -62,6 +99,7 @@ export const Text = ({
             weightMap[weight],
             sizeMap[size],
             colorMap[color],
+            boxColorMap[boxColor],
```
imoutaharik, line 102: I think that default option being empty string is confusing. Can we just make the property optional and have `boxColor && boxColorMap[boxColor]` on line 102 ?

src/typography/Text/Text.tsx
ryancarville: @imoutaharik understood.  Will make the required changes.

src/typography/Text/Text.tsx
ryancarville: No problem

src/typography/Text/Text.stories.tsx
```diff
@@ -24,6 +24,10 @@ export default {
             options: ['default', 'weak', 'x-weak', 'disabled', 'negative', 'positive', 'warning', 'interactive'],
             control: { type: 'select' },
         },
+        boxColor: {
+            options: [undefined, 'neutral', 'selected', 'disabled', 'positive', 'negative', 'warning'],
```
imoutaharik, line 28: When you put undefined as an option, it shows as "Undefined" on the dropdown. I think it will look better to have an empty string. It will interpret it as undefined.

imoutaharik (changes requested):
Thanks for the changes :). It would also be nice to add a story with a boxColor example when the Text component is wrapped in a container using box token.

imoutaharik (approved):

imoutaharik:
@ryancarville You just have one test to fix on the Text component. It's failing for the textColor classes, it's not flaky.

imoutaharik (approved):

PR #1602, opened 2023-09-29
chore: add axis padding, axis margin and axis gap props

ryancarville:
Adds `paddingX`, `paddingY`, `marginX`, `marginY`, `spacingX`, and `spacingY` props and logic to layout components. 

(FYI: `spacing` translates to `gap` css property)

Sprint Card: https://app.clickup.com/t/867845k1e


src/components/Flex/Flex.tsx
```diff
@@ -30,6 +33,10 @@ export type FlexAlignContent =
 export type FlexProps = {
     direction?: FlexDirection;
     wrap?: FlexWrap;
+    gap?: SpacingValue;
```
imoutaharik, line 36: @ryancarville While trying out with the control on the story, I'm a bit confused between gap and spacing. Gap doesn't seem to work at all.

ryancarville: @imoutaharik good spot.  The `gap` prop was initially added to align the prop terminology with the css properties, but then I realised it would be a breaking change.  `spacing` is used for the `gap` css value and I will remove `gap` from the type definition.

imoutaharik (approved):

PR #1609, opened 2023-10-10
feat(NumberInput): onChange value and negative sign on clear

ryancarville:
1.  Update to the `onChange` prop return value.  Previously, it would send a `0` if empty (`undefined`) which tells the `parent` there is still a `value` of `0` for the `number` input.  This needed to be changed to enable the passing of a `undefined` if no value existed.

2.  The `-` symbol would persist when the `value` was deleted if the `sufffix` `prop` was present (input of type `string`) and caused a `NaN` error.  Logic to remove the `-` symbol if `cleared` or `backspaced` was added.


src/components/NumberInput/NumberInput.tsx
```diff
@@ -92,7 +92,8 @@ export const NumberInput = ({
         const { key, shiftKey } = event;
         if (SPECIAL_KEYS.includes(key)) {
             const currentValue = getCurrentValueWithoutSuffix();
-            if (currentValue?.length === 0) {
+            const isNegativeSymbol: boolean = currentValue.length === 2 && currentValue.includes('-');
```
SamuelAlev, line 95: Can you add a test case to avoid regression?

src/components/NumberInput/NumberInput.tsx
```diff
@@ -167,7 +168,7 @@ export const NumberInput = ({
 
     const handleOnChange = useCallback(() => {
         const valueWithoutSuffix = getCurrentValueWithoutSuffix();
-        onChange?.(Number(valueWithoutSuffix));
+        onChange?.(valueWithoutSuffix ? Number(valueWithoutSuffix) : undefined);
```
bojangles-m, line 171: why dont you pass it just 0? and could it be possible you dont need to change the types?

src/types/input.ts
```diff
@@ -9,6 +9,6 @@ export type InputBaseProps<TValue> = {
     suffix?: string;
     status?: Validation;
     valueSelect?: boolean;
-    onChange?: (value: TValue) => void;
+    onChange?: (value?: TValue) => void;
```
bojangles-m, line 12: If you pass 0 you can revert this.

src/components/NumberInput/NumberInput.stories.tsx
```diff
@@ -79,7 +79,7 @@ export const WithDecoratorAndIncrementable: StoryFn<NumberInputProps> = (args) =
 export const WithErrorLessThanZero: StoryFn<NumberInputProps> = (args) => {
     const [status, setStatus] = useState<Validation>(Validation.Success);
     const [valueSelect, setValueSelect] = useState<boolean>(false);
-    const handleValidation = (value: number) => {
+    const handleValidation = (value: number = 0) => {
```
bojangles-m, line 82: If you pass undefined, value will turn to 0

ryancarville: True, but i set a `default` value to avoid the `type` check issues in the conditions that follow

src/types/input.ts
ryancarville: But the issue with `0` is what happens when there input is cleared?  Being a `number` input `0` is a valid value.  We must be able to pass a `undefined`.

src/components/NumberInput/NumberInput.tsx
ryancarville: test coverage added

src/components/NumberInput/NumberInput.tsx
bojangles-m: Maybe I asked wrongly, the question was for undefined. Correct me if I am wrong but if you pass undefined to the onChange function it will turn into 0?

ryancarville: That is 100% correct if the component stays as it is. But that is why this PR is required. We have to allow `undefined` to avoid it returning `0` since this is a `number` input.

bojangles-m (approved):

src/components/NumberInput/NumberInput.tsx
bojangles-m: @ryancarville got you, thnx for explanation

bojangles-m (approved):

src/components/NumberInput/NumberInput.tsx
ryancarville: Anytime. Thank you for your review!

SamuelAlev (approved):

Issue #621, opened 2022-02-16
TextInput: onEnterPressed & onChange behave differently 


getflourish:
I’m trying to use a `TextInput` and only want to get the value when the user presses "enter". 

I would expect to listen to the `onChange` event. Unexpectedly, `onChange` fires on every key press. I figured out that React goes against web standards here. 🤦‍♂️

I then looked into the source code for the `TextInput` and noticed that there’s a `onEnterPressed` callback prop that I can use. It works, but the return value is inconsistent with the `onChange` callback prop. 

onChange: returns `value`
onEnterPressed: returns `SyntheticBaseEvent`

It would be nice if both callback props would behave similar. :)

Usage right now:

```js
onChange={(value) => {
  // → value
}}
onEnterPressed={(event) => {
 // → event.target.value
}}
```

ryancarville:
@getflourish @jonasbuechel I have made a PR for this and it is approved, but it is a breaking change so wanted to get your feedback before I merged.  At your connivence.  Thanks!

PR #1612, opened 2023-10-10
[WIP]fix(TextInput)!: update onEnterPress prop to return value

ryancarville:
Fix for reported issue: https://github.com/Frontify/fondue/issues/621


src/components/TextInput/TextInput.tsx
```diff
@@ -61,7 +61,7 @@ export type TextInputBaseProps = {
     copyable?: boolean;
     value?: string;
     onChange?: (value: string) => void;
-    onEnterPressed?: (event: KeyboardEvent<HTMLInputElement>) => void;
```
imoutaharik, line 64: Please mark this PR as a breaking change.

ryancarville: @imoutaharik yes, I will.  Also I asked from some feedback on current usage.  But of course doesn't change the fact it could cause issues.  Open to discussing this PR's value as well.

imoutaharik: You'll need to use the `!` otherwise the check for Pr title won't pass.
`fix(TextInput)!: update onEnterPress...`

src/components/TextInput/TextInput.tsx
```diff
@@ -181,7 +181,7 @@ export const TextInput = ({
 
     const onKeyDown = (event: KeyboardEvent<HTMLInputElement>) => {
         if (event.key === 'Enter' && onEnterPressed) {
-            onEnterPressed(event);
+            onEnterPressed(inputElement.current ? inputElement.current.value : '');
```
bojangles-m, line 184:
```suggestion
            onEnterPressed(inputElement.current?.value ?? '');
```

bojangles-m (approved):

src/components/TextInput/TextInput.tsx
ryancarville: request updated

src/components/TextInput/TextInput.tsx
```diff
@@ -61,7 +61,7 @@ export type TextInputBaseProps = {
     copyable?: boolean;
     value?: string;
     onChange?: (value: string) => void;
-    onEnterPressed?: (event: KeyboardEvent<HTMLInputElement>) => void;
+    onEnterPressed?: (value: string) => void;
```
imoutaharik, line 64: If no one gave you feedback, you can also search the usage in the web-app yourself to asses the impact of the change.

ryancarville: Looks like the prop `onEnterPressed` is used 33% of all `TextInput` implementation in `web-app` as a method handler.  Will need to restructure this to be more versatile

imoutaharik: @ryancarville What's the status of this PR, Should we close?

ryancarville: We can close it actually since we are going to make a new component anyways and this change would be a breaking change. Will keep it in mind for the new one though.

PR #1613, opened 2023-10-11
fix: block scrolling on modal

bojangles-m:
The loom shows scrolling when modal is not open and blocked scroll when modal is open. You cant see by i try to scroll when modal is open but it does not move :)
https://www.loom.com/share/fd863df2fa93459fa26a83886bff01e5

This video shows the behaviour before this changes
https://www.loom.com/share/bcad8fe83a8a4fe88b569b8bb0ccbd38

### DIFF
--- a/src/components/Dialog/Dialog.stories.tsx
+++ b/src/components/Dialog/Dialog.stories.tsx
@@ -25,6 +25,7 @@ export default {
     tags: ['autodocs'],
     args: {
         modality: Modality.Modal,
+        strategy: 'absolute',
         minWidth: 600,
         maxWidth: 0,
         autoHeight: false,
--- a/src/components/Dialog/Dialog.stories.tsx
+++ b/src/components/Dialog/Dialog.stories.tsx
@@ -41,6 +41,10 @@ export default {
             options: Object.values(Modality),
             control: { type: 'select' },
         },
+        strategy: {
+            options: ['absolute', 'fixed'],
+            control: { type: 'radio' },
+        },
         'data-test-id': {
             type: 'string',
         },
--- a/src/components/Dialog/Dialog.tsx
+++ b/src/components/Dialog/Dialog.tsx
@@ -40,6 +40,7 @@ export const Dialog = ({
             zIndex={Z_INDEX_MODAL}
             isDetached={true}
             verticalAlignment={verticalAlignment}
+            strategy={modality === Modality.NonModal ? 'absolute' : 'fixed'}
         >
             {children}
         </Overlay>
--- a/src/components/InlineDialog/InlineDialog.stories.tsx
+++ b/src/components/InlineDialog/InlineDialog.stories.tsx
@@ -29,7 +29,7 @@ export default {
         flip: true,
         minWidth: 0,
         maxWidth: 400,
-        autoHeight: true,
+        autoHeight: false,
         maxHeight: 'auto',
         minHeight: 0,
         darkUnderlay: false,
--- a/src/components/InlineDialog/InlineDialog.stories.tsx
+++ b/src/components/InlineDialog/InlineDialog.stories.tsx
@@ -29,7 +29,7 @@ export default {
         flip: true,
         minWidth: 0,
         maxWidth: 400,
-        autoHeight: false,
+        autoHeight: true,
         maxHeight: 'auto',
         minHeight: 0,
         darkUnderlay: false,
--- a/src/components/InlineDialog/InlineDialog.tsx
+++ b/src/components/InlineDialog/InlineDialog.tsx
@@ -23,7 +23,7 @@ export const InlineDialog = ({
     handleClose,
     'data-test-id': dataTestId = 'fondue-inlineDialog',
     darkUnderlay = false,
-    autoHeight = true,
+    autoHeight = false,
 }: InlineDialogProps) => {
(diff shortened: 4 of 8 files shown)


ryancarville (changes requested):
Hey @bojangles-m a few questions/comments here.

1. The `Popper` is a heavily reused component in `Fondue` and looks like this could be a `breakingChange` in expected behaviours.  E.g. The `InlineDialog` defaults to a locked scroll when this new `isStrageyFixed` `prop` defaulted to `false`.  Currently it should scroll freely. Can you please have a look into any side effects?

2. Can you please update the Storybook to classify this new prop `isStratageyFixed`?  The control for it is a raw `object` and should be a `boolean` switch in all the touched components? E.g. `Dialog` `InlineDialog` `Popover` `Popper` etc.  Seems like a `prop` that should be documented as it effect `UX` behaviours.

3. I know @imoutaharik is currently working on the `Popper` as part of a larger task with regard to the `Tooltip` component.  I have not seen any of her work thus far so I can not be 100% sure if this has no conflicts with her work.

bojangles-m:
@ryancarville at the moment it is just the property on the popper so it is not available as property on any other element because it is connected with the Modality. So if the opened dialog is modal then it should be fixed on the screen and I dont get it why InlineDialog can be Modal, does not make sense, because modal as aweb page element is a component that deactivates all other page content. 
What is your thought on that.

bojangles-m:
@ryancarville I see the problem that we have now with this solutions. What I would suggest is to make 'position or strategy' property to pass in dialog and can be controlled from outside. You have this two positions then 'fixed' & 'absolute', default is 'absolute'.
Not the best solution to expose more props, but it is necessary to have modal window not moving and as long as we allow InlineDialog to be modal i is hard to control it with Modality prop which is the way to go.

ryancarville:
@bojangles-m The issue is the `Popper` component as it is reused in the others. Considering this is a fix that is specific to the `Modal` (which I agree needs this fixed scrolling feature) would there be a approach to scope this just to that component?

imoutaharik:
@bojangles-m @ryancarville 
- I'm rethinking the Popper component, it works well for the InlineDialog and Dialog but seems too much for the Tooltip, which I tried to simplify as much as possible considering some usage in the web-app. It would be an overkill for it. So for now I may even release the Tooltip with its proper popper logic. But what I'm trying to improve here is how to pass the Trigger more than the popper-container itself. (Some usage are a tooltip wrapping a Button that triggers a Flyout)
- Modality is not about the UI element but is now a property for both InlineDialog and Dialog component. It's more about focusTrap and outside interaction. I documented the differences here : https://weare.frontify.com/document/1268#/overlays/modality-dark-underlay
- `Modal` component will eventually be deprecated once `Dialog` is ready.

bojangles-m:
@ilham, @ryancarville That is nicely written Use Cases:
﻿Modal dialogs are often used for critical messages, important alerts, or tasks that require the user's immediate attention. They force the user to address the dialog's content or complete a task before continuing with other parts of the application.
﻿And in that case the scroller should also be blocked.

bojangles-m:
> @bojangles-m The issue is the `Popper` component as it is reused in the others. Considering this is a fix that is specific to the `Modal` (which I agree needs this fixed scrolling feature) would there be a approach to scope this just to that component?

To which component? It has to be on Proper unless it wont work or hacking|(I dont prefer :))

imoutaharik:
@ryancarville @bojangles-m 
The issue I see with InlineDialog is that when on modal and when the content is quite big, you can't scroll down and can't see the full content of it but it's not an issue when autoHeight is on or maxHeight specified.
I would simply set the property `Autoheight` to `true` by default on the InlineDialog. 
WDYT?

ryancarville:
> @ryancarville @bojangles-m The issue I see with InlineDialog is that when on modal and when the content is quite big, you can't scroll down and can't see the full content of it but it's not an issue when autoHeight is on or maxHeight specified. I would simply set the property `Autoheight` to `true` by default on the InlineDialog. WDYT?

I agree @imoutaharik .  As long as the `InlineDialog` is still readable when open and has a large amount of content then I don't see any issues here.  I just tested the updates changes @bojangles-m and it works much better.  Please update the default value for `Autoheight` on the `InlineDialog` to `true` (which Ilham also mentioned) and I think we should be good.

ryancarville (approved):
Code LGTM 👍 (Sorry premature approval.  Storybook needs a few updates)

bojangles-m:
@ryancarville @imoutaharik changed auto height and updated storybook Popper that prop:strategy is exposed, others dont need it

ryancarville:
@bojangles-m Some storybook feedback.
`InlineDialog` still has the `Autoheight` default value as false & it is displaying the `strategy` prop.  Can you please update the boolean to `true` and hide the `strategy` prop?

bojangles-m:
> @bojangles-m Some storybook feedback.
> 
> `InlineDialog` still has the `Autoheight` default value as false & it is displaying the `strategy` prop. Can you please update the boolean to `true` and hide the `strategy` prop?
> 
> Also worth noting the `Dialog` and `Popover` SB's also has the `strategy` props exposed

Updated: Inline Dialog has no exposed strategy prop now.
The Dialog and Popper must have to be used.

PR #1622, opened 2023-10-17
fix(SegmentedControls): update element and multipliers to correct active border positions

ryancarville:
Active border not properly aligned in `Segmented Controls`.  This fix adjusts the calculation of its position by removing multiplier and initiated `translateX` value, moving the `motion.div` outside the `Fieldset` and updating the element stying.

Bug Repot: https://frontify.slack.com/archives/C02PSJTPA3D/p1693467213832059

ClickUp: https://app.clickup.com/t/8692mtd17

### DIFF
--- a/src/components/SegmentedControls/SegmentedControls.tsx
+++ b/src/components/SegmentedControls/SegmentedControls.tsx
@@ -164,17 +164,24 @@ export const SegmentedControls = ({
     const width = hugWidth ? '' : 'tw-w-full';
     const alignment = hugWidth ? 'tw-flex' : 'tw-grid tw-grid-flow-col tw-auto-cols-fr tw-justify-evenly';
 
+    const isSmallOrHugWidth = size === 'small' || hugWidth;
     const getSliderX = () => {
-        const isLastElement = selectedIndex === itemsRef.current.length - 1;
-        let translateX = isLastElement && !hugWidth ? 1 : 0;
+        let translateX = isSmallOrHugWidth ? -1 : 0;
         for (let i = 0; i < selectedIndex; i++) {
             translateX += itemsRef.current[i]?.clientWidth ?? 0;
         }
         return `${translateX}px`;
     };
 
+    const getSliderWidth = () => {
+        const isLastElement = selectedIndex === itemsRef.current.length - 1;
+        const baseValue = isSmallOrHugWidth ? 1 : 2;
+        const width = isLastElement ? baseValue + 1 : baseValue;
+
+        return `${(itemsRef.current[selectedIndex]?.clientWidth ?? 0) + width}px`;
+    };
     const sliderTranslation = getSliderX();
-    const sliderWidth = `${itemsRef.current[selectedIndex]?.clientWidth ?? 0}px`;
+    const sliderWidth = getSliderWidth();
 
     return (
         <div className="tw-flex">
--- a/src/components/SegmentedControls/SegmentedControls.tsx
+++ b/src/components/SegmentedControls/SegmentedControls.tsx
@@ -163,15 +163,15 @@ export const SegmentedControls = ({
     const alignment = hugWidth ? 'tw-flex' : 'tw-grid tw-grid-flow-col tw-auto-cols-fr tw-justify-evenly';
 
     const getSliderX = () => {
-        let translateX = 0;
+        let translateX = size === 'small' ? -1 : 0;
         for (let i = 0; i < selectedIndex; i++) {
             translateX += itemsRef.current[i]?.clientWidth ?? 0;
         }
         return `${translateX}px`;
     };
 
-    const sliderWidth = hugWidth ? `${itemsRef.current[selectedIndex]?.clientWidth}px` : `${100 / items.length}%`;
-    const sliderTranslation = hugWidth ? getSliderX() : `calc(${100 * selectedIndex}% - ${selectedIndex}px)`;
+    const sliderTranslation = getSliderX();
+    const sliderWidth = `${itemsRef.current[selectedIndex]?.clientWidth ?? 0}px`;
 
     return (
         <div className="tw-flex">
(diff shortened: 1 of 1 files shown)


bojangles-m (changes requested):
I checked on beta and apparently is still slight off.

ryancarville:
> I checked on beta and apparently is still slight off.

Thanks @bojangles-m will look at it again.  Difficult one as it must be dynamically sized on demand.

ryancarville:
> Before: Screenshot 2023-10-18 at 10 20 36
> 
> Now: Screenshot 2023-10-18 at 10 20 28

The `before` had this issue but only on the last child in the list.  But looks like this update shifted it to the front slightly.

ryancarville:
@bojangles-m the `hugWidth` calculation work well and make sense that all segments should use it since the active border should dynamically adjust to the selected size.  There does seem to be a slight ratio consideration between a `base` implementation (`size` `medium`) and a `size` `small` but by a factor of `1`.  I think this solution should hold now.

bojangles-m:
I have no idea have it should be, but here it is still off a bit. If that is ok, I am ok with it. Can you give me some reference how the end result should look like?
Screenshot 2023-10-18 at 12 05 15

ryancarville:
@bojangles-m where are you testing this?  For me on my local and on `preview` I am getting this: 
Screenshot 2023-10-18 at 12 07 49

bojangles-m:
> @bojangles-m where are you testing this? For me on my local and on `preview` I am getting this: Screenshot 2023-10-18 at 12 07 49

On macbook pro and i think as I see also on your is a bit off on the right side.

ryancarville:
@bojangles-m such a odd one.  I am not seeing it on my end. I just pushed a even simpler approach to this and lets hope it should solves the issue on your side.  See below I do not see the overlap issue.
Screenshot 2023-10-18 at 12 30 33

bojangles-m:
> @bojangles-m such a odd one. I am not seeing it on my end. I just pushed a even simpler approach to this and lets hope it solves the issue on your side. See below I do not see the overlap issue. Screenshot 2023-10-18 at 12 30 33

Uuuuu, nice, looks so much better :) but still some small tweaks 

Screenshot 2023-10-18 at 12 47 17
Screenshot 2023-10-18 at 12 47 23
Screenshot 2023-10-18 at 12 48 30

ryancarville:
I am not sure if it can get much better.  There seems to be a very subtle element width issue depending on the zoom factor of the screen. Any suggestions as I am at a lost.

bojangles-m:
> I am not sure if it can get much better. There seems to be a very subtle element width issue depending on the zoom factor of the screen. Any suggestions as I am at a loss. 🫠
> 
> Maybe a effect hook to re-render the elements to ensure the widths are updated? But this feels excessive IMO.

It looks mega already. I am just wondering why second last item does not work as all of them?

second last
Screenshot 2023-10-18 at 18 24 41

and one before
Screenshot 2023-10-18 at 18 24 49

bojangles-m (approved):

ryancarville:
@bojangles-m I totally agree that it is very puzzling that the second to last one on a `size` small variant has this output. But for the life of me I can not see anything causing it. I will take one more deep look tomorrow and if I can't uncover anything new I say we leave it. Overall is solves the reported issue and is better than before.

ryancarville:
> I agree it looks so much better then before, great job! What maybe would be better if this moving button would be outside of the filedset. Fieldset has border and the button is trapped inside of this border but it should be over the border so it can be easily positioned.

This approach does work to fix this edge case issue.  However, it presents a new issue which effects all transitions.  We have to set the `active` background color on the element itself which causes a not so attractive transition. E.g. The active element turns white -> then the slider border is positioned correctly.  It's very noticeable.

bojangles-m:
> @bojangles-m i just pushed this work to move the `motion.div` up on level. This morning I realised if we add a transition to the element that matched the duration of the `motion.div` we can align the two to make transition harmonious. Thanks for the suggestion!

Looks like better control over it. But the last item is in trouble, with border white something? pixel in front misisng.
Screenshot 2023-10-19 at 09 31 22

ryancarville:
> > @bojangles-m i just pushed this work to move the `motion.div` up on level. This morning I realised if we add a transition to the element that matched the duration of the `motion.div` we can align the two to make transition harmonious. Thanks for the suggestion!
> 
> Looks like better control over it. But the last item is in trouble, with border white something? pixel in front misisng. Screenshot 2023-10-19 at 09 31 22

What is your screen zoom percentage for this?  That seem to be the constant variable I see.   It alters the `width` of the elements marginally for some reason.

bojangles-m:
> > > @bojangles-m i just pushed this work to move the `motion.div` up on level. This morning I realised if we add a transition to the element that matched the duration of the `motion.div` we can align the two to make transition harmonious. Thanks for the suggestion!
> > 
> > 
> > Looks like better control over it. But the last item is in trouble, with border white something? pixel in front misisng.
> 
> What is your screen zoom percentage for this? That seem to be the constant variable I see. It alters the `width` of the elements marginally for some reason but not consistently. I get this:
> 
> Screenshot 2023-10-19 at 09 52 00

I am just pinching out on macbook. It is easier to see the problems which are also visible at normal size.
Do you using a normal monitor?

ryancarville:
> I am just pinching out on macbook. It is easier to see the problems which are also visible at normal size. Do you using a normal monitor?

I am on a external monitor and the macbook screen.  I get different renders/alignments depending on the zoom factor in the browser. It is odd.  But only on the last element in the list.

bojangles-m:
> > I am just pinching out on macbook. It is easier to see the problems which are also visible at normal size. Do you using a normal monitor?
> 
> I am on a external monitor and the macbook screen. I get different renders/alignments depending on the zoom factor in the browser. It is odd. But only on the last element in the list.

Looks great now :) thxn for patience.

bojangles-m (approved):

ryancarville:
> > > I am just pinching out on macbook. It is easier to see the problems which are also visible at normal size. Do you using a normal monitor?
> > 
> > 
> > I am on a external monitor and the macbook screen. I get different renders/alignments depending on the zoom factor in the browser. It is odd. But only on the last element in the list.
> 
> Looks great now :) thxn for patience.

Thanks for the excellent review and your patience!

PR #1623, opened 2023-10-17
fix(Accordion): update focus styling to include header

ryancarville:
Update to the `FOCUS_STYLE` properties on the `Accordion` to wrap all elements and not just collapsible content.  

Secondly, because the neighbouring `div` has a `button` child, the focused element needs to be set to `position: relative` to allow for the z-index to take priority and not be overlapped.

Bug Report: https://frontify.slack.com/archives/C02PSJTPA3D/p1689079198399909

ClickUp: https://app.clickup.com/t/862k4a2gm

### DIFF
--- a/src/components/Accordion/Accordion.tsx
+++ b/src/components/Accordion/Accordion.tsx
@@ -34,7 +34,7 @@ const AriaAccordionItem = ({
         <div
             key={item.key}
             className={merge([
-                isFocusVisible ? `${FOCUS_STYLE_NO_OFFSET} tw-relative` : '',
+                isFocusVisible ? `${FOCUS_STYLE_NO_OFFSET} tw-relative tw-rounded` : '',
                 divider && 'tw-divide-y tw-divide-black-10',
             ])}
         >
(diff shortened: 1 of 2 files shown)


bojangles-m:
First child
Screenshot 2023-10-18 at 10 27 04

Second child
Screenshot 2023-10-18 at 10 27 13

And then I found also this one!
Screenshot 2023-10-18 at 10 28 15

bojangles-m (changes requested):
Great take, I just found some inconsistency

ryancarville:
> Great take, I just found some inconsistency

I don't understand.  Could you elaborate on the inconsistencies?

bojangles-m:
> > Great take, I just found some inconsistency
> 
> I don't understand. Could you elaborate on the inconsistencies? For the last one, is it the overlap on the `switch`? If so this is a separate issue, but can be addressed by adding a larger default top `margin` on the `Collapsible` child.

yes it is the overlap. Ok if that is already known you can just ignore it.

bojangles-m:
The other thing is inconsistency with the frame around first child and other children if you look in the picture?

ryancarville:
We can add a minimal `marginTop` to the child to get it a little more space. Not a bad idea in general.  As for the `focusRing` it is a difficult one as there is a `2px` offset which overlaps the neighbours.  Checking with the designer to see if there are objections to using a ring with no offset.

ryancarville:
@bojangles-m added a no offset focus styling variant and margin top to content child.

bojangles-m:
> @bojangles-m added a no offset focus styling variant and margin top to content child.

If we do it like that and then the style compared to other components is not consistant?

ryancarville:
No really as the  `Accordion` has no border.  Only the "divider".  So technically it is not included in with the element.

bojangles-m:
Great job thnx

bojangles-m (approved):

PR #1624, opened 2023-10-18
[WIP]feat(Dropdown): open to opposite direction if there is no space

fulopdaniel:
The dropdown component always opened to the direction specified by the `position` prop, even if there was not enough space.

This PR proposes to automatically open to top if there is not at least 130px available space below the trigger.

Tests have not been modified yet.

### DIFF
--- a/src/components/Dropdown/Dropdown.tsx
+++ b/src/components/Dropdown/Dropdown.tsx
@@ -188,7 +188,7 @@ export const Dropdown = ({
         ['Bottom-End']: 'bottom-end',
     };
     const popperInstance = usePopper(triggerRef?.current, dropdownRef.current, {
-        placement: placementMap[`${adjustedPosition}-${alignment}`],
+        placement: placementMap[`${position}-${alignment}`],
         strategy: 'fixed',
         modifiers: [
             {
--- a/src/components/Dropdown/Dropdown.tsx
+++ b/src/components/Dropdown/Dropdown.tsx
@@ -155,7 +155,14 @@ export const Dropdown = ({
             return DropdownPosition.Bottom;
         }
     })();
-    const adjustedMaxHeight = adjustedPosition === DropdownPosition.Top ? maxHeight.toTop : maxHeight.toBottom;
+
+    const adjustedMaxHeight = (() => {
+        if (flip) {
+            return adjustedPosition === DropdownPosition.Top ? maxHeight.toTop : maxHeight.toBottom;
+        } else {
+            return position === DropdownPosition.Top ? maxHeight.toTop : maxHeight.toBottom;
+        }
+    })();
 
     const heightIsReady = !autoResize || maxHeight.toBottom || maxHeight.toTop;
 
--- a/src/components/Dropdown/Dropdown.tsx
+++ b/src/components/Dropdown/Dropdown.tsx
@@ -136,11 +136,11 @@ export const Dropdown = ({
     const spaceAvailableToTop = maxHeight.toTop ?? 0;
     const spaceAvailableToBottom = maxHeight.toBottom ?? 0;
 
-    const hasTopMoreSpaceThanBottom = spaceAvailableToTop > spaceAvailableToBottom;
-    const hasBottomMoreSpaceThanTop = !hasTopMoreSpaceThanBottom;
+    const isTopSpaceLarger = spaceAvailableToTop > spaceAvailableToBottom;
+    const isBottomSpaceLarger = !isTopSpaceLarger;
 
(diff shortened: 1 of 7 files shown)


src/components/Dropdown/Dropdown.tsx
```diff
@@ -131,7 +131,35 @@ export const Dropdown = ({
 
     const { maxHeight } = useDropdownAutoHeight(triggerRef, { isOpen, autoResize });
 
-    const heightIsReady = !autoResize || maxHeight !== DEFAULT_DROPDOWN_MAX_HEIGHT;
+    const minimumSpaceNeededToOpen = 130;
+
+    const spaceAvailableToTop = maxHeight.toTop ?? 0;
+    const spaceAvailableToBottom = maxHeight.toBottom ?? 0;
+
+    const hasTopMoreSpaceThanBottom = spaceAvailableToTop > spaceAvailableToBottom;
```
ryancarville, line 139: A cleaner name might be `isTopLarger`

src/components/Dropdown/Dropdown.tsx
```diff
@@ -131,7 +131,35 @@ export const Dropdown = ({
...
+    const hasBottomMoreSpaceThanTop = !hasTopMoreSpaceThanBottom;
```
ryancarville, line 140: Same here `isBottomLarger`

src/components/Dropdown/Dropdown.tsx
```diff
@@ -131,7 +131,35 @@ export const Dropdown = ({
...
+            } else {
+                return DropdownPosition.Top;
+            }
+        } else {
+            if (canOpenToTop || !canOpenToBottom) {
+                return DropdownPosition.Top;
+            } else {
+                return DropdownPosition.Bottom;
+            }
+        }
```
ryancarville, line 158:
```suggestion
        } else if (canOpenToTop || !canOpenToBottom) {
              return DropdownPosition.Top;
        } else {
              return DropdownPosition.Bottom;
        }
        
```

ryancarville: A single `if` statement shouldn't be the only condition in a `else`.

src/hooks/useDropdownAutoHeight.ts
```diff
@@ -2,22 +2,21 @@
...
-
 type DropdownAutoHeightProps = {
     isOpen: boolean;
     autoResize?: boolean;
     isDialog?: boolean;
 };
 
 export const getInnerOverlayHeight = (triggerRef: MutableRefObject<HTMLElement | null>) => {
-    let maxHeight = 'auto';
+    const maxHeight: { toTop?: number; toBottom?: number } = {};
```
ryancarville, line 12: We should think about having a default `auto` value to ensure no breakage?  At least for `toBottom` as this is used as the key returned value in most components.

src/hooks/useDropdownAutoHeight.ts
```diff
@@ -2,22 +2,21 @@
 
 import { MutableRefObject, useLayoutEffect, useState } from 'react';
```
ryancarville, line 4:
```suggestion
type MaxHeightObject = {
   toTop?: number;
   toBottom: number | 'auto';
}
```

src/hooks/useDropdownAutoHeight.ts
```diff
@@ -2,22 +2,21 @@
...
 
-export const DEFAULT_DROPDOWN_MAX_HEIGHT = 'auto';
-
 type DropdownAutoHeightProps = {
```
ryancarville, line 5:
```suggestion
type MaxHeightObject = {
   toTop?: number;
   toBottom: number | 'auto';
}

type DropdownAutoHeightProps = {
```

ryancarville, line 5:
```suggestion
type MaxHeightObject = {
   toTop?: number;
   toBottom: number | 'auto';
}

type DropdownAutoHeightProps = {
```

src/hooks/useDropdownAutoHeight.ts
ryancarville, line 4:
```suggestion
type MaxHeightObject = {
   toTop?: number;
   toBottom: number | 'auto';
}
```

ryancarville: This type i used in two places (potentially more in the future) so would be nice to make a `type` of it.

ryancarville:
Nice work!  Just some minor things to consider.  Also please make sure to update the tests and even add 1 or 2 to cover this new auto feature.

PR #1626, opened 2023-10-19
fix: change the type by which are children filtered

bojangles-m:
It turns out that our app build has is optimized for production which includes also minification. All of our component names and types have now been minified to something that is completely unpredictable. 
For that purposes I used displayName which we already set and will be the same also in production environment.

### DIFF
--- a/src/components/Popper/Popper.tsx
+++ b/src/components/Popper/Popper.tsx
@@ -23,6 +23,7 @@ export const Popper = ({
     verticalAlignment,
     strategy = 'absolute',
 }: PopperProps) => {
+    const arrayChildren = Children.toArray(children);
     const [referenceElement, setReferenceElement] = useState<HTMLDivElement | null>(null);
     const [popperElement, setPopperElement] = useState<HTMLDivElement | null>(null);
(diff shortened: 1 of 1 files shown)


ryancarville:
@SamuelAlev commented on the use of `displayName` in the `Tooltip` PR: 
> ...relying on displayName might be a problem, as unexpected issue might appear if we forget to set it in one component...

Thoughts?

bojangles-m:
> @SamuelAlev commented on the use of `displayName` in the `Tooltip` PR:
> 
> > ...relying on displayName might be a problem, as unexpected issue might appear if we forget to set it in one component...
> 
> Thoughts?

We can not rely on the type name. We have to make our own prop for that to name it or we have to have displayName.

bojangles-m:
@ryancarville my proposal is to create a ticket to update it in the future. We have the display name for both components Trigger and Content so for now it is good but we should avoid using type for that.

https://frontify.slack.com/archives/CQVQHGY2J/p1697707381221509

ryancarville (approved):
👍

PR #1628, opened 2023-10-19
fix:  positioning of the popper

bojangles-m:


ryancarville:
The `Dialog` seems to to be fixed, but can you check the `Popper` on Storybook?  It's does not seem to work as expected for me. First click nothing happens.  Second click it opens.  Third click it closes.  Fourth click is renders `300px` down the page.

bojangles-m:
> The `Dialog` seems to to be fixed, but can you check the `Popper` on Storybook? It's does not seem to work as expected for me. First click nothing happens. Second click it opens. Third click it closes. Fourth click is renders `300px` down the page.

Fixed and changed back to useState it looks like popper needs it.

ryancarville (approved):
Thanks for re-checking.  Great work! 👍

PR #1629, opened 2023-10-20
fix(Textarea): show focus ring when using keyboard

mnoleto:
Fix the focus ring display for Textarea component

- bug can be reproduced on the Libraries sidebar, where fields like description isn't showing the focus ring when focused using the keyboard


ryancarville:
Thanks @mnoleto for this contribution!   This is a known issue and we are currently working on a new `hook` for focus handling.  But this is a great solution in the mean time.

ryancarville (approved):

PR #1630, opened 2023-10-20
feat(LegacyTooltip): add withStrongBorder prop

fulopdaniel:
Needed for [Clickup Task](https://app.clickup.com/t/8692mvr65) and [PR](https://github.com/Frontify/web-app/pull/1662)

@todo:

[ x ] update tests
[ x ] add storybook story

### DIFF
--- a/src/components/LegacyTooltip/LegacyTooltip.tsx
+++ b/src/components/LegacyTooltip/LegacyTooltip.tsx
@@ -151,7 +151,7 @@ export const LegacyTooltip = ({
     children,
     position = TooltipPosition.Bottom,
     alignment = TooltipAlignment.Middle,
-    withStrongBorder,
+    withStrongBorder = false,
     withArrow,
     flip = true,
     triggerElement,
--- a/src/components/LegacyTooltip/LegacyTooltip.tsx
+++ b/src/components/LegacyTooltip/LegacyTooltip.tsx
@@ -409,7 +409,9 @@ export const LegacyTooltip = ({
                                 withArrow &&
                                     'tw-popper-arrow tw-z-[110000] tw-absolute tw-w-3 tw-h-3 tw-pointer-events-none before:tw-absolute before:tw-w-3 before:tw-h-3 before:tw-rotate-45 before:tw-border',
                                 withArrow && arrowStyling,
-                                withStrongBorder ? 'before:tw-border-line-strong' : 'before:tw-border-line',
+                                withStrongBorder
+                                    ? 'before:tw-border-line-strong'
+                                    : withArrow && 'before:tw-border-line',
                             ])}
                         />
                     </div>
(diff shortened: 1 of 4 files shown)


src/components/LegacyTooltip/LegacyTooltip.tsx
```diff
@@ -315,6 +317,7 @@ export const LegacyTooltip = ({
                     className={merge([
                         'tw-popper-container tw-inline-block tw-max-w-[200px] tw-dark tw-bg-base tw-rounded-md tw-shadow-mid tw-text-text tw-z-[120000]',
                         !isOpen && 'tw-opacity-0 tw-h-0 tw-w-0 tw-overflow-hidden',
+                        withStrongBorder && 'tw-border tw-border-line-strong',
```
SamCreasey, line 320: Since there's no other border option we could just use 'border' or 'withBorder'

SamCreasey (approved):

src/components/LegacyTooltip/LegacyTooltip.tsx
SamCreasey: As Discussed: 'withStrongBorder' is used because there is already a normal border on the arrow component

src/components/LegacyTooltip/LegacyTooltip.tsx
```diff
@@ -150,6 +151,7 @@ export const LegacyTooltip = ({
     children,
     position = TooltipPosition.Bottom,
     alignment = TooltipAlignment.Middle,
+    withStrongBorder,
```
ryancarville, line 154: Can we set a default value of `false` to be explicit and avoid any possible type corrosion qualifiers?  e.g. `withStrongBorder === false`

ryancarville (approved):
Two nitpick comments.  Otherwise LGTM 👍

src/components/LegacyTooltip/LegacyTooltip.tsx
```diff
@@ -404,8 +407,9 @@ export const LegacyTooltip = ({
                             style={popperInstance.styles.arrow}
                             className={merge([
                                 withArrow &&
-                                    'tw-popper-arrow tw-z-[110000] tw-absolute tw-w-3 tw-h-3 tw-pointer-events-none before:tw-absolute before:tw-w-3 before:tw-h-3 before:tw-rotate-45 before:tw-border before:tw-border-line',
+                                    'tw-popper-arrow tw-z-[110000] tw-absolute tw-w-3 tw-h-3 tw-pointer-events-none before:tw-absolute before:tw-w-3 before:tw-h-3 before:tw-rotate-45 before:tw-border',
```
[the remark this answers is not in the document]
fulopdaniel: The color of the border moved a couple lines down as now it can be changed based on the prop.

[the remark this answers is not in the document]
ryancarville: But the previous border only rendered when `withArrow` was `true`.  This will apply the border by default now.  Maybe keep the `before:tw-border-line` in the `withArrow` condition and the `withStrongBorder` check will overwrite it if applied?

PR #1635, opened 2023-10-25
fix(SegmentedControls): box and text colors base and when disabled

ryancarville:
Reported `a11y` issue.  Need to alter the box color and text for both enabled and disable states.


ryancarville:
Outdated with later PR

PR #1638, opened 2023-10-26
feat(new): Password Input Component

ryancarville:
New Password Input from series of all new type specific inputs.
[ClickUp Card](https://app.clickup.com/t/86782muva) 

## Definition of Done


src/components/PasswordInput/PasswordInput.tsx
```diff
@@ -0,0 +1,137 @@
...
+    onFocus,
+    onBlur,
+    onKeyDown,
+    'data-test-id': dataTestId = 'fondue-password-input',
+}: PasswordInputProps) => {
+    const inputElementRef = useRef<HTMLInputElement | null>(null);
+    const [isObfuscated, setIsObfuscated] = useState<boolean>(obfuscated);
+
+    const { isFocusVisible, focusProps } = useFocusRing({ within: true, isTextInput: true });
+    const { isFocusVisible: passwordButtonIsFocusVisible, focusProps: passwordButtonFocusProps } = useFocusRing();
```
imoutaharik, line 40: Are those 2 `isFocusVisible` const necessary? Are we not using Fondue buttons that already have it for those extra actions?

src/components/PasswordInput/PasswordInput.tsx
```diff
@@ -0,0 +1,137 @@
...
+            inputElementRef.current.value = '';
+            handleOnChange();
+        }
+    };
+
+    return (
+        <div
+            {...focusProps}
+            className={merge([
+                'tw-flex tw-items-center tw-justify-between tw-h-9 tw-gap-2 tw-px-3 tw-py-3 tw-transition tw-text-s tw-font-sans tw-relative tw-bg-white dark:tw-bg-transparent tw-border tw-rounded tw-line-strong',
```
imoutaharik, line 58: We shouldn't use `bg-white` here, only alias tokens -> `tw-bg-base`

src/components/PasswordInput/PasswordInput.tsx
```diff
@@ -0,0 +1,137 @@
...
+                readOnly={readOnly}
+                ref={inputElementRef}
+                required={required}
+                size={size}
+                type={isObfuscated ? InputTypes.Password : InputTypes.Text}
+                value={value}
+                className={merge([
+                    'tw-w-full tw-border-none tw-outline-none tw-bg-transparent tw-hide-input-arrows tw-text-sm',
+                    disabled || readOnly
+                        ? 'tw-text-black-40 tw-placeholder-black-30 dark:tw-text-black-30 dark:tw-placeholder-black-40 hover:tw-cursor-not-allowed'
```
imoutaharik, line 89: Same here with the alias tokens. Don't we have them specified? if not, also reach out to Designers.

src/components/PasswordInput/PasswordInput.tsx
```diff
@@ -0,0 +1,137 @@
...
+                        : 'tw-text-black tw-placeholder-black-60 dark:tw-text-white',
+                ])}
+            />
+            <span className="tw-flex tw-justify-between tw-items-center t-max-w-sm">
+                <button
+                    {...passwordButtonFocusProps}
```
imoutaharik, line 95: Any issue with the Icon Button on weak emphasis that you prefer not to use it here?

src/components/PasswordInput/PasswordInput.tsx
```diff
@@ -0,0 +1,137 @@
...
+                    type="button"
+                    className={merge([
+                        'tw-flex tw-items-center tw-justify-center tw-transition-colors tw-rounded tw-text-text-weak tw-p-1',
+                        disabled
+                            ? 'tw-cursor-default tw-text-black-40 hover:tw-cursor-not-allowed'
+                            : 'tw-text-black-60 hover:tw-text-black-100 hover:tw-rounded-sm hover:tw-bg-box-neutral hover:tw-text-box-neutral-inverse',
+                        passwordButtonIsFocusVisible && FOCUS_STYLE,
+                    ])}
+                >
+                    {isObfuscated ? <IconEye /> : <IconEyeOff />}
```
imoutaharik, line 110: If the size of the icon is not going to change, choose a sized one to import just like you did with IconCross.

src/components/PasswordInput/PasswordInput.tsx
ryancarville: We can, but then we run into the issue of the input ring being visible when the buttons are actually focused.

src/components/PasswordInput/PasswordInput.tsx
ryancarville: This could be me as I used existing input styles as a reference.  I will double check and loop in Designers if needed.

src/components/PasswordInput/PasswordInput.tsx
ryancarville: Just the focus state issues stated above in the first comment.

src/components/PasswordInput/PasswordInput.tsx
```diff
@@ -0,0 +1,121 @@
...
+                placeholder={placeholder}
+                id={useMemoizedId(propId)}
+                aria-label="Password Input"
+                autoComplete={autocomplete ? 'on' : 'off'}
+                onClick={() => inputElementRef.current?.focus()}
+                type={isObfuscated ? InputTypes.Password : InputTypes.Text}
+                className={merge([
+                    'tw-w-full tw-border-none tw-outline-none tw-bg-transparent tw-hide-input-arrows tw-text-sm',
+                    disabled || readOnly
+                        ? 'tw-text-text-disabled tw-placeholder-text-disabled dark:tw-text-black-30 dark:tw-placeholder-black-40 hover:tw-cursor-not-allowed'
```
imoutaharik, line 86: We need to sync with the Designers as it still doesn't look good on dark mode. 
We may also need to align on the alias tokens to be used here:
Screenshot 2023-11-07 at 16 01 58
Screenshot 2023-11-07 at 16 01 48

ryancarville: I will raise this in the channel

ryancarville: > We need to sync with the Designers as it still doesn't look good on dark mode. We may also need to align on the alias tokens to be used here:
> 
> Otherwise, it looks really good. Great work!! 🎉

@imoutaharik tokens have be confirmed and are updated.

imoutaharik: Thanks, It looks much better on dark mode, especially for the text color. But on dark mode, validation doesn't change the border color.

ryancarville: @imoutaharik good spot.  The validation map in general was missing `dark` theme presets.  This has been updated

imoutaharik (approved):

PR #1640, opened 2023-10-26
feat: passed callback prop

bojangles-m:
If the component using modal needs to execute sam function at the closing event then this is the callback that can be triggered at closing modal.

I used effect instead of this and it was triggered twice if we use callback is triggered once.

And also removed the function on the 30 line because it is not needed the setAction can be passed directly.

### DIFF
--- a/src/hooks/useToggleOverlay.ts
+++ b/src/hooks/useToggleOverlay.ts
@@ -4,19 +4,23 @@ import { useCallback, useEffect, useState } from 'react';
 
 export const useToggleOverlay = (
     initialState = false,
-    { isBlockingModal }: { isBlockingModal?: boolean } = { isBlockingModal: false },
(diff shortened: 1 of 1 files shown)


ryancarville:
Hey @bojangles-m.  I test this locally and it looks like it is effecting the `Dialog` positioning.  Also not sure if this is the reason, but the `Modal` and `Dropdown` (both use `useOverlay`) are also not rendering as they should in Storybook. 

Screenshot 2023-10-27 at 09 28 13
Screenshot 2023-10-27 at 09 29 13
Screenshot 2023-10-27 at 09 29 55

bojangles-m:
I guess only the last beta in this PR was missing.

bojangles-m:
@ryancarville could you please confirm that?

ryancarville (approved):
👍

PR #1641, opened 2023-10-31
fix(Tree): update deeper group test to scope check in seperate get call

ryancarville:
`Tree` still has one flaky test.  After discovery in Cypress about how to handle nested checks, they recommend using a new `get` call on the parent and then use the `within` callback to check the nested state.


anxobotana (approved):
LGTM

PR #1642, opened 2023-11-02
feat(Toast): add mighty dark theme and border style

fulopdaniel:


SamCreasey (approved):

src/components/Toast/types.tsx
```diff
@@ -29,3 +30,8 @@ export const toastStylesBackgroundColorsMap: Record<ToastStyle, string> = {
     [ToastStyle.Success]: 'tw-bg-box-positive-strong-pressed',
     [ToastStyle.Warning]: 'tw-bg-box-warning-strong',
 };
+
+export enum BorderStyle {
+    Default = 'Default',
+    Mighty = 'Mighty',
+}
```
ryancarville, line 37: Since this is specific to `Toast` maybe we should use a `map` `Record` like we do for the background?  Then implement is directly in the component?

ryancarville, line 37:
```suggestion
}

export const toastBorderStylesMap: Record<BorderStyle, string> = {
    [BorderStyle.Default]: 'tw-shadow-inner-line-x-strong',
    [BorderStyle.Mighty]: 'tw-shadow-inner-mighty',
};
```

src/components/Toast/types.tsx
```diff
@@ -8,6 +8,7 @@ export type ToastProps = {
     icon: ReactNode;
     animationDirection?: ToastAnimationDirection;
     children?: ReactNode;
+    variant?: ToastVariant;
```
imoutaharik, line 11: Thanks a lot! @fulopdaniel 😄 
It looks good to me, the only request I'm going to make before we merge this as I just aligned with designers...is renaming the property "variant" to "theme". We plan to use `variant` in other components for something different. 😉

imoutaharik (changes requested):
Thanks, just one commit away to ship this :)

imoutaharik (approved):
Thanks a lot!

PR #1651, opened 2023-11-06
fix(Dropdown): aria-hidden elements should not be a focusable element.

imoutaharik:
## Definition of Done

### DIFF
--- a/src/components/Trigger/Trigger.tsx
+++ b/src/components/Trigger/Trigger.tsx
@@ -3,7 +3,7 @@
 import { IconCaretDown, IconCross, IconExclamationMarkTriangle, IconTrashBin } from '@foundation/Icon/Generated';
 import { IconSize } from '@foundation/Icon/IconSize';
 import { useFocusRing } from '@react-aria/focus';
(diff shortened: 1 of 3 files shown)


src/components/Trigger/Trigger.tsx
```diff
@@ -113,14 +113,13 @@ export const Trigger = ({
                         <IconTrashBin size={IconSize.Size12} />
                     </button>
                 )}
-                <button
+                <div
                     {...buttonProps}
```
imoutaharik, line 117: I think we don't even need those buttonProps, do we?

ryancarville: Looks like it was added to handle caret positioning but maybe this is outdated now?

imoutaharik:
@ryancarville Yes, caret should not be interactive nor focusable.

ryancarville (approved):

imoutaharik:
> @ryancarville Yes, caret should not be interactive nor focusable.

But it blocks the input interaction so it's annoying.

imoutaharik:
sync with Lidija. Not needed after all. For now.

PR #1652, opened 2023-11-07
fix(TextInput): update invalid tailwind class for font size

ryancarville:
[Bug Report](https://frontify.slack.com/archives/C02PSJTPA3D/p1698223512499749)
[ClickUp](https://app.clickup.com/t/8692zy02m)

`TextInput` had odd behaviour when text was selected and `mouseDown` held while moving `pointer` up and down (2px shift).  Seems that the tailwind class for `font` size `small` was not correct (`tw-font-s`) and when updated to `tw-font-sm` shift was removed.

### DIFF
--- a/src/components/TextInput/TextInput.tsx
+++ b/src/components/TextInput/TextInput.tsx
@@ -199,7 +199,7 @@ export const TextInput = ({
             {...focusProps}
             {...props}
             className={merge([
-                'tw-flex tw-items-center tw-h-9 tw-gap-2 tw-px-3 tw-border tw-transition tw-rounded tw-text-s tw-font-sans tw-relative tw-bg-white dark:tw-bg-transparent',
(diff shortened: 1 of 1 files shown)


src/components/TextInput/TextInput.tsx
```diff
@@ -199,7 +199,7 @@ export const TextInput = ({
             {...focusProps}
             {...props}
             className={merge([
-                'tw-flex tw-items-center tw-h-9 tw-gap-2 tw-px-3 tw-border tw-transition tw-rounded tw-text-s tw-font-sans tw-relative tw-bg-white dark:tw-bg-transparent',
+                'tw-flex tw-items-center tw-h-9 tw-gap-2 tw-px-3 tw-border tw-transition tw-rounded tw-text-sm tw-font-sans tw-relative tw-bg-white dark:tw-bg-transparent',
```
imoutaharik, line 202: Indeed the class is not valid but you're replacing it with a Tailwind default. Could you try using our typography alias tokens -> `tw-text-body-small` instead.

imoutaharik (approved):

PR #1657, opened 2023-11-14
fix(Overlays): Improve Dialog responsiveness and Tooltip trigger

imoutaharik:
1- Adjust Tooltip Trigger to be able to accept a Fondue Button.
2- Improve Dialog responsiveness when screen is resized.


ryancarville (approved):
Nice update!

PR #1658, opened 2023-11-14
fix: export tooltip

SamCreasey:
Tooltip was not included in the package due to not being exported correctly.


imoutaharik (approved):
Why do I always forget that part! 🤦🏼‍♀️

PR #1659, opened 2023-11-14
feat(Toast): add prop for exit animation

fulopdaniel:


src/components/Toast/types.tsx
```diff
@@ -9,6 +9,7 @@ export type ToastProps = {
     animationDirection?: ToastAnimationDirection;
     children?: ReactNode;
     theme?: ToastTheme;
+    shouldAnimateExit?: boolean;
```
imoutaharik, line 12: Just leaving the comment here as well for the record:
we talked in the Team and we thought that a better name could be `enableExitAnimation` with a default to false
So usage would be `<Toast enableExitAnimation />`

imoutaharik (approved):

PR #1663, opened 2023-11-16
Feat(new): Textarea Component

ryancarville:
## Definition of Done

### DIFF
--- a/src/components/Textarea/Textarea.tsx
+++ b/src/components/Textarea/Textarea.tsx
@@ -1,0 +1,165 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+import { useDebounce } from '@hooks/useDebounce';
+import { useMemoizedId } from '@hooks/useMemoizedId';
+import { useFocusRing } from '@react-aria/focus';
+import { FOCUS_STYLE } from '@utilities/focusStyle';
+import { GetStatusIcon, InputStylesDarkTheme, InputStylesLightTheme } from '@utilities/input';
+import { merge } from '@utilities/merge';
+import { Validation, validationClassMap, validationTextClassMap } from '@utilities/validation';
+import { KeyboardEvent, ReactElement, TextareaHTMLAttributes, useEffect, useRef } from 'react';
+import TextareaAutosize, { TextareaAutosizeProps } from 'react-textarea-autosize';
+import { InputSharedBaseProps } from 'src/types/input';
+import { ON_SAVE_DELAY_IN_MS } from '..';
+
+export type TextareaProps = {
+    autosize?: boolean;
+    focusOnMount?: boolean;
+    height?: number;
+    resizable?: boolean;
+    selectable?: boolean;
+    defaultValue?: string;
+    value?: string;
+    onChange?: (value?: string) => void;
+    onInput?: (value?: string) => void;
+    onKeyDown?: (event: KeyboardEvent<HTMLTextAreaElement>) => void;
+    onEnterPressed?: (event: KeyboardEvent<HTMLTextAreaElement>) => void;
+} & Omit<InputSharedBaseProps, 'hugWidth'> &
+    TextareaHTMLAttributes<HTMLTextAreaElement> &
+    TextareaAutosizeProps;
+
+export const Textarea = ({
+    autocomplete,
+    autosize = false,
+    decorator,
+    disabled = false,
+    focusOnMount = false,
+    id: propId,
+    minRows,
+    maxRows,
+    placeholder,
+    readOnly,
+    resizable = false,
+    required,
+    selectable = false,
+    status = Validation.Default,
+    defaultValue,
+    value,
+    onBlur,
+    onChange,
+    onEnterPressed,
+    onKeyDown,
+    onFocus,
+    onInput,
+    'data-test-id': dataTestId = 'fondue-textarea',
+    ...props
+}: TextareaProps): ReactElement => {
+    const Component = autosize ? TextareaAutosize : 'textarea';
+
+    const textareaRef = useRef<HTMLTextAreaElement | null>(null);
+
+    const { isFocusVisible, focusProps } = useFocusRing({ isTextInput: true, within: true });
+
+    const handleOnChange = useDebounce(() => {
+        if (textareaRef.current) {
+            onChange?.(textareaRef.current.value);
+            onInput?.(textareaRef.current.value);
+        }
+    }, ON_SAVE_DELAY_IN_MS);
+
+    const handleOnKeyDown = (event: KeyboardEvent<HTMLTextAreaElement>) => {
+        if (onEnterPressed && event.key === 'Enter') {
+            onEnterPressed(event);
+        } else {
+            onKeyDown?.(event);
+        }
+    };
+
+    useEffect(() => {
+        focusOnMount && textareaRef.current?.focus();
+    }, [focusOnMount]);
+
+    const autosizeProps = { minRows, maxRows };
+
+    return (
+        <div className={'tw-relative'} {...focusProps}>
+            {decorator ? (
+                <div
+                    className={merge([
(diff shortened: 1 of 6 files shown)


src/utilities/input.tsx
```diff
@@ -13,6 +23,46 @@ type HelperTextProps = {
...
+    | 'width'
+    | 'height'
+    | 'element'
+    | 'focus'
+    | 'focusWithin'
+    | 'hover'
+    | 'disabled'
+    | 'readOnly';
+
+export const InputStylesLightTheme: Record<InputStyleGroup, string> = {
```
imoutaharik, line 37: @ryancarville I am wondering if it's really necessary to separate light from dark theme or if we just have `InputStyles` -> 
and each state base, hover, disabled etc... would include their base classes and dark prefixed ones.
Or what was your reasoning?

ryancarville: I was trying to extract the specific styling based on their type or theme for more granular control/exporting. But can combined if desired.

imoutaharik: @ryancarville In this case, I think it's better to combine as we don't have a theme property on the inputs (yet). We now have some themed components in Fondue and it could become confusing. Take the `Toast` for example, in light theme it uses `base` which changes to gray on darkMode. But on dark theme, it uses mighty which on darkMode changes to mighty-inverse.

ryancarville: Understood. I will combine them.

src/components/Textarea/Textarea.tsx
```diff
@@ -0,0 +1,176 @@
...
+                paddingClass = 'tw-pr-[2.5rem]';
+                break;
+            default:
+                paddingClass = 'tw-pr-[1rem]';
+        }
+        return paddingClass;
+    };
+
+    return (
+        <div className={'tw-relative'}>
```
imoutaharik, line 116: If it's going to be a string, no need to use curly brackets.

src/components/Textarea/Textarea.tsx
```diff
@@ -0,0 +1,176 @@
...
+                        'tw-absolute tw-top-[0.7rem] tw-left-[0.7rem] tw-z-10',
+                        disabled ? InputStyles.disabled : '',
+                        readOnly ? InputStyles.readOnly : '',
+                    ])}
+                    data-test-id={`${dataTestId}-decorator`}
+                >
+                    {decorator}
+                </div>
+            ) : null}
+            <span {...focusProps}>
```
imoutaharik, line 129: `span` element are usually not focusable. Would that be correct from an accessibility point of view?

src/utilities/input.tsx
```diff
@@ -3,8 +3,18 @@
 import { ReactElement } from 'react';
 import { Validation, validationTextClassMap } from './validation';
 import { merge } from './merge';
-import { IconCheckMark16, IconExclamationMarkTriangle16 } from '@foundation/Icon';
-import { LoadingCircle, LoadingCircleSize } from '@components/index';
+import { IconCheckMark16, IconCross16, IconExclamationMarkTriangle16, IconEye16, IconEyeOff16 } from '@foundation/Icon';
```
imoutaharik, line 6: Try to have the icon imports coming from the `Generated` folder

src/utilities/input.tsx
```diff
@@ -58,4 +95,68 @@ export const GetStatusIcon = (status: Validation, dataTestId: string): ReactElem
...
+    return (
+        <span className="tw-flex tw-items-center tw-justify-between tw-w-auto">
+            {obfuscated && (
+                <Button
+                    disabled={disabled}
+                    size={ButtonSize.Small}
+                    style={ButtonStyle.Default}
+                    emphasis={ButtonEmphasis.Weak}
+                    onClick={callbacks.password}
+                    aria-label={`${isObfuscated ? 'show' : 'hide'} text input`}
```
imoutaharik, line 118: At some point, we will have to discuss internally those aria-label values. We enforce English language here and our internal buttons should support multi-language but I wouldn't worry about that for this specific PR but definitely a topic to review.

src/components/Textarea/Textarea.tsx
ryancarville: The reason for this it there is a overwrite conflict with the `style` prop on the `textarea` when `react-aria` is applied directly.  So the `textarea` is wrapped in a `span` and the `useFocuseRing` hook is set to `within` to trigger when the `textarea` is focused within the `span`.  This has to be the structure as we also have faceable buttons `within` the same parent to the `textarea` needs its own wrapper.  Not sure of another approach here.

src/utilities/input.tsx
ryancarville: Yes this is a good (and big) point.

imoutaharik:
@ryancarville While testing on Storybook, I found a few issues:
1- When you switch the `autosize` control, the Text value get deleted.
2- on both `disabled` and `read-only`, action buttons like `clear btn` are still actionable.
3- resizable goes on vertical and horizontal directions but once resized, actions buttons keep floating in their original position:
Screenshot 2023-11-20 at 16 50 31
4- It would be nice to have a focusOnMount story
5- ExtraActions have double focus ring, probably one coming from the Tooltip itself (I may need to change that one myself on my current PR):
Screenshot 2023-11-20 at 16 52 03
Screenshot 2023-11-20 at 16 51 52

ryancarville:
@imoutaharik To add this functionality in SB we will need to remove the use of the `denouncer`.  Is this a valid trade-off?
> > 1- When you switch the `autosize` control, the Text value get deleted.

imoutaharik:
@ryancarville I Would definitely go for only Vertical resize.

imoutaharik:
Screenshot 2023-11-29 at 12 24 21

@ryancarville Should the default have that scrollbar? It wasn't there before.

src/utilities/input.tsx
```diff
@@ -43,7 +43,7 @@ type InputStyleGroup =
 export const InputStyles: Record<InputStyleGroup, string> = {
     base: 'tw-flex tw-items-center tw-justify-between tw-gap-2 tw-px-3 tw-transition tw-text-sm tw-font-sans tw-relative tw-bg-base tw-border tw-rounded tw-line-strong dark:tw-w-full dark:tw-flex dark:tw-items-center dark:tw-justify-between dark:tw-gap-2 dark:tw-px-3 dark:tw-transition dark:tw-text-sm dark:tw-font-sans dark:tw-relative dark:tw-bg-base dark:tw-border dark:tw-rounded dark:tw-line-strong',
     width: 'tw-w-full dark:tw-w-full',
-    height: 'tw-h-9 tw-min-h-[2.25rem] dark:tw-h-9 dark:tw-min-h-[2.25rem]',
```
imoutaharik: I think that it's not enough, I can still see the scrollbar. Maybe increase a tiny more.

Alos, I just noticed but I don't think you need to add the `dark:` prefix on everything but only on what's going to change, in this case only background and text color.

ryancarville: Hmm I did increase it to `2.35rem` odd is was not committed.  Will address.

imoutaharik: I still see the scroll bar:

Screenshot 2023-12-05 at 16 35 39

ryancarville: I am not sure how?! I am not getting it.  😕 
Screenshot 2023-12-05 at 16 38 23

imoutaharik: I use chrome. Am I the only one who sees it? I'm wondering.

imoutaharik: It's gone now. Weird. Ok, I'll approve the PR.

ryancarville: I have checked on Firefox, Chrome and Edge and I can not see it. But you can't be the only one. 😁 I will do some more digging and see what edge case this could be.

imoutaharik (approved):

PR #1665, opened 2023-11-21
fix(Dialog): set a fixed width of 90vw on mobile view.

imoutaharik:
Quick Fix for Dialog width issue on mobile view.

verticalAlignment and position will be discussed later.


ryancarville (approved):

src/components/Dialog/Dialog.stories.tsx
```diff
@@ -69,27 +69,31 @@ export default {
 
 const Template: StoryFn<DialogProps> = (args) => {
     const [isOpen, setIsOpen] = useToggleOverlay(false, { isBlockingModal: args.modality === Modality.BlockingModal });
+    const [showContent, setShowcontent] = useState(false);
```
ryancarville, line 72: `C` should be capitalised in `setShowContent`

imoutaharik: I saw it and just fixed.

PR #1668, opened 2023-11-22
fix: some Tabs improvements

imoutaharik:


src/components/Tabs/TabItem.tsx
```diff
@@ -30,7 +29,7 @@ export const TabItem = ({ active, disabled, children, id }: TabItemProps & { act
             ref={ref}
             id={`${id}-content`}
             aria-labelledby={id}
-            className={merge([!active || disabled ? 'tw-hidden' : '', !hasInteractiveElements && FOCUS_VISIBLE_STYLE])}
+            className={merge([!active || disabled ? 'tw-hidden' : '', 'focus:tw-ring-4 focus:tw-ring-blue'])}
```
ryancarville, line 32: Should we include `tw-rounded` here to ensure the ring style is consistent?

imoutaharik: Hello, just checked with Johanna, For this case as the component has a square shape, the focus shouldn't be rounded.

bojangles-m: Is it rounded option missing on the focus ring?

bojangles-m (approved):
LGTM: when is deployed I can test it modal then :)

PR #1675, opened 2023-12-11
refactor(InputExtraActions): allow any ReactElement as child

ryancarville:
## Definition of Done

### DIFF
--- a/src/types/input.ts
+++ b/src/types/input.ts
@@ -52,7 +52,8 @@ export type InputActionsProps = {
 };
 
 export type ExtraAction = {
-    tooltip: TooltipProps;
+    element?: ReactElement;
+    tooltip?: TooltipProps;
 };
 
 export type InputExtraActionsProps = {
--- a/src/types/input.ts
+++ b/src/types/input.ts
@@ -1,5 +1,6 @@
 /* (c) Copyright Frontify Ltd., all rights reserved. */
 
+import { TooltipProps } from '@components/Tooltip';
 import { Validation } from '@utilities/validation';
 import type { AriaAttributes, InputHTMLAttributes, ReactElement } from 'react';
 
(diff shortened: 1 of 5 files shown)


src/types/input.ts
```diff
@@ -51,9 +52,7 @@ export type InputActionsProps = {
 };
 
 export type ExtraAction = {
-    icon: ReactElement;
-    title: string;
-    callback: () => void;
+    tooltip: TooltipProps;
```
imoutaharik, line 55: Are we enforcing the use of Tooltip or just provide the option to add one if we want to?
Content on the Tooltip is required, I'm wondering what would be usage guidelines here => Actions buttons should always have a tooltip?

ryancarville: Currently the structure is that any `extraAction` is a `Tooltip` with any `child` element they wish to use.  This is what is reflected in the Figma file.  But can also modify this to allow for more flexibility.

imoutaharik: I see. Then that's fine but we could anyway double-check and get confirmation with the Designers.

ryancarville: OK will raise it at today's sync then can proceed from there.

ryancarville: @imoutaharik I have made the updates to allow a tooltip or element.  Interested in your thoughts on the implementation.

src/utilities/input.tsx
```diff
@@ -137,24 +145,35 @@ export const InputActions = ({
...
-        );
+        if (tooltip) {
+            return (
+                <Tooltip
+                    content={tooltip.content}
+                    key={`extra-action-tooltip-${id}`}
+                    disabled={tooltip.disabled ?? disabled}
+                    data-test-id={`${dataTestId}-extra-action`}
+                >
+                    {cloneElement(tooltip.children, { disabled: tooltip.disabled ?? disabled })}
```
imoutaharik, line 164: Is it the reason why we choose between Tooltip and Element? Like the whole Tooltip doesn't work if passed as ReactElement?

ryancarville: @imoutaharik I am update this to allow any `ReactElement`.

PR #1676, opened 2023-12-13
fix(SegmentedControls): allow for item in list to be disabled

ryancarville:
[ClickUp Report](https://app.clickup.com/t/8693avw8n)

Report: Request to `disable` single items in a group of `items`

Solution: 
1. Add a `disabled` prop to the `item` type.  
2. Update the `map` of the `item` component to reference `item.disabled || disabled` (default to the group disabled prop)
3. Update the `selectedIndex` method to by-pass the item if it is `disabled` (this is the key index that sets the `activeBorder` motion div)
4. Move the motion div outside the `fieldset` and update the styles accordingly.  The is required to allow the border to be the proper `z-index` else the disabled `background` will overlap the left border edge.  Added bonus is it is much more responsive to the correct sizing.


packages/fondue/src/components/SegmentedControls/SegmentedControls.tsx
```diff
@@ -142,15 +146,24 @@ export const SegmentedControls = ({
...
             />
         ));
     }, [items, id, disabled, radioGroupState, size]);
-    const selectedIndex = items.findIndex((item) => item.id === radioGroupState.selectedValue);
+
+    const selectedIndex = items.findIndex((item, index) => {
+        const isActiveItem = item.id === radioGroupState.selectedValue;
+        if (item.disabled && isActiveItem) {
+            const nextIndex = index + 1;
+            radioGroupState.setSelectedValue(items[nextIndex].id);
```
SamuelAlev, line 162: Goes out of bounds in case of the last item being `disabled`, also a disabled item should be able to be selected which makes the changes in this function not necessary 🤔

noahwaldner: yes i agree, this can be reverted to it's original function. i don't se why a disabled item couldn't be default

SamuelAlev (approved):

PR #1679, opened 2023-12-13
fix: focus trap

bojangles-m:
- updated focus trap
- removed close modal onBlur 
- removed content of non active tab item

### DIFF
--- a/src/components/Tabs/TabItem.tsx
+++ b/src/components/Tabs/TabItem.tsx
@@ -2,7 +2,6 @@
 
 import { ReactElement, ReactNode, useEffect, useState } from 'react';
 import { BadgeProps } from '@components/Badge';
-import { FOCUS_VISIBLE_STYLE } from '@utilities/focusStyle';
 import { merge } from '@utilities/merge';
 import { checkIfContainInteractiveElements } from '@utilities/elements';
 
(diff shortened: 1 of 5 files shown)


src/components/Tabs/TabItem.tsx
```diff
@@ -13,27 +14,28 @@ export type TabItemProps = {
...
-    return (
+    return active ? (
         <div
             role="tabpanel"
-            ref={ref}
+            ref={setRef}
             id={`${id}-content`}
             aria-labelledby={id}
-            className={merge([!active || disabled ? 'tw-hidden' : '', 'focus:tw-ring-4 focus:tw-ring-blue'])}
+            className={merge([!active || disabled ? 'tw-hidden' : '', !hasInteractiveElements && FOCUS_VISIBLE_STYLE])}
```
imoutaharik, line 34: Sorry, in the snippet I sent you, my branch wasn't updated yet with the latest changes I made on the Tab.
So the 'focus:tw-ring-4 focus:tw-ring-blue' needs to stay instead of the !hasInteractiveElements && FOCUS_VISIBLE_STYLE

bojangles-m: Updated. can you check, please?

imoutaharik: All good with the focus style, thanks.

imoutaharik (approved):
Thanks! Everything works as expected! 🎉 
It also works nice with the new usage 😄 . I just need to dedicate a bit of time to make it work when Portal is enabled to change it to an element listener instead of window.

bojangles-m:
> Thanks! Everything works as expected! 🎉 It also works nice with the new usage 😄 . I just need to dedicate a bit of time to make it work when Portal is enabled to change it to an element listener instead of window.

Thnx @imoutaharik for great collaboration 👍

PR #1687, opened 2024-01-18
fix: A11y Issues

noahwaldner:
- asset picker
	- Added a role to the element
- inline dialog trigger
	- Added a Aria Label to the Provided Trigger in the story
- menu
	- removed unnecessary role
- switch
	- added customizable aria label
- breadcrumbs
	- replaced "empty" element span by li
- multiselect
	- fix role

### DIFF
--- a/src/components/Menu/Menu.tsx
+++ b/src/components/Menu/Menu.tsx
@@ -98,7 +98,6 @@ export const Menu = ({
     return (
         <nav
             className={merge([CONTAINER_CLASSES, isMenuOpen ? 'tw-block' : 'tw-hidden'])}
-            role={isMenuOpen ? 'dialog' : ''}
             ref={setMenuContainerRef}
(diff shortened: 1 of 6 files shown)


src/components/Menu/Menu.tsx
```diff
@@ -98,7 +98,6 @@ export const Menu = ({
     return (
         <nav
             className={merge([CONTAINER_CLASSES, isMenuOpen ? 'tw-block' : 'tw-hidden'])}
-            role={isMenuOpen ? 'dialog' : ''}
```
SamuelAlev, line 101: Not so sure about this one, why the removal?

noahwaldner: The Nav element wrapping an element with the role "menu" doesn't necessarily need an explicit role specification. 

I addressed it though, and now added an explicit role "navigation"

SamuelAlev (approved):

PR #1688, opened 2024-01-18
fix(RTE): toolbar height is too big

fulopdaniel:
image

also on marketplace
Screenshot 2024-01-18 at 11 15 55

Somehow the `.tw-min-h-12` wasn't included in the bundle before


ragi96 (approved):
LGTM

PR #1689, opened 2024-01-19
fix: [a11y] Fix a1y issues in Fonude

noahwaldner:
Fixed Components:

- Number Input
- Text Input
- Text Area
- Multi Imput
- RTE -> Empty
- Checklist
- RadioList
- Segmented Controls

### DIFF
--- a/src/components/Checklist/Checklist.tsx
+++ b/src/components/Checklist/Checklist.tsx
@@ -83,7 +83,7 @@ export const Checklist = ({
     'data-test-id': dataTestId = 'checklist',
     ...props
 }: ChecklistProps) => {
-    const listContainerRef = useRef<HTMLUListElement | null>(null);
+    const listContainerRef = useRef<HTMLDivElement | null>(null);
     const state = useCheckboxGroupState({
         value: activeValues,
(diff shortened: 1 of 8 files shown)


src/components/Checklist/Checklist.tsx
```diff
@@ -98,7 +98,7 @@ export const Checklist = ({
     const columns = ('columns' in props && props.columns) || 1;
 
     return (
-        <ul
+        <div
```
SamuelAlev, line 101: Why replacing the correct semantics with `div`?

src/components/TextInput/TextInput.stories.tsx
```diff
@@ -7,6 +7,7 @@ import { TextInput, TextInputProps, TextInputType } from './TextInput';
 import { Validation } from '@utilities/validation';
 import { IconIcon } from '@foundation/Icon/Generated';
 import { action } from '@storybook/addon-actions';
+import { FormField } from '..';
```
SamuelAlev, line 10: Prefer direct import rather than `from '..';` or `from '.';`

SamuelAlev (approved):

PR #1700, opened 2024-02-01
fix: stop scrolling only on modal popups

bojangles-m:


noahwaldner (approved):

SamuelAlev:
@bojangles-m One test is failing (please don't fix it with just `{force:true}` in the test as it is a breaking behavior for users).
Screenshot 2024-02-01 at 13 54 40

.changeset/odd-countries-add.md
```diff
@@ -0,0 +1,5 @@
+---
+'@frontify/fondue': patch
+---
+
+Fix error when scrolling in modal
```
SamuelAlev, line 5: `fix: <description of what it solves>`

Try to be a bit more precise than "error"

SamuelAlev (approved):

PR #1706, opened 2024-02-07
fix: tooltip component styles

bojangles-m:


.changeset/stupid-poets-bathe.md
```diff
@@ -0,0 +1,5 @@
+---
+'@frontify/fondue': patch
+---
+
+Fixed tooltip component styles
```
SamuelAlev, line 5:
```suggestion
fix: don't use dark variant for the tooltip styles
```

SamuelAlev (approved):
Left a small comment

imoutaharik (commented):
Perfect! Thanks. Could you just remove the Header from the InContext Story -> 
in-context--with-header-and-footer-and-dark-theme.

As I said, this won't work anymore but no big deal. We should just remove from examples.

bojangles-m:
> Perfect! Thanks. Could you just remove the Header from the InContext Story -> in-context--with-header-and-footer-and-dark-theme.
> 
> As I said, this won't work anymore but no big deal. We should just remove from examples.

Removed

.changeset/stupid-poets-bathe.md
bojangles-m: Updated

imoutaharik (approved):

PR #1712, opened 2024-02-13
fix: tooltip - removed dark underlay and refactored click outside tooltip

bojangles-m:
Fixed tooltip:
- to not have dar underlay
- to only close when clicked outside of tooltip area


packages/fondue/src/utilities/dialogs/Overlay.tsx
```diff
@@ -146,7 +146,7 @@ export const Overlay = ({
                 </div>
             </Popper>
 
-            {(isMobile || modality !== Modality.NonModal) && open && (
```
imoutaharik, line 149: I see this is an issue for the Tooltip having a darkUnderlay in mobile view. But this breaks the InlineDialog.
Only dialog are affected by darkUnderlay (Popover and Tooltip not).
Dialog has it as a property.
InlineDialog has not dark underlay ever on large screen but get it on mobile view by default.

To fix it I wouldn't remove darkunderlay but add `isDialog`:

```
{isDialog && modality !== Modality.NonModal && open && (
                <Portal>
                    <div
                        data-test-id={`${dataTestId}-underlay`}
                        className={merge([
                            'tw-fixed tw-top-0 tw-left-0 tw-bottom-0 tw-right-0',
                            (darkUnderlay || isMobile) &&
                                'tw-transition-opacity tw-bg-box-neutral-mighty tw-opacity-50',
                        ])}
                        style={{ zIndex: Z_INDEX_MODAL_BACKDROP }}
                    />
                </Portal>
            )}

imoutaharik (approved):
LGTM.

bojangles-m:
@SamuelAlev can we merge this, because the Lint and Test error are not connected to this changes in PR?

PR #1715, opened 2024-02-19
WIP: allow for nested routes

noahwaldner:


docs/contentlayer.config.ts
```diff
@@ -18,6 +18,17 @@ const Component = defineDocumentType(() => ({
...
+                    return pathArray.at(-2);
+                } else {
+                    return pathArray.slice(-2).join('/');
+                }
+            },
+        },
+        parentFolder: {
             type: 'string',
             resolve: (doc) => doc._raw.sourceFileDir.toLocaleLowerCase().split('/').at(-1),
         },
```
SamuelAlev, line 34: `parentFolder` should be `.at(-2)`

`'/fondue/tag/content'.split('/')`:
- `.at(-1)` is `content`
- `.at(-2)` is `tag`

docs/src/pages/components/+Layout.tsx
```diff
@@ -2,22 +2,56 @@
...
                     <span className="tw-text-lg tw-font-bold">Components</span>
                     <ul className="tw-mt-4">
-                        {allComponents.map((component) => (
-                            <li key={component._id}>
-                                <a href={`/components/${component.route}`}>{component.title}</a>
-                            </li>
-                        ))}
+                        {Object.keys(componentTree).map((componentGroupKey) => {
+                            const componentGroup = componentTree[componentGroupKey] || [];
+                            if (componentGroup.length > 1) {
```
SamuelAlev, line 30: make it a component please

docs/src/pages/components/component/+route.ts
SamuelAlev, line 1: the folder structure is incorrect, please correct it and prefer it to `+route.ts`

SamuelAlev (changes requested):

docs/src/pages/components/component/+route.ts
noahwaldner: from what i saw in the docs you cannot do it with folder structure if some files are nested and others aren't

src/components/@componentID

can match .../components/button
but not .../components/tag/tagContent

Why do you preffer the directory routing over +route.ts?

SamuelAlev: > src/components/@componentID
> can match .../components/button
> but not .../components/tag/tagContent

This could possibly be solved with 2 routes:
`/src/components/@parentOrComponentId` => catches `/components/xyz`
`/src/components/@parentOrComponentId/@componentId` => catches `/components/foo/xyz`

> Why do you preffer the directory routing over +route.ts?
`+route.ts` isn't scoped to only the folder, meaning you will catch other routes too.

docs/src/pages/components/+Layout.tsx
noahwaldner: created

docs/src/pages/components/component/+route.ts
noahwaldner: I changed the route.ts file to export a route string that allows for us to catch nested routes. This takes care of all the scoping

docs/src/components/ComponentNavigation/ComponentNavigation.tsx
```diff
@@ -0,0 +1,52 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+import { type Component, allComponents } from '#contentlayer/generated';
+
+export const ComponentNavigation = () => {
+    type ComponentTree = {
+        [key: string]: Component[];
+    };
```
SamuelAlev, line 8: Can be moved outside

SamuelAlev (approved):
Left a comment but non blocking

PR #1732, opened 2024-03-05
fix(RichTextEditor): flyout modals not closing when clicked outside

SamCreasey:
Fixes an issue where toolbar flyouts would remain open when the user clicked a different spot inside the RichTextEditor.
Also reduces the number of calls and the complexity of the blur observer by only activating it when the editor is in focus.

Linked pr: https://github.com/Frontify/brand-sdk/pull/803

How to test with a guideline-block:
To be sure that you are using local code add a console.log() to the RTE component. (It will only show in edit mode since view mode uses the serialized html)

In @frontify/fondue
```
pnpm link --global
pnpm build
```

In @frontify/brand-sdk
```
rm -rf node_modules/.vite && rm -rf packages/guideline-blocks-settings/node_modules/.vite && pnpm build 
cd packages/guideline-blocks-settings
pnpm link --global
cd ../app-bridge
pnpm link --global

```

In @frontify/guideline-blocks
```
cd packages/text-block
pnpm i --frozen-lockfile
pnpm link --global @frontify/fondue @frontify/app-bridge @frontify/guideline-blocks-settings
rm -rf node_modules/.vite && pnpm serve

```

### DIFF
--- a/packages/fondue/src/components/RichTextEditor/BlurObserver.ts
+++ b/packages/fondue/src/components/RichTextEditor/BlurObserver.ts
@@ -2,7 +2,9 @@
 
 import { useEditorRef } from '@udecode/plate-core';
 import { floatingLinkActions, useFloatingLinkSelectors } from '@udecode/plate-link';
-import { useCallback, useEffect } from 'react';
+import { useCallback, useEffect, useState } from 'react';
+
+import { FLOATING_MODAL_SELECTOR } from './components';
 
 export const BlurObserver = ({
     hideExternalFloatingModals,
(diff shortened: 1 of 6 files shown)


packages/fondue/src/components/RichTextEditor/BlurObserver.ts
```diff
@@ -27,35 +30,58 @@ export const BlurObserver = ({
...
+
+        if (!richTextEditor) {
+            return;
+        }
+
+        const setupBlurObserver = () => {
+            setIsBlurObserverActive(true);
+        };
+
+        richTextEditor.addEventListener('focusin', setupBlurObserver);
```
fulopdaniel, line 79: Shall we unsubscribe on focusout? Otherwise click in => click out => click in will have 2 listeners for the same thing, right?

SamCreasey: click out should always trigger the unsubscribe from the hook above

SamCreasey: adding the focusout handler means repeating a lot of the conditional checks that are already prevent in the document event listener (check if its in a modal or the toolbar etc). It seems like it would make sense for it to unsubscribe itself

fulopdaniel: yeah my bad

fulopdaniel (approved):

PR #1735, opened 2024-03-06
fix(RTE): toolbar closes on safari

fulopdaniel:
https://app.clickup.com/t/2523021/TASK-10508


ragi96 (approved):
Thanks for fixing, works now in safari, also tested for firefox and chrome ✅

PR #1738, opened 2024-03-06
test: add component testing with cypress

floroz:
### Description

This PR introduces `cypress` component testing as integration test tool within `packages/components`.

- adds cypress in CI (with separate caching from node_modules)
- demo cypress test on `Tag` component
- adds `@testing-library/cypress` as selector library


.github/workflows/fondue-pull-requests.yaml
```diff
@@ -148,8 +148,19 @@ jobs:
...
                   cache: "pnpm"
 
+            - name: Cache Cypress binary
+              uses: actions/cache@v3
+              with:
+                  path: ~/.cache/Cypress
+                  key: cypress-13.6.6-${{ runner.os }}-cypress-${{ hashFiles('**/pnpm-lock.yaml')}}
+                  restore-keys: |
+                      cypress-13.6.6-${{ runner.os }}-cypress-
+
```
floroz, line 158: I have noticed the "old-component" jobs are building a cache for the `node_modules` that persists across workflow (and the cypress cache from those jobs is also handled within the same action).

Right now in this current workflow (`Components`) we are not doing that, so I think it's best to keep the cache and actions separated.

packages/components/package.json
```diff
@@ -23,7 +23,9 @@
         "typecheck": "tsc --noEmit",
         "test": "vitest run --coverage",
         "test:ui": "vitest --ui",
-        "test:watch": "vitest"
+        "test:watch": "vitest",
+        "test:integ": "cypress open --component",
+        "test:integ:ci": "cypress run --component --headless --browser electron"
```
floroz, line 28: for now is only electron, but we can (_should?_) expand to Chrome too in parallel

packages/components/package.json
```diff
@@ -34,22 +36,27 @@
...
         "eslint-plugin-notice": "^0.9.10",
         "happy-dom": "^13.3.8",
         "postcss": "^8.4.35",
         "postcss-import": "^16.0.1",
         "prettier": "^3.2.5",
         "react": "^18.2.0",
+        "react-dom": "^18.2.0",
         "tailwindcss": "^3.4.1",
         "typescript": "^5.3.3",
+        "vite": "^5.1.3",
```
floroz, line 59: @noahwaldner we might need a tool to synchronize the dev dependencies and tooling between packages to avoid battling different versions at the same point in time.

In the past, I've used https://www.npmjs.com/package/syncpack and was quite happy with it, but we can consider other alternatives.

packages/components/tsconfig.json
```diff
@@ -16,8 +16,9 @@
         "noUnusedLocals": true,
         "noUnusedParameters": true,
         "noFallthroughCasesInSwitch": true,
-        "baseUrl": "."
+        "baseUrl": ".",
+        "types": ["cypress", "@testing-library/cypress"]
     },
-    "include": ["src/**/*"],
+    "include": ["src/**/*", "cypress"],
```
floroz, line 22: I think we need to create a `tsconfig.lib.json` to feed into `dts` to distinguish between the types we need to build and ship, and those who are now overlapping with Cypress.

Thoughts?

packages/fondue/package.json
```diff
@@ -103,8 +103,8 @@
         "chalk": "^5.3.0",
         "concat": "^1.0.3",
         "cross-env": "^7.0.3",
-        "cypress": "^13.6.4",
-        "cypress-real-events": "^1.12.0",
```
floroz, line 107: I am pinning these versions as we want to share the same cache. We are not exactly doing this yet in the actions, but I didn't want to increase the scope and make changes into the existing `old-component` workflow

.github/workflows/fondue-pull-requests.yaml
```diff
@@ -112,11 +112,18 @@ jobs:
             - name: Setup pnpm cache
               uses: actions/cache@v3
               with:
-                  path: |
-                      ${{ steps.pnpm-cache.outputs.pnpm_cache_dir }}
-                      ~/.cache/Cypress
-                  key: pnpm-with-cypress-${{ runner.os }}-${{ hashFiles('pnpm-lock.yaml') }}
-                  restore-keys: pnpm-with-cypress-${{ runner.os }}-${{ hashFiles('pnpm-lock.yaml') }}
```
floroz, line 119: splitting the 2 cache operations separately to be more explicit and matching the new component workflow, where we are not caching the entire node_modules directory.

packages/components/tsconfig.json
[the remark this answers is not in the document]
floroz: What do you see as the main limitation of this approach?

As far as I could tell, the `tsconfig.json` under `/components` is not used to emit declarations files so this change is only to enhance VSCode.

We could have a _cleaner_ approach by splitting a `tsconfig.json` for the VSCode TS Server and a `tsconfig.lib.json`, but it felt a bit overkill for this PR (without any real benefits).

But I am happy to hear your thoughts!

Regarding Cypress as a tool, and the alternative you suggested: I used Playwright in the past and it's an exceptional tool for e2e testing.

However, [its Component Testing runner is still in the experimental phase](https://playwright.dev/docs/test-components). 
I don't have any experience with Vite Browser Mode, but besides being also in Experimental mode, seems quite cutting-edge technology.

Another big point in favour of Cypress is that is already used within the old component, so from a cognitive perspective would make migrating some stuff easier, and potentially help move the tests quicker with a minimal footprint. So there's quite a bit of gain in terms of effort/time by keeping the same tool.

I am generally conservative when it comes to picking the main tooling for a project like ours, but I am happy to discuss it with you guys and consider alternatives :) cc @SamuelAlev @noahwaldner

packages/components/package.json
noahwaldner: let's talk about thhis!

PR #1744, opened 2024-03-07
chore: bundle fondue-components with fondue

noahwaldner:


.changeset/little-moons-play.md
```diff
@@ -0,0 +1,5 @@
+---
+"@frontify/fondue": patch
+---
+
+bundle fondue-components
```
floroz, line 5:
```suggestion
build: bundle fondue-components
```

@noahwaldner @SamuelAlev should we consider something like committizen to automate this check on the changeset commit? Do we use conventional commits?

packages/fondue/vite.config.ts
```diff
@@ -9,7 +9,8 @@ import tsConfigPaths from 'vite-tsconfig-paths';
 import { dependencies as dependenciesMap, peerDependencies as peerDependenciesMap } from './package.json';
 
 const peerDependencies = Object.keys(peerDependenciesMap);
-const dependencies = Object.keys(dependenciesMap);
+const dependencies = Object.keys(dependenciesMap).filter((dependency) => !dependency.startsWith('@frontify/fondue'));
```
floroz, line 12:
```suggestion
// TODO: review this approach to exclude all deps from the final bundles.
// The filtering is used for an hotfix and is a temporary workaround until we have refined how to improve and stabilize the deps building
const dependencies = Object.keys(dependenciesMap).filter((dependency) => !dependency.startsWith('@frontify/fondue'));
```

floroz (approved):

floroz (approved):

PR #1745, opened 2024-03-07
fix: cleanup segmented controls

SamCreasey:
This pr removes the jank visible in the Segmented controls when controls change width, for example, when the content reaches a height which requires a scrollbar. 
Eg: 
Screenshot 2024-03-07 at 20 16 14

To test with changing width: 
```
const SegmentedControlsTemplate: StoryFn<SegmentedControlsProps> = (args: SegmentedControlsProps) => {
    const [activeItemId, setActiveItemId] = useState(args.activeItemId || args.items[0].id);
    const [controlledWidth, setControlledWidth] = useState(false);

    return (
        <div className={controlledWidth ? 'tw-w-[500px]' : ''}>
            <SegmentedControls
                {...args}
                key={String(args.hugWidth)}
                activeItemId={activeItemId}
                onChange={setActiveItemId}
            />
            <div>
                <button onClick={() => setControlledWidth((controlled) => !controlled)}>Click</button>
            </div>
        </div>
    );
};
```


fulopdaniel (approved):
Works nicely 👍

SamuelAlev (approved):
Looks solid, thanks for taking care!

PR #1751, opened 2024-03-13
feat: Filterable MultiSelect

selimovicz:
## Definition of Done


src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -50,6 +50,8 @@ export type MultiSelectProps = {
     flip?: boolean;
     emphasis?: TriggerEmphasis;
     enablePortal?: boolean;
+    filterable?: boolean;
```
mnoleto, line 53: With this type you can end up with a filter without a label or a label without a filter. It is always better to make it impossible to create a mess. I think it is enough to have the `filterLabel`. If the label is provided it means that is filterable.
If you want to still have a boolean to control the display of the filter:
`filter?: {.enabled: true; label: string }`

src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -130,7 +134,7 @@ export const MultiSelect = ({
     };
 
     const handleSpacebarToggle = (e: KeyboardEvent<HTMLDivElement>) => {
-        if (e.code === 'Space') {
+        if (e.code === 'Space' && document.activeElement !== filterInputRef.current) {
```
mnoleto, line 137: Pls, change the parameter to `event`

src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -142,6 +146,21 @@ export const MultiSelect = ({
         return TagType.SelectedWithFocus;
     };
 
+    const handleFilterChange = (e: ChangeEvent<HTMLInputElement>) => {
```
mnoleto, line 149: Same here.

src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -142,6 +146,21 @@ export const MultiSelect = ({
...
+        setCheckboxes(
+            items.reduce((acc: Item[], item) => {
+                if (
+                    item.isCategory ||
+                    item.isDivider ||
+                    item.value.toLowerCase().includes(e.currentTarget.value.toLowerCase())
+                ) {
+                    acc.push({ ...item, label: item.value });
```
mnoleto, line 157: Pls, avoid argument mutation.
`return [...acc, { ...item, label: item.value }]`

mnoleto (approved):

packages/fondue/src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -239,6 +268,15 @@ export const MultiSelect = ({
 
                             {activeItemKeys.length === 0 && placeholder && <Text color="weak">{placeholder}</Text>}
                         </div>
+                        {filter.enabled && (
+                            <input
+                                data-test-id="filter-input"
+                                ref={filterInputRef}
+                                className="tw-outline-none tw-bg-transparent tw-placeholder-black-60 tw-text-s"
```
floroz, line 275: Alignment of the text input should follow the chips and wrap on a new row, but it's centered in a weird way right now: 
Screenshot 2024-03-14 at 09 59 25

floroz (changes requested):

packages/fondue/src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -239,6 +268,15 @@ export const MultiSelect = ({
...
+                                placeholder={activeItemKeys.length === 0 ? placeholder : filter.label}
+                                onChange={handleFilterChange}
```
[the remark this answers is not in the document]
selimovicz: Yep, that was the plan. That's why I left unchecked the list item in the description saying:  User interface and experience have been verified by a designer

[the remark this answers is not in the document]
urietberger: Hey @floroz, I took over this task from Zeid as he's off this week. After talking to our designer, I applied a change where we no longer use the placeholder within the newly added input, which removes the redundant placeholder. However, it's still possible to have 2 placeholders by defining both, the `placeholder` and the `filter.enabled` properties. I'd say that's fine and is up to the user of the component whether or not they want to define. Agreed?

packages/fondue/src/components/MultiSelect/MultiSelect.tsx
urietberger: Fully agreed. Checked it out with our designer and changed the input to be rendered inline after the active items.

floroz:
Thanks @urietberger.

A few thoughts from my side:

My main concern is that this change transforms the component into becoming an `Autocomplete`/`Combobox` without the UX context necessary for the end user.

For example these two instances, one can contain an input to filter and the other doesn't. But they both use a placeholder.
The end user has no indication of something that can be filterable or not

- Non filterable
image

- Filterable
image

Furthermore, we are missing empty state for the filtering:

image


I think we need some design input to correctly plan an Autocomplete/Combobox component.

I'd like to get some feedback from @noahwaldner and @SamuelAlev

noahwaldner:
I agree with @floroz on this one that this needs some thorough UX evaluation before we make this change as it changes the type of place where this component is used.

I will set up meetings to talk about this cc: @urietberger @floroz @selimovicz

selimovicz:
Closing this one in favour of [this](https://github.com/Frontify/fondue/pull/1769).

PR #1752, opened 2024-03-13
fix: replaced mouse events with pointer events

bojangles-m:
In tooltip replaced mouse events with pointer events to have control over tooltip and dont show it on touch devices if the type of it is touch or pen. Only show it when pointer type is mouse.

### DIFF
--- a/.changeset/lucky-rivers-doubt.md
+++ b/.changeset/lucky-rivers-doubt.md
@@ -1,5 +1,5 @@
 ---
-"@frontify/fondue": major
+"@frontify/fondue": minor
 ---
 
-fix(Tooltip): Replaced onmouse events with pointer events, to distinguish between devices.
+refactor(Tooltip): replaced `onMouse` events with `onPointer` events to distinguish between devices.
--- a/.changeset/lucky-rivers-doubt.md
+++ b/.changeset/lucky-rivers-doubt.md
@@ -1,0 +1,5 @@
+---
+"@frontify/fondue": major
+---
+
+fix(Tooltip): Replaced onmouse events with pointer events, to distinguish between devices.
--- a/packages/fondue/src/components/FormField/FormField.cy.tsx
+++ b/packages/fondue/src/components/FormField/FormField.cy.tsx
@@ -147,14 +147,5 @@ describe('Form Field Component', () => {
             });
             cy.get(FORM_FIELD_TOOLTIP_TWO).should('exist');
         });
-
-        it('should not open 1st tooltip on touch event', () => {
-            cy.mount(COMPONENT_BASE_WITH_TOOLTIPS);
-            cy.get(FORM_FIELD_TOOLTIP_ONE_ICON).trigger('pointerover', {
-                eventConstructor: 'MouseEvent',
-                pointerType: 'touch',
-            });
-            cy.get(FORM_FIELD_TOOLTIP_ONE).should('not.exist');
-        });
     });
 });
--- a/packages/fondue/src/components/FormField/FormField.cy.tsx
+++ b/packages/fondue/src/components/FormField/FormField.cy.tsx
@@ -117,27 +117,44 @@ describe('Form Field Component', () => {
         cy.get(FORM_FIELD_TEXT_INPUT).focused();
     });
 
-    it('should render tooltip', () => {
-        cy.mount(COMPONENT_BASE_WITH_TOOLTIPS);
-        cy.get(FORM_FIELD_TOOLTIP_ONE_ICON).should('exist');
-    });
+    describe('Tooltips', () => {
+        it('should render', () => {
+            cy.mount(COMPONENT_BASE_WITH_TOOLTIPS);
+            cy.get(FORM_FIELD_TOOLTIP_ONE_ICON).should('exist');
+        });
 
-    it('should render multiple tooltips', () => {
-        cy.mount(COMPONENT_BASE_WITH_TOOLTIPS);
-        cy.get(FORM_FIELD_LABEL).children().should('have.length', 2);
-        cy.get(FORM_FIELD_TOOLTIP_ONE_ICON).should('exist');
-        cy.get(FORM_FIELD_TOOLTIP_TWO_ICON).should('exist');
-    });
+        it('should render multiple', () => {
+            cy.mount(COMPONENT_BASE_WITH_TOOLTIPS);
+            cy.get(FORM_FIELD_LABEL).children().should('have.length', 2);
+            cy.get(FORM_FIELD_TOOLTIP_ONE_ICON).should('exist');
+            cy.get(FORM_FIELD_TOOLTIP_TWO_ICON).should('exist');
+        });
 
-    it('should open 1st tooltip dialog on hover', () => {
-        cy.mount(COMPONENT_BASE_WITH_TOOLTIPS);
-        cy.get(FORM_FIELD_TOOLTIP_ONE_ICON).trigger('mouseover');
-        cy.get(FORM_FIELD_TOOLTIP_ONE).should('exist');
-    });
+        it('should open 1st tooltip dialog on hover', () => {
+            cy.mount(COMPONENT_BASE_WITH_TOOLTIPS);
+            cy.get(FORM_FIELD_TOOLTIP_ONE_ICON).trigger('pointerover', {
+                eventConstructor: 'MouseEvent',
+                pointerType: 'mouse',
+            });
+            cy.get(FORM_FIELD_TOOLTIP_ONE).should('exist');
+        });
 
-    it('should open 2nd tooltip dialog on hover', () => {
-        cy.mount(COMPONENT_BASE_WITH_TOOLTIPS);
-        cy.get(FORM_FIELD_TOOLTIP_TWO_ICON).trigger('mouseover');
-        cy.get(FORM_FIELD_TOOLTIP_TWO).should('exist');
+        it('should open 2nd tooltip dialog on hover', () => {
+            cy.mount(COMPONENT_BASE_WITH_TOOLTIPS);
+            cy.get(FORM_FIELD_TOOLTIP_TWO_ICON).trigger('pointerover', {
+                eventConstructor: 'MouseEvent',
+                pointerType: 'mouse',
+            });
+            cy.get(FORM_FIELD_TOOLTIP_TWO).should('exist');
+        });
+
+        it('should not open 1st tooltip on touch event', () => {
+            cy.mount(COMPONENT_BASE_WITH_TOOLTIPS);
+            cy.get(FORM_FIELD_TOOLTIP_ONE_ICON).trigger('pointerover', {
+                eventConstructor: 'MouseEvent',
+                pointerType: 'touch',
+            });
+            cy.get(FORM_FIELD_TOOLTIP_ONE).should('not.exist');
+        });
     });
 });
--- a/packages/fondue/src/components/Tooltip/Tooltip.tsx
+++ b/packages/fondue/src/components/Tooltip/Tooltip.tsx
@@ -36,14 +36,14 @@ const formatTooltipText = (text: string) => {
     return text.split(lineBreakRegex).join('\n');
 };
 
-let timeoutId: NodeJS.Timeout | undefined;
+let timeoutId: number | undefined;
 const handleTimeout = (callback: () => void, delay: number) => {
     if (timeoutId) {
-        clearTimeout(timeoutId);
+        window.clearTimeout(timeoutId);
     }
 
     if (delay) {
-        timeoutId = setTimeout(callback, delay);
+        timeoutId = window.setTimeout(callback, delay);
     } else {
         callback();
     }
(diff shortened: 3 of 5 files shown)


packages/fondue/src/components/Tooltip/Tooltip.tsx
```diff
@@ -36,22 +36,21 @@ const formatTooltipText = (text: string) => {
...
 };
 
-type TimeoutRef = {
-    current: NodeJS.Timeout | null;
-};
-
-const handleTimeout = (callback: () => void, delay: number, timeoutRef: TimeoutRef) => {
-    if (timeoutRef.current) {
-        clearTimeout(timeoutRef.current);
+let timeoutId: NodeJS.Timeout | undefined;
```
SamuelAlev, line 39: Please put it in the component, having it outside is a side-effect.

SamuelAlev, line 39: Also 
```suggestion
let timeoutId: number | undefined;
```
and you need to use `window.setTimeout` instead of just `timeout`

packages/fondue/src/components/Tooltip/Tooltip.tsx
```diff
@@ -36,22 +36,21 @@ const formatTooltipText = (text: string) => {
...
+const handleTimeout = (callback: () => void, delay: number) => {
+    if (timeoutId) {
+        clearTimeout(timeoutId);
     }
 
     if (delay) {
-        timeoutRef.current = setTimeout(callback, delay);
+        timeoutId = setTimeout(callback, delay);
```
SamuelAlev, line 46:
```suggestion
        timeoutId = window.setTimeout(callback, delay);
```

packages/fondue/src/components/FormField/FormField.cy.tsx
```diff
@@ -117,27 +117,44 @@ describe('Form Field Component', () => {
...
+        it('should open 2nd tooltip dialog on hover', () => {
+            cy.mount(COMPONENT_BASE_WITH_TOOLTIPS);
+            cy.get(FORM_FIELD_TOOLTIP_TWO_ICON).trigger('pointerover', {
+                eventConstructor: 'MouseEvent',
+                pointerType: 'mouse',
+            });
+            cy.get(FORM_FIELD_TOOLTIP_TWO).should('exist');
+        });
+
+        it('should not open 1st tooltip on touch event', () => {
```
SamuelAlev, line 151: Please remove these tests and add them in `packages/fondue/src/components/Tooltip/Tooltip.cy.tsx`

.changeset/lucky-rivers-doubt.md
```diff
@@ -0,0 +1,5 @@
+---
+"@frontify/fondue": major
+---
+
+fix(Tooltip): Replaced onmouse events with pointer events, to distinguish between devices.
```
SamuelAlev, line 5:
```suggestion
refactor(Tooltip): replaced `onMouse` events with `onPointer` events to distinguish between devices.
```

.changeset/lucky-rivers-doubt.md
```diff
@@ -0,0 +1,5 @@
+---
+"@frontify/fondue": major
```
SamuelAlev, line 2:
```suggestion
"@frontify/fondue": minor
```
Is it really a breaking change?

packages/fondue/src/components/Tooltip/Tooltip.tsx
bojangles-m: I didn't put the function back I passed the ref as prop so that the function can leave outside of the main component. 
Great catch for the side effect, thx

.changeset/lucky-rivers-doubt.md
bojangles-m: Updated

.changeset/lucky-rivers-doubt.md
bojangles-m: Updated

packages/fondue/src/components/Tooltip/Tooltip.tsx
bojangles-m: Updated.

packages/fondue/src/components/Tooltip/Tooltip.tsx
bojangles-m: Changed

packages/fondue/src/components/FormField/FormField.cy.tsx
bojangles-m: I will leave the previous test inside, because they cover the Tooltips for this component and will add additional test to the Tooltip.

bojangles-m: Updated

packages/fondue/src/components/Tooltip/Tooltip.tsx
bojangles-m: Updated

packages/fondue/src/components/Tooltip/Tooltip.tsx
```diff
@@ -37,21 +37,22 @@ const formatTooltipText = (text: string) => {
 };
 
 type TimeoutRef = {
-    current: NodeJS.Timeout | null;
+    current?: number;
```
floroz, line 40: could you help me understand this change? Isn't the ref holding a reference to the timer?

packages/fondue/src/components/Tooltip/Tooltip.tsx
```diff
@@ -37,21 +37,22 @@ const formatTooltipText = (text: string) => {
...
 };
-
 const handleTimeout = (callback: () => void, delay: number, timeoutRef: TimeoutRef) => {
```
floroz, line 42: I know this is not part of your changes, but since we are refactoring this component, could we consider moving the `handleTimeout` within the component? 

This removes the need to create the `TimeoutRef` type and provides better cohesion with the side-effect operation performed by the function.

packages/fondue/src/components/Tooltip/Tooltip.tsx
```diff
@@ -71,39 +72,48 @@ export const Tooltip = ({
...
+            leaveDelay,
+            timeoutRef,
+        );
+    }, [leaveDelay]);
 
     const { dismissibleElementRef } = useClickOutside<HTMLDivElement>(handleHideTooltip, []);
 
     const focusAndMouseAttributes = {
         onBlur: handleHideTooltip,
         onFocus: handleShowTooltip,
```
floroz, line 100: right now a PointerDevice can still activate the tooltip upon click. This is probably a UX-related question on whether we should disable it altogether or not? cc @noahwaldner @SamuelAlev 

It could be a controlled input since we have a `openOnMount` (which should be renamed since there is no tied logic to the lifecycle and it's just a prop passed by the consumer).

floroz (commented):
Hi @bojangles-m 👋🏻 ! Just adding a few thoughts from my side

packages/fondue/src/components/Tooltip/Tooltip.tsx
bojangles-m: yes it is, why?

packages/fondue/src/components/Tooltip/Tooltip.tsx
bojangles-m: I would like to make the main function readable and that is the main reason why it is out. The main ref `timeoutRef` is inside of the main function which prevents the side effects.

packages/fondue/src/components/Tooltip/Tooltip.tsx
bojangles-m: Yes I would agree, just the click only happens on a mouse event, the click on mobile will not trigger it.
Its a good question if we want to trigger tooltip on a click or only on a hover.
If there is a  discussion how this tooltip should behave @floroz maybe we could discuss it and update it in another PR. What do you think?

packages/fondue/src/components/Tooltip/Tooltip.tsx
floroz: Unfortunately, side-effects are not prevented since we are mutating the value of the `ref` internally.

If we want to improve readability, may I suggest we remove the need for that function altogether? It's only used in `handleShowTooltip` and `handleHideTooltip` and seems a premature abstraction.

Can I suggest the following refactor:

```tsx
   const handleShowTooltip = () => {
        if (disabled) {
            return;
        }

        if (timeoutRef.current) {
            window.clearTimeout(timeoutRef.current);
        }

        if (enterDelay === 0) {
            setOpen(true);
        } else {
            timeoutRef.current = window.setTimeout(() => {
                setOpen(true);
            }, enterDelay);
        }
    };

    const handleHideTooltip = useCallback(() => {
        if (timeoutRef.current) {
            window.clearTimeout(timeoutRef.current);
        }

        if (leaveDelay === 0) {
            setOpen(false);
        } else {
            timeoutRef.current = window.setTimeout(() => {
                setOpen(false);
            }, leaveDelay);
        }
    }, [leaveDelay]);
```
Yes, there is a little bit of duplication (3 lines?) but we remove altogether the need to:

- have a shared function
- pass the ref within it
- pass a callback

What do you think?

packages/fondue/src/components/Tooltip/Tooltip.tsx
floroz: Yes, that sounds good to me 👍🏻

packages/fondue/src/components/Tooltip/Tooltip.tsx
bojangles-m: Unfortunately, this creates more cognitive load on the simple function and main function, and the the handle methods have to much responsibility and it is not clear where all the flow happens inside of the main function. and also the function should to less things and not more to easier grasp the content.

I can not relate to that because the var is coming from the main component.

packages/fondue/src/components/Tooltip/Tooltip.tsx
floroz: my bad! I have always typed it to`ReturnType<typeof setTimeout>` but since this only executes with the browser timer, and not Node, and using the`window` namespace, the `number` is fine :)

packages/fondue/src/components/Tooltip/Tooltip.tsx
floroz: Yes, I think I can agreed on the multiple responsibility (state update + timer management). 

I would like to propose an alternative solution to avoid side-effects outside of the component:

```ts
    const delayedSetOpen = (show: boolean, delay: number) => {
        if (timeoutRef.current) {
            window.clearTimeout(timeoutRef.current);
        }

        timeoutRef.current = undefined;

        if (delay === 0) {
            setOpen(show);
        } else {
            timeoutRef.current = window.setTimeout(() => {
                setOpen(show);
            }, delay);
        }
    };

    const handleShowTooltip = () => {
        if (disabled) {
            return;
        }

        delayedSetOpen(true, enterDelay);
    };

    const handleHideTooltip = () => {
        delayedSetOpen(false, leaveDelay);
    );

```

bojangles-m: I appreciate your input, its helpful and would like to propose another solution to avoid side effects:
```
this function is still outside of the main component 
const handleTimeout = (callback: () => void, delay: number, timeoutRef: RefObject<number | undefined>) => {
    let id = timeoutRef.current;

    if (id) {
        window.clearTimeout(id);
    }

    if (delay > 0) {
        id = window.setTimeout(callback, delay);
    } else {
        callback();
    }

    return id;
};

and this are inside the Tooltip method
const handleShowTooltip = () => {
        if (!disabled) {
            timeoutRef.current = handleTimeout(() => setOpen(true), enterDelay, timeoutRef);
        }
    };

    const handleHideTooltip = useCallback(() => {
        timeoutRef.current = handleTimeout(() => setOpen(false), leaveDelay, timeoutRef);
    }, [leaveDelay]);
```

If the name is not right in your opinion I would not name it `delayedSetOpen` because the funciton is also used for delayed closing the Tooltip.

floroz: Yes, I like your proposal. If we have a signature of:

```ts
type HandleTimeout = (cb: () => void, delay: number, timer?: number): number 
```
we have removed the side-effect from its body.

```ts
const handleTimeout = (callback: () => void, delay: number, timer?: number) => {
   let newTimer: number;

    if (timer) {
        window.clearTimeout(timer);
    }

    if (delay > 0) {
        newTimer = window.setTimeout(callback, delay);
    } else {
        callback();
    }

    return newTimer;
};
```

With this change, since we want to keep it outside from the component, we have completely decoupled the logic from React APIs.

bojangles-m: yes true, to decouple and get rid of side-effects.
I just left null because it is the clear timeout is using it.
```
const handleTimeout = (callback: () => void, delay: number, timeoutId: number | null) => {
    let newTimeoutId = timeoutId;

    if (newTimeoutId) {
        window.clearTimeout(newTimeoutId);
    }

    if (delay > 0) {
        newTimeoutId = window.setTimeout(callback, delay);
    } else {
        callback();
    }

    return newTimeoutId;
};
```

bojangles-m: If we want pure functions in this case :)

bojangles-m:
@SamuelAlev I updated it, could you check again please?

SamuelAlev (approved):

PR #1755, opened 2024-03-15
chore: Handling z-index on InlineDialog

selimovicz:


floroz (commented):
Can I ask what is reason behind this change? 

Accepting an arbitrary z-index for overlayed component is problematic because it can results in z-index conflicts if we don't enforce a scaling system.

From a UX standpoint, this `InlineDialog` seems more of a Popover to me than a Dialog.

Do you need to pass a custom `z-index` to lower or increase the value?

selimovicz:
> Can I ask what is reason behind this change?
> 
> Accepting an arbitrary z-index for overlayed component is problematic because it can results in z-index conflicts if we don't enforce a scaling system.
> 
> From a UX standpoint, this `InlineDialog` seems more of a Popover to me than a Dialog.
> 
> Do you need to pass a custom `z-index` to lower or increase the value?

Reasonable concern. So basically we are having Asset chooser module, which is written in terrific, usually rendered in modal. The InlineDialog component is rendered as part of such module. Since z-index of the Asset chooser modal is bigger than z-index of the InlineDialog we are stumbling upon issue where InlineDialog is not visible at all. Take a look into the screenshot below, the issue is with Colors dropdown, where InlineDialog is not visible.

Screenshot 2024-03-15 at 09 34 27

ftascon:
> > Can I ask what is reason behind this change?
> > Accepting an arbitrary z-index for overlayed component is problematic because it can results in z-index conflicts if we don't enforce a scaling system.
> > From a UX standpoint, this `InlineDialog` seems more of a Popover to me than a Dialog.
> > Do you need to pass a custom `z-index` to lower or increase the value?
> 
> Reasonable concern. So basically we are having Asset chooser module, which is written in terrific, usually rendered in modal. The InlineDialog component is rendered as part of such module. Since z-index of the Asset chooser modal is bigger than z-index of the InlineDialog we are stumbling upon issue where InlineDialog is not visible at all. Take a look into the screenshot below, the issue is with Colors dropdown, where InlineDialog is not visible.
> 
> Screenshot 2024-03-15 at 09 34 27

a bit of [context here](https://www.loom.com/share/73222b3f97274658ab31cceffcd4b258?sid=a0726e60-a272-4dc6-a09a-597cbcaf68ba)

floroz:
Thank you @selimovicz for the context and @ftascon for creating the video. I've got a better understanding now.

My proposal would be to add the following prop:

```ts
type Props = {
   // @default false; 
  useWithinDialog: boolean;
}
```

I think it's important we retain control of zindex within the library to maintain consistency across usage of layered element.

Now, when `useWithinDialog` is `true` a new z-index of ~1500 should be applied, but from your video @ftascon seems that even 2000 doesn't work correctly. 

The underlying modal should be using a `z-index: 1200` but seems that's not the case. 

Could you tell me which is the file that is being updated in `web-app` so that I can take a look?

floroz:
> Thank you @selimovicz for the context and @ftascon for creating the video. I've got a better understanding now.
> 
> My proposal would be to add the following prop:
> 
> ```ts
> type Props = {
>    // @default false; 
>   useWithinDialog: boolean;
> }
> ```
> 
> I think it's important we retain control of zindex within the library to maintain consistency across usage of layered element.
> 
> Now, when `useWithinDialog` is `true` a new z-index of ~1500 should be applied, but from your video @ftascon seems that even 2000 doesn't work correctly.
> 
> The underlying modal should be using a `z-index: 1200` but seems that's not the case.
> 
> Could you tell me which is the file that is being updated in `web-app` so that I can take a look?

btw this is a UX pattern we need to structure at a library level in terms of managing those token/values, Popover, Tooltip, etc and their usage within Dialogs/Modal

cc @noahwaldner @SamuelAlev

PR #1758, opened 2024-03-19
feat: disabled items unchecked on multiselect tree

anxobotana:
Issue: disabled items get selected and disabled when selecting a parent.
The intention of this PR is not showing as selected a disabled item, it's confusing and misleading.

## Definition of Done

### DIFF
--- a/packages/fondue/src/components/Tree/TreeItem/TreeItemMultiselect.tsx
+++ b/packages/fondue/src/components/Tree/TreeItem/TreeItemMultiselect.tsx
@@ -14,7 +14,7 @@ import {
 } from '../helpers/multiselectTreeItemstyling';
 
 import { ExpandButton } from './ExpandButton';
-import { Checkbox, CheckboxEmphasis, CheckboxSize } from '@components/Checkbox/Checkbox';
+import { Checkbox, CheckboxEmphasis, CheckboxSize, CheckboxState } from '@components/Checkbox/Checkbox';
 import { Container } from '@components/Container';
 import { useMultiselectTreeItem } from './useMultiselectTreeItem';
 
@@ -139,7 +139,11 @@ export const TreeItemMultiselect = memo(
                         label=""
                         onChange={checkBoxOnSelect}
                         size={CheckboxSize.Default}
-                        state={getMultiselectCheckBoxState(isSelected, isPartialSelected)}
+                        state={
+                            isDisabled
+                                ? CheckboxState.Unchecked
+                                : getMultiselectCheckBoxState(isSelected, isPartialSelected)
+                        }
                         tooltip={[]}
                         value={id}
                     />
--- a/.changeset/five-moons-teach.md
+++ b/.changeset/five-moons-teach.md
@@ -1,0 +1,5 @@
+---
+"@frontify/fondue": patch
+---
+
+feat: disabled items unchecked on multiselect tree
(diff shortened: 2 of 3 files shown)


floroz:
>The intention of this PR is not showing as selected a disabled item, it's confusing and misleading.

I think it's reasonable for something to be selected _and_ disabled. We have disabled checkboxes that are selected.

If I am understanding your description correctly, the issue is that, when a parent is disabled, you can still interact with its children?

@anxobotana did I get it right?

anxobotana:
> > The intention of this PR is not showing as selected a disabled item, it's confusing and misleading.
> 
> I think it's reasonable for something to be selected _and_ disabled. We have disabled checkboxes that are selected.
> 
> If I am understanding your description correctly, the issue is that, when a parent is disabled, you can still interact with its children?
> 
> @anxobotana did I get it right?

No, the problem is that the disabled children get selected.
As an example, imaging we want to select items to move or copy to a different place, how do I know as an user if the disabled and **selected** item is going to be copied/moved or is going to be excluded because is disabled? It is confusing and misleading

So currently the multiselect behaves like this:
Screenshot 2024-03-19 at 16 43 40

Notice the `Subitem 2` which is disabled and checked

While my suggestion and our expected behaviour will be:
Screenshot 2024-03-19 at 16 44 37

packages/fondue/src/components/Tree/TreeItem/TreeItemMultiselect.tsx
```diff
@@ -139,7 +139,11 @@ export const TreeItemMultiselect = memo(
                         label=""
                         onChange={checkBoxOnSelect}
                         size={CheckboxSize.Default}
-                        state={getMultiselectCheckBoxState(isSelected, isPartialSelected)}
+                        state={
+                            isDisabled
+                                ? CheckboxState.Unchecked
+                                : getMultiselectCheckBoxState(isSelected, isPartialSelected)
+                        }
```
floroz, line 146: Thanks @anxobotana 

I think I understand better now.

My only concern with this change is that it _always_ unchecks the Checkbox when node is disabled.

What if we have an initial state of both _checked_ and _disabled_? That would be valid but this logic would reset the state to unchecked.

anxobotana: I understand your concern. I don't think your suggestion will work, since the selected state is calculated in the `useMultiselectTreeItem` and then this selected state propagates up and down in the tree. 

What about adding a new prop where the consumer can choose between both behaviors: `uncheckOnDisabled` default to `false`. 
and then
```suggestion
                          isDisabled && uncheckOnDisabled
                                ? CheckboxState.Unchecked
                                : getMultiselectCheckBoxState(isSelected, isPartialSelected)
```

floroz: Sorry @anxobotana for the confusion, mine was a question not a suggestion 😄 

I'll take a look into code to get more familiar with the logic and will share with you any findings

anxobotana: No problem at all, @floroz 

Thanks for looking into this

floroz:
@anxobotana could you create a `patch` changeset for this to be merged?

anxobotana:
> @anxobotana could you create a `patch` changeset for this to be merged?

@floroz added
Thanks

floroz (approved):

PR #1759, opened 2024-03-19
test: add percy.io visual regression

floroz:
## Description

This PR integrates Percy on top of Playwright execution, by differentiating two separate run of playwright

- files with `*.ct.tsx` are component testing that run for every commit
- files with `*.percy.tsx` are percy visual snapshots that we can control to only run under certain conditions (to control amount of snapshot generated).

Percy runs only when the label `percy` is applied to mitigate against excessive (and expensive) multiple snapshot runs.

Task CU-86943w14a


floroz:
Integration with the App is completed:

Screenshot 2024-03-20 at 11 39 38

PR #1760, opened 2024-03-20
feat: add tabIndex prop to Button

mikeyrayvon:
This adds a prop to the Button to control the `tabIndex` of the child `button` element. Increases opportunities for a11y compliance

CU-869447ekj

## Definition of Done

### DIFF
--- a/packages/fondue/src/components/Button/Button.tsx
+++ b/packages/fondue/src/components/Button/Button.tsx
@@ -51,6 +51,7 @@ export type ButtonProps = {
     /** @deprecated inverted can be done by wrapping the component in a className="tw-dark" */
     inverted?: boolean;
     'data-test-id'?: string;
+    tabIndex?: number;
 };
 
 const ButtonComponent: ForwardRefRenderFunction<HTMLButtonElement | null, ButtonProps> = (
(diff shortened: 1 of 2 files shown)


packages/fondue/src/components/Button/Button.tsx
```diff
@@ -122,6 +124,7 @@ const ButtonComponent: ForwardRefRenderFunction<HTMLButtonElement | null, Button
             disabled={disabled}
             form={formId}
             title={title}
+            tabIndex={tabIndex}
```
floroz, line 127: is this change needed because you have to disable the element (`tabindex="-1"`) or do you want to set a custom tabbable order?

mikeyrayvon: I want to set a custom tabbable order

mikeyrayvon: Ideally we would have this prop on every interactive component. @LidijaStoilova perhaps you have thoughts on this as well?

Separate issue, but there are also a few Fondue form elements that require two `tab` key presses to bypass in the tab order

floroz: >Separate issue, but there are also a few Fondue form elements that require two tab key presses to bypass in the tab order

Yeah, that sounds like a bug. @mikeyrayvon  Do know which components so we can create ticket for our backlog?

LidijaStoilova: Agree with @floroz. Unfortunately I'm not able to access the deploy preview, but happy to take a look at it together @mikeyrayvon.

PR #1763, opened 2024-03-21
feat(fondue-ui): setup build for standalone library

floroz:
## Description

Setup a standalone library build for `fondue-ui` (formerly `fondue-components`). 

The build produces ESM (with modules preserved) and UMD artefacts, alongside type declaration and sourcemaps.

- The library is not minified: this is responsability of consumers
- The library ships without `preflight` (CSS reset) from Tailwind: this needs to be discuss further as we have 2 layers of Tailwind to resolve (`fondue-ui` and `fondue`).

## Open Questions

- How should the library be consumed? 

- How to deal with Tailwind styles duplication?

- How to deal with 2 baselines `style.css` that consumers will need?

- How do we phase out components from `fondue` to `fondue-ui`


packages/components/package.json
```diff
@@ -12,12 +11,21 @@
     "bugs": {
         "url": "https://github.com/frontify/fondue/issues"
     },
-    "main": "src/index.ts",
-    "types": "src/index.ts",
-    "engines": {
```
floroz, line 17: engine and package manager are only needed at the root of the project `package.json`

packages/components/package.json
```diff
@@ -12,12 +11,21 @@
...
+        "CHANGELOG.md"
+    ],
+    "exports": {
+        ".": {
+            "import": "./dist/index.es.js"
+        },
+        "./style.css": "./dist/style.css"
     },
     "scripts": {
+        "build": "vite build",
```
SamuelAlev, line 28: This forces us to build everytime we make a change, for better DX, what if we do it only on CI (changing the `main` from `src/index.ts` to `dist/index.js`, ...)?

packages/components/package.json
```diff
@@ -12,12 +11,21 @@
...
+    "main": "dist/index.umd.js",
+    "module": "dist/index.es.js",
+    "types": "dist/index.d.ts",
+    "files": [
+        "dist",
+        "CHANGELOG.md"
+    ],
+    "exports": {
+        ".": {
+            "import": "./dist/index.es.js"
```
SamuelAlev, line 23: Please add `type` ([docs](https://www.typescriptlang.org/docs/handbook/modules/reference.html#example-explicit-types-condition))

packages/components/package.json
SamuelAlev: We actually need it when installing the package alone (without `@frontify/fondue`)

packages/components/vite.config.ts
```diff
@@ -1,19 +1,60 @@
...
+        minify: false,
+        sourcemap: true,
+        rollupOptions: {
+            external: [
+                ...Object.keys(pkg.peerDependencies ?? {}),
+                ...Object.keys(pkg.dependencies ?? {}),
+                'react/jsx-runtime',
+            ],
+            output: [
+                {
```
SamuelAlev, line 45: Please go full ES and no more UMD

packages/components/vitest.config.ts
```diff
@@ -0,0 +1,17 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+import { defineConfig } from 'vitest/config';
+
+export default defineConfig({
```
SamuelAlev, line 5: Why split it?

packages/components/package.json
floroz: There's a few workaround to improve the DX (I think first we need to identify which is the sandbox/development platform, if we have an `fondue-ui/storybook` then we won't have that problem at all).

To avoid creating separate processes for local and CI, we can easily disable declarations and have `vite build watch` when developing.
I want to keep this draft as a base for discussion, but before we can commit to any solution we need to tackle a few higher level decisions:

- how do we deal with Tailwind
- how do we phase out components
- how do we compose Storybook

Based on these discussions we can review the best strategy to build/ship `fondue-ui`

packages/components/package.json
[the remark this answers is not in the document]
floroz: yes, holding off on that until we've got time to chat all together about the plan

packages/components/vitest.config.ts
floroz: separation of concerns, to avoid bloating the vite config with testing setup responsibilities.

packages/components/tailwind.config.ts
```diff
@@ -12,6 +12,13 @@ export default {
             strategy: 'class',
         }),
     ],
+    corePlugins: {
+        /**
+         * Disable preflight to avoid conflicts with global styles
+         * TODO: there are several architectural conflicts with a Tailwind toolchain of libraries that needs to be resolved before we commit to use Tailwind in this package.
+         */
+        preflight: false,
+    },
```
floroz, line 21: this comment (or even this change) won't need to be merged. This is to start a discussion around Tailwind

packages/components/vite.config.ts
```diff
@@ -1,19 +1,53 @@
...
+        dts({
+            compilerOptions: {
+                sourceMap: true,
+            },
+            exclude: ['**/*.stories.*', '**/*.test.*', '**/*.spec.*', '**/*.ct.*', '**/__tests__/**', '**/*.mdx'],
+        }),
+    ],
+    build: {
+        lib: {
+            entry: join(__dirname, 'src/index.ts'),
```
SamuelAlev, line 33:
```suggestion
            entry: './src/index.ts',
```

packages/components/vite.config.ts
```diff
@@ -1,19 +1,53 @@
...
+        rollupOptions: {
+            external: [
+                ...Object.keys(pkg.peerDependencies ?? {}),
+                ...Object.keys(pkg.dependencies ?? {}),
+                'react/jsx-runtime',
+            ],
+            output: {
+                format: 'es',
+                name: 'FondueUI',
+                preserveModulesRoot: join(__dirname, 'src'),
```
SamuelAlev, line 47:
```suggestion
                preserveModulesRoot: 'src',
```

packages/components/package.json
```diff
@@ -48,8 +57,11 @@
         "postcss-import": "^16.0.1",
         "prettier": "^3.2.5",
         "react": "^18.2.0",
+        "react-dom": "^18.2.0",
         "tailwindcss": "^3.4.1",
         "typescript": "^5.3.3",
+        "vite": "^5.1.3",
```
SamuelAlev, line 63:
```suggestion
        "vite": "^5.2.6",
```

packages/components/package.json
```diff
@@ -48,8 +57,11 @@
...
+        "vite-plugin-dts": "^3.7.2",
```
SamuelAlev, line 64:
```suggestion
        "vite-plugin-dts": "^3.7.3",
```

packages/components/package.json
SamuelAlev: Not if you install directly `@frontify/fondue-ui` (which is also a possibility)

PR #1764, opened 2024-03-21
fix: implemented forward ref on text input component

bojangles-m:


packages/fondue/src/components/TextInput/TextInput.tsx
```diff
@@ -125,246 +126,261 @@ const ExtraActionButton: ForwardRefRenderFunction<HTMLButtonElement | null, Extr
...
+            focusOnMount,
+            selectable = false,
+            extraActions,
+            maxLength,
+            ...props
+        },
+        ref,
+    ): ReactElement => {
+        const inputElement = useRef<HTMLInputElement | null>(null);
+        useImperativeHandle(ref, () => inputElement.current!, []);
```
floroz, line 162: Is the imperative handle needed? We are returning the entire ref and not a subset of it.

I think we can remove both the handle and the `inputElement` internal ref, and just proxy the `ref` coming to the component directly to the DOM node?

bojangles-m: We have useEffect which sets the focus on the element thats why is Imperative handle there. Do you suggest to change it?

packages/fondue/src/components/TextInput/TextInput.tsx
```diff
@@ -125,246 +126,261 @@ const ExtraActionButton: ForwardRefRenderFunction<HTMLButtonElement | null, Extr
...
-    useEffect(() => {
-        if (typeof obfuscated === 'boolean') {
-            setIsObfuscated(obfuscated);
-        }
-    }, [obfuscated]);
+        useEffect(() => {
+            setTimeout(() => {
+                focusOnMount && inputElement.current?.focus();
+            }, 0);
+        }, [focusOnMount]);
```
floroz, line 181: We don't need this logic here if we expose the entire `ref` (or even just the `focus` method). 

The clients can decide how to manage the focus themselves (on mount, on action, etc). Basically this `useEffect` should live in the client's code:

```ts
import { Button } from '@frontify/fondue';

// the rest of the code [...]

const fondueBtnRef = useRef();

useEffect(() => {
 if (fondueBtnRef.current) {
   fondueBtnRef.current.focus();
}, []);


return <Button ref={fondueBtnRef}>Click Me</Button>
```

packages/fondue/src/components/TextInput/TextInput.tsx
floroz: okay, my bad, the GitHub diff showed that part as a change of this PR, but I can see now that it was already there, sorry! 🙏🏻

packages/fondue/src/components/TextInput/TextInput.tsx
floroz: nevermind, it was a glitch in the GitHub diff, this code was already there and was not changed!

floroz (approved):

packages/fondue/src/components/TextInput/TextInput.tsx
bojangles-m: Just to clarify, should  I just remove the 'focusOnMount' prop?
I agree to this option. this is a breaking change in web-app and a lot of times needed?

floroz: yep we should keep it to avoid breaking changes!

bojangles-m:
@floroz thnx for input, I updated the story and deprecate the prop. Could you check please.

.changeset/wicked-parrots-live.md
```diff
@@ -0,0 +1,5 @@
+---
+"@frontify/fondue": minor
+---
+
+fix(TextInput): exposed component ref and set focusOnMount prop to be deprecated.
```
floroz, line 5:
```suggestion
feat(TextInput): exposed component ref and set focusOnMount prop to be deprecated.
```

packages/fondue/src/components/TextInput/TextInput.tsx
```diff
@@ -63,6 +64,7 @@ export type TextInputBaseProps = {
     maxLength?: number;
     size?: number;
     spellcheck?: boolean;
+    /** @deprecated since v-beta.418, to set focus use ref exposed from the component */
```
floroz, line 67: We don't need to mention the version here, this comment won't be visible prior to `v-beta.418`

```suggestion
    /** @deprecated a `ref` is now available so that focus can be set programmatically 
    
         @example
         ```tsx
          import { TextInput } from '@frontify/fondue';
          
         const Component = () => {
         const inputRef = useRef();
         
         useEffect(() => {
            if (inputRef.current) {
                inputRef.current.focus();
            }
         }, [)
           return <TextInput ref={inputRef} />
         }
         ```
      */
```

packages/fondue/src/components/TextInput/TextInput.stories.tsx
```diff
@@ -152,7 +162,7 @@ export const FocusOnMount = TextInputTemplate.bind({});
 
 FocusOnMount.args = {
     value: 'Value text',
-    focusOnMount: true,
+    focus: true,
```
floroz, line 165: I think since we've deprecated the `focusOnMount` from the API, we can remove this Story altogether, or not make changes to the current Storybook.

The way to setup focus programmatically is pretty standard and not a foundue-specific implementation, so we don't need to document it further

.changeset/wicked-parrots-live.md
bojangles-m: Updated

packages/fondue/src/components/TextInput/TextInput.tsx
bojangles-m: Added

packages/fondue/src/components/TextInput/TextInput.stories.tsx
bojangles-m: Removed the story

floroz (approved):

PR #1768, opened 2024-03-25
fix(TextInput): Disable Styling with grey background

imoutaharik:
Match same disable styling as the Password Input


floroz (commented):
Hey @imoutaharik, would you mind adding a `Disabled` story to this component?

floroz (approved):

PR #1770, opened 2024-03-25
chore: fix prettier errors and add changelog to prettierignore

floroz:
## Description

- fixes `prettier` error in the project
- adds `CHANGELOG.md` to ignore file

### DIFF
--- a/packages/components/CHANGELOG.md
+++ b/packages/components/CHANGELOG.md
@@ -4,4 +4,4 @@
 
 ### Minor Changes
 
-- [#1731](https://github.com/Frontify/fondue/pull/1731) [`ef0519d`](https://github.com/Frontify/fondue/commit/ef0519d8021204b280b65812c98686b4ce3b909e) Thanks [@noahwaldner](https://github.com/noahwaldner)! - added export for the new components in fondue
+-   [#1731](https://github.com/Frontify/fondue/pull/1731) [`ef0519d`](https://github.com/Frontify/fondue/commit/ef0519d8021204b280b65812c98686b4ce3b909e) Thanks [@noahwaldner](https://github.com/noahwaldner)! - added export for the new components in fondue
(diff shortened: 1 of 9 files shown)


packages/components/CHANGELOG.md
```diff
@@ -4,4 +4,4 @@
 
 ### Minor Changes
 
-- [#1731](https://github.com/Frontify/fondue/pull/1731) [`ef0519d`](https://github.com/Frontify/fondue/commit/ef0519d8021204b280b65812c98686b4ce3b909e) Thanks [@noahwaldner](https://github.com/noahwaldner)! - added export for the new components in fondue
+-   [#1731](https://github.com/Frontify/fondue/pull/1731) [`ef0519d`](https://github.com/Frontify/fondue/commit/ef0519d8021204b280b65812c98686b4ce3b909e) Thanks [@noahwaldner](https://github.com/noahwaldner)! - added export for the new components in fondue
```
floroz, line 7: There could be a conflict between `prettier` and `markdown` generation here where we do automatic releases.

Should we ignore `CHANGELOG.md` or run a `prettier write` as part of the release creation job? cc @noahwaldner @SamuelAlev

floroz: I added an exclusion for `CHANGELOG.md` in `.prettierignore`.

.github/workflows/fondue-pull-requests.yaml
SamuelAlev, line 1: PR is not neccesary with the latest version of the Frontify ESLint/Prettier config (should be handled there instead of separate Prettier job)

floroz: updated the PR 👍🏻

SamuelAlev (approved):

packages/components/CHANGELOG.md
SamuelAlev: Don't we need to revert this file?

PR #1773, opened 2024-03-25
ci: changeset workflow

floroz:
Automates check for `.changeset/` files in PR

After merging, we need to add to "Required" steps in branch protection

### DIFF
--- a/.github/workflows/changeset-check.yaml
+++ b/.github/workflows/changeset-check.yaml
@@ -19,7 +19,7 @@ jobs:
                     fetch-depth: 0
             - name: Check for changeset file
               run: |
-                    if git diff --name-only --diff-filter=A origin/main HEAD '.changeset/*.md' | grep '.*'; then; then
+                    if git diff --name-only --diff-filter=A origin/main HEAD '.changeset/*.md' | grep '.*'; then
                         echo "Changeset file found!"
                         exit 0
                     else
--- a/.github/workflows/changeset-check.yaml
+++ b/.github/workflows/changeset-check.yaml
@@ -14,10 +14,11 @@ jobs:
         steps:
             - name: Checkout code
               uses: actions/checkout@v2
-
+              with:
+                    fetch-depth: 0
             - name: Check for changeset file
               run: |
-                    if git diff --name-only HEAD^ HEAD | grep -qE '^packages/.*\.changeset/.*\.md'; then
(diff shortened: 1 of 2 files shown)


floroz:
Run without changeset added: https://github.com/Frontify/fondue/actions/runs/8423035268/job/23063734426?pr=1773

floroz:
Run with changeset added: https://github.com/Frontify/fondue/actions/runs/8423143329/job/23064078188?pr=1773

noahwaldner (approved):

.github/workflows/changeset-check.yaml
```diff
@@ -0,0 +1,26 @@
...
+        runs-on: ubuntu-latest
+
+        steps:
+            - name: Checkout code
+              uses: actions/checkout@v2
+              with:
+                    fetch-depth: 0
+            - name: Check for changeset file
+              run: |
+                    if git diff --name-only origin/main..HEAD -- '.changeset/' | grep -qE '^.changeset/.*\.md'; then
```
SamuelAlev, line 21:
```suggestion
                    if git diff --name-only --diff-filter=A origin/main HEAD '.changeset/**.md'; then
```
Then we only check for "Added" files

.github/workflows/changeset-check.yaml
```diff
@@ -0,0 +1,26 @@
...
+                        echo "Changeset file found!"
```
SamuelAlev, line 22:
```suggestion
                        echo "Changeset file found!"
                        exit 0
```

.github/workflows/changeset-check.yaml
```diff
@@ -0,0 +1,26 @@
...
+on:
+    pull_request:
+        branches:
+            - main
+        paths:
+            - 'packages/**'
+
+jobs:
+    changeset-check:
+        runs-on: ubuntu-latest
```
SamuelAlev, line 12:
```suggestion
        runs-on: ubuntu-latest
        timeout: 3
```

.github/workflows/changeset-check.yaml
floroz: Implemented your suggestion but now we don't get a correct failure and have a false positive: https://github.com/Frontify/fondue/actions/runs/8433112659/job/23093496088?pr=1773

floroz: fixed, we were missing to pipe `grep` to read the output of `git diff`

SamuelAlev (approved):

PR #1794, opened 2024-04-08
feat: add theming to global storybook

noahwaldner:


packages/components/.storybook/preview.ts
```diff
@@ -17,6 +17,31 @@ const preview: Preview = {
                 date: /Date$/i,
             },
         },
+        status: {
+            type: 'legacy',
+            statuses: {
```
SamuelAlev, line 22: What about adding `in progress`?

packages/fondue/vite.config.ts.timestamp-1712583916564-e3090d6204079.mjs
SamuelAlev, line 1: can we add a gitignore for the `timestamp` files?

packages/components/.storybook/preview.ts
noahwaldner: added

PR #1798, opened 2024-04-10
refactor: refactor fondue button

noahwaldner:
Task CU-86949yvwn


packages/components/src/utilities/focusStyle.ts
```diff
@@ -0,0 +1,12 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+import { twMerge } from 'tailwind-merge';
```
SamuelAlev, line 3: Could you create a utility `cn` or `merge` or something else which just wraps the `twMerge` utility?
Also please add tests to it

packages/components/vite.config.ts
```diff
@@ -1,10 +1,11 @@
 /* (c) Copyright Frontify Ltd., all rights reserved. */
 
 import react from '@vitejs/plugin-react';
-import { defineConfig } from 'vitest/config';
+import tsConfigPaths from 'vite-tsconfig-paths';
+import { type UserConfig, defineConfig } from 'vitest/config';
 
 export default defineConfig({
-    plugins: [react()],
+    plugins: [react(), tsConfigPaths()] as UserConfig['plugins'],
```
SamuelAlev, line 8:
```suggestion
    plugins: [react(), tsConfigPaths()],
```

packages/icons/src/createFondueIcon.ts
```diff
@@ -18,7 +18,7 @@ export type IconNode = [elementName: keyof ReactSVG, attrs: Record<string, strin
 export type SVGAttributes = Partial<SVGProps<SVGSVGElement>>;
 type ComponentAttributes = RefAttributes<SVGSVGElement> & SVGAttributes;
 
-type IconSizeAsNumber = 12 | 16 | 20 | 24 | 32;
+export type IconSizeAsNumber = 12 | 16 | 20 | 24 | 32;
```
SamuelAlev, line 21: Not a fan of exposing it, could you maybe use `FondueIconProps['size']` instead?

packages/components/src/components/Button/tests/Button.test.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+import { Button, type ButtonStyle } from '..';
+
+const BUTTON_TEXT = 'Frontify';
+const BUTTON_TEST_ID = 'button';
+
+const styles: ButtonStyle[] = ['default', 'positive', 'negative', 'danger', 'loud'];
+
+describe('Button component', () => {
+    for (const style of styles) {
```
SamuelAlev, line 15: Not a good practices to add a for each like this as it doesn't really test anything.

Also shouldn't the test be in Playwright?

packages/components/src/components/Button/styles/buttonStyles.ts
```diff
@@ -0,0 +1,167 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+import { type IconSizeAsNumber } from '@frontify/fondue-icons';
+import { tv } from 'tailwind-variants';
+
+import { type ButtonSize } from '../';
```
SamuelAlev, line 6: Please put the full path

packages/components/src/components/Button/Button.tsx
```diff
@@ -0,0 +1,105 @@
...
+                disabled,
+                rounding,
+                size,
+                hugWidth,
+                emphasis,
+                style,
+                iconOnly: (icon && !children) || hideLabel,
+            })}
+        >
+            {icon && (
```
SamuelAlev, line 83: Icon should not be known by the Button component, please remove it and maybe we can add a size `icon` or `square` which make the button the same width and height.

packages/components/src/components/Button/Button.tsx
```diff
@@ -0,0 +1,105 @@
...
+        icon,
+        children,
+        onClick,
+        hugWidth = true,
+        formId,
+        'aria-label': ariaLabel,
+        'aria-describedby': ariaDescribedBy,
+        'data-test-id': dataTestId = 'button',
+    }: ButtonProps,
+    ref: React.ForwardedRef<HTMLButtonElement | null>,
```
SamuelAlev, line 59:
```suggestion
    ref: ForwardedRef<HTMLButtonElement | null>,
```
We don't use the React namespace

SamuelAlev (changes requested):

packages/components/src/utilities/focusStyle.ts
noahwaldner: alias added

packages/components/vite.config.ts
noahwaldner: all plugin cause this

 Type 'PluginOption[]' is not assignable to type 'PluginOption

noahwaldner: why do you dislike the array type?

packages/icons/src/createFondueIcon.ts
noahwaldner: agree, changed

packages/components/src/components/Button/styles/buttonStyles.ts
noahwaldner: added

packages/components/src/components/Button/Button.tsx
noahwaldner: how do you mean? you don't want to pass in the icon as a prop to the button?

packages/components/src/components/Button/Button.tsx
noahwaldner: yep, oversight from my side.

Removed

packages/components/src/components/Button/Button.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+import { buttonIconSizeMap, buttonStyles } from './styles/buttonStyles';
+import { iconStyles } from './styles/iconStyles';
+import { textStyles } from './styles/textStyles';
+
+export type ButtonRounding = 'medium' | 'full';
+
+export type ButtonStyle = 'default' | 'positive' | 'negative' | 'danger' | 'loud';
+
+export type UpdatedButtonStyle = 'default' | 'positive' | 'negative' | 'danger' | 'loud';
```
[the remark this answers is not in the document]
noahwaldner: removed

packages/components/src/components/Button/styles/textStyles.ts
```diff
@@ -0,0 +1,145 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+import { tv } from 'tailwind-variants';
```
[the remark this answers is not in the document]
noahwaldner: utility added

packages/components/src/components/Button/Button.tsx
```diff
@@ -0,0 +1,105 @@
...
+        hideLabel = false,
+        disabled = false,
+        icon,
+        children,
+        onClick,
+        hugWidth = true,
+        formId,
+        'aria-label': ariaLabel,
+        'aria-describedby': ariaDescribedBy,
+        'data-test-id': dataTestId = 'button',
```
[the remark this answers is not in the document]
noahwaldner: changed

packages/components/src/components/Button/Button.tsx
SamuelAlev: Exactly, IMO should be
```tsx
<Button size="icon"> // or square
  <IconABC size={20} />
</Button>
```

SamuelAlev: So we can avoid the `cloneElement`

noahwaldner: let's have a chat about this tomorrow

packages/components/src/components/Button/styles/textStyles.ts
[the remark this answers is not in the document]
noahwaldner: tests added

packages/components/vite.config.ts
noahwaldner: @SamuelAlev the issues persists for me

packages/components/src/components/Button/tests/Button.test.tsx
noahwaldner: fixed

packages/components/package.json
```diff
@@ -55,12 +54,13 @@
...
         "@testing-library/jest-dom": "^6.4.2",
-        "@testing-library/react": "^15.0.0",
+        "@testing-library/react": "^15.0.1",
         "@types/node": "^20.12.7",
-        "@types/react": "^18.2.75",
+        "@types/react": "^18.2.77",
         "@vitejs/plugin-react": "^4.2.1",
         "@vitest/coverage-v8": "^1.4.0",
         "@vitest/ui": "^1.4.0",
+        "babel-plugin-macros": "^3.1.0",
```
SamuelAlev, line 63:
```suggestion
```

packages/components/src/components/Button/Button.tsx
```diff
@@ -0,0 +1,103 @@
...
+    size?: ButtonSize;
+    rounding?: ButtonRounding;
+    disabled?: boolean;
+    icon?: ReactElement;
+    children?: ReactNode;
+    onClick?: (event?: MouseEvent<HTMLButtonElement>) => void;
+    hugWidth?: boolean;
+    'aria-label'?: string;
+    'aria-describedby'?: string;
+    formId?: string;
```
SamuelAlev, line 34: We should enforce having the button _inside_ a form and not outside, what do you think about removing it?

packages/components/src/components/Button/Button.tsx
```diff
@@ -0,0 +1,103 @@
...
+            {children && (
+                <span className={textStyles({ hideLabel, emphasis, style })} data-test-id={`${dataTestId}-text`}>
+                    {children}
+                </span>
+            )}
+        </button>
+    );
+};
+
+export const Button = forwardRef<HTMLButtonElement | null, ButtonProps>(ButtonComponent);
```
SamuelAlev, line 102: What about removing the extra `const` and extra typing `ref: ForwardedRef<HTMLButtonElement | null>` and the props `}: ButtonProps,`? Like this

Screenshot 2024-04-15 at 10 39 19

packages/components/src/components/Button/tests/Button.test.tsx
```diff
@@ -0,0 +1,30 @@
...
+        const clickListener = vi.fn();
+        const { getByTestId } = render(<Button onClick={clickListener}>{BUTTON_TEXT}</Button>);
+        const button = getByTestId(BUTTON_TEST_ID);
+        expect(clickListener).not.toHaveBeenCalled();
+        button.click();
+        expect(clickListener).toHaveBeenCalledOnce();
+    });
+
+    it('does not react on click when disabled', () => {
+        const clickListener = vi.fn();
```
SamuelAlev, line 22: The click listener isn't bound with a `onClick=` so it will never get called

packages/icons/src/createFondueIcon.ts
SamuelAlev, line 21: `export` might have been readded by mistake during a merge
```suggestion
type IconSizeAsNumber = 12 | 16 | 20 | 24 | 32;
```

packages/components/src/utilities/tests/styleUtilities.test.tsx
SamuelAlev, line 1: File doesn't contains any JSX, could be renamed to `.ts`

packages/components/src/components/Button/tests/Button.ct.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+test('should render in positive medium with an icon and text.', async ({ mount }) => {
+    const component = await mount(
+        <Button style="positive" size="medium" icon={<IconIcon />}>
+            {BUTTON_TEXT}
+        </Button>,
+    );
+    await expect(component).toBeVisible();
+    await expect(component).toHaveText(BUTTON_TEXT);
+    await expect(component.locator('css=span>svg')).toBeVisible();
```
SamuelAlev, line 58: Can you use the icon test id or name instead of `css=span>svg`? (Same for the other tests)

packages/components/src/components/Button/tests/Button.ct.tsx
```diff
@@ -0,0 +1,59 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+import { IconIcon } from '@frontify/fondue-icons';
+import { expect, test } from '@playwright/experimental-ct-react';
+
+import { Button } from '../Button';
+
+const BUTTON_TEXT = 'Frontify';
+
+test.use({ viewport: { width: 500, height: 500 } });
```
SamuelAlev, line 10:
```suggestion
```
Not needed

packages/components/src/utilities/tests/styleUtilities.test.tsx
noahwaldner: renamed

packages/components/src/components/Button/Button.tsx
noahwaldner: This will not allow us to access the inner component and it's type directly and break the docs generation in storybook. 

Apart from that i like to have this separated for improved readability

packages/components/src/components/Button/Button.tsx
noahwaldner: agree, that makes sense.

removed

packages/components/src/components/Button/tests/Button.ct.tsx
noahwaldner: changed

packages/components/src/components/Button/Button.tsx
SamuelAlev: It doesn't break actually, and for the readability, I would let the formatter do it and probably you could use the spread operator a bit more instead of destructuring everything.

I did the exercise (spreading, adding `defaultVariants` in the `sv`), and instead of the 103 lines, we have 68 (a bit more readable) lines.

Screenshot 2024-04-15 at 13 31 37

packages/components/src/components/Button/styles/iconStyles.ts
```diff
@@ -0,0 +1,151 @@
...
+            style: 'loud',
+            emphasis: 'strong',
+            class:
+                'tw-text-box-selected-strong-inverse ' +
+                'group-hover:tw-text-box-selected-strong-inverse' +
+                'group-active:tw-text-box-selected-strong-inverse tw-leading-none',
+        },
+    ],
+    defaultVariants: {
+        emphasis: 'strong',
```
SamuelAlev, line 149: missing `style` default

packages/components/src/components/Button/styles/textStyles.ts
```diff
@@ -0,0 +1,148 @@
...
+            style: 'loud',
+            emphasis: 'strong',
+            class:
+                'tw-text-box-selected-strong-inverse ' +
+                'group-hover:tw-text-box-selected-strong-inverse' +
+                'group-active:tw-text-box-selected-strong-inverse',
+        },
+    ],
+    defaultVariants: {
+        emphasis: 'strong',
```
SamuelAlev, line 146: missing `style` default

SamuelAlev (approved):
Left 2 comments, could you have a look before merging? Otherwise looks good!

PR #1801, opened 2024-04-11
feat: add Fondue Charts

SamuelAlev:


packages/fondue/package.json
```diff
@@ -63,7 +63,7 @@
         "test:browser": "cypress open --component",
         "test:file": "cypress run --component --spec",
         "typecheck": "tsc --noEmit",
-        "storybook": "storybook dev -p 6006",
+        "storybook": "storybook dev -p 6007",
```
noahwaldner, line 66: what is the reason to switch the ports between the components and legacy fondue?

packages/components/package.json
```diff
@@ -22,7 +22,7 @@
         "install:playwright": "playwright install --with-deps",
         "lint": "eslint .",
         "lint:fix": "eslint --fix .",
-        "storybook": "storybook dev -p 6007",
+        "storybook": "storybook dev -p 6006",
```
noahwaldner, line 25: what is the reason to switch the ports between the components and legacy fondue?

noahwaldner (approved):

PR #1802, opened 2024-04-12
feat: add build steps

SamuelAlev:
Tasks CU-86949whdd CU-86949wh8g


packages/fondue/src/components/Tooltip/Tooltip.cy.tsx
```diff
@@ -239,7 +239,7 @@ describe('Tooltip Component', () => {
         });
     });
 
-    describe('Pointer Events for touch devices', () => {
+    describe.skip('Pointer Events for touch devices', () => {
```
noahwaldner, line 242: why are we skipping this test?

noahwaldner (approved):

packages/fondue/src/components/Tooltip/Tooltip.cy.tsx
noahwaldner: please add a comment

PR #1818, opened 2024-04-18
fix: fix button padding when only showing icon

noahwaldner:


SamuelAlev (approved):

packages/components/src/components/Button/styles/buttonStyles.ts
```diff
@@ -164,6 +164,21 @@ export const buttonStyles = sv({
...
+        {
+            iconOnly: false,
+            size: 'medium',
+            class: 'tw-px-4',
+        },
+        {
+            iconOnly: false,
+            size: 'small',
+            class: 'tw-px-6',
+        }
```
SamuelAlev, line 181: lint error
```suggestion
        },
```

PR #1825, opened 2024-04-19
fix: icon handling in button

noahwaldner:


packages/components/src/components/Button/Button.tsx
```diff
@@ -49,27 +49,35 @@ export type ButtonProps = {
...
+     */
+    aspect?: ButtonAspect;
     /**
      * @default true
      */
     hugWidth?: boolean;
+    /**
+     * @deprecated please pass the Icon as a child and use the `aspect` prop
+     */
     icon?: ReactElement;
```
SamuelAlev, line 63: as it is the new component, can be removed, wdyt?

packages/components/src/components/Button/Button.tsx
```diff
@@ -78,31 +86,27 @@ export const Button = forwardRef<HTMLButtonElement | null, ButtonProps>(
...
+                    buttonStyles({
+                        size,
+                        style,
+                        ...props,
+                    }),
+                    textStyles({ style, ...props }),
+                    iconStyles({
+                        style,
+                        ...props,
+                    }),
```
SamuelAlev, line 99:
```suggestion
                    buttonStyles({ size, style, ...props }),
                    textStyles({ style, ...props }),
                    iconStyles({ style, ...props }),
```

packages/components/src/components/Button/Button.tsx
```diff
@@ -78,31 +86,27 @@ export const Button = forwardRef<HTMLButtonElement | null, ButtonProps>(
...
-                        className={iconStyles({
-                            iconSpacing: children && !hideLabel ? size : 'none',
-                            style,
-                            ...props,
-                        })}
-                    >
+                    <span data-test-id={`${dataTestId}-icon`}>
                         {cloneElement(icon, { size: buttonIconSizeMap[size] })}
                     </span>
                 )}
```
SamuelAlev, line 108: Same as the prop `icon`, if we agree to remove it, then should be removed too

packages/components/src/components/Button/Button.stories.tsx
```diff
@@ -20,38 +20,55 @@ const meta: Meta<typeof Button> = {
...
+export const TextLabelOnly: Story = {
     name: 'Text Label Only',
 };
 
 export const WithIcon: Story = {
     name: 'Icon Only',
     args: {
-        hideLabel: true,
-        icon: <IconIcon />,
+        icon: <IconIcon size={20} />,
```
SamuelAlev, line 50: should be `children` as `icon` will be removed / is deprecated

packages/components/src/components/Button/Button.tsx
noahwaldner: true, removed

packages/components/src/components/Button/Button.tsx
noahwaldner: adjusted

packages/components/src/components/Button/Button.tsx
noahwaldner: removed

packages/components/src/components/Button/Button.stories.tsx
noahwaldner: changed

SamuelAlev (approved):

PR #1827, opened 2024-04-19
feat: segmented control

SamuelAlev:
Task CU-8694bef9u CU-8694dbqht

TODO:
- ~Missing border around the segmented control~
- ~See if the disabled item makes sense or if just disabling the segmented control is fine~
- ~Prevent the untoggle~
- ~Emit event up~
- ~Fix hover background~
- ~Add tests~

Preview: https://deploy-preview-1827--fondue-components.netlify.app/?path=/docs/current_components-segmented-control--documentation


packages/components/MIGRATING.md
```diff
@@ -0,0 +1,60 @@
...
+Changes:
+
+-   The component has been renamed from `SegmentedControls` to its singular form `SegmentedControl`.
+-   The `SegmentedControl` component now has a `Root` and `Item` component to separate the control from its items.
+-   `items` array has been removed and replaced by `children` to follow the composition pattern.
+    -   `children` can have any kind of content (icons, text, etc.) and will be rendered as the content of the item.
+-   `activeItemId` has been replaced by `value`.
+-   `onChange` has been replaced by `onValueChange`.
+-   `disabled` has been removed from individual items and moved to the `Root` component.
+    -   If you need to disable per item, you will need to conditionally render the item.
```
noahwaldner, line 27: I'm not sure about this behaviour. 

From what i understand there is a need that a option is visible abut not selectable. I believe there is no way to do this here?

It is not just about removing an item, but making it un-selectable.

What we also need to watch out, is what happens to the focus when the currently focused item becomes disabled or is removed

SamuelAlev: I already asked Johanna as the design doesn't show any "item disabled" state and it seems it was never a desired state / old state which should be removed.

noahwaldner: I remember this was something we implemented just a few weeks ago, so it was defenitely a need by a team, we should double check this

noahwaldner (approved):

PR #1833, opened 2024-04-23
fix: add default value for disabled prop

noahwaldner:


.changeset/orange-ligers-sniff.md
```diff
@@ -0,0 +1,5 @@
+---
+"@frontify/fondue-components": patch
+---
+
+fix: add default value for diabled
```
SamuelAlev:
```suggestion
fix: add default value for `disabled`
```

PR #1849, opened 2024-05-02
feat: add `TextInput`

SamuelAlev:
Task CU-8694fqqvv CU-8693p1c7n

TODO:
- ~Add tests~
- ~Add sizes (check with Johanna)~ (new design doesn't have it)
- ~Add border dotted (check with Johanna if needed)~ (new design doesn't have it)
- ~What about `clearable` and `copyable`?~ (new design doesn't have it builtin)
- ~Test accessibility and add missing props~
- ~Rename slot `side` to `name`~

Preview: https://deploy-preview-1849--fondue-components.netlify.app/?path=/docs/current_components-text-input--documentation

Syntax:
Simple usage:
```tsx
<TextInput type="password" />
// Or the alias is also working
<TextInput.Root type="password" />
```

Complex usage with a `Button` with an `Icon` and a (fake) `Dropdown` component on the right side:
```tsx
<TextInput.Root {...args}>
    <TextInput.Slot side="right">
        <Button aspect="square" rounding="full" emphasis="weak" size="small">
            <IconIcon size={16} />
        </Button>

        <Dropdown.Root>
            <Dropdown.Trigger>
                <Button aspect="square" rounding="full" emphasis="weak" size="small">
                    <IconIcon size={16} />
                </Button>
            </Dropdown.Trigger>
            <Dropdown.Content>
                    <Dropdown.Item label="foo" />
                    <Dropdown.Item label="bar" />
            </Dropdown.Content>
        </Dropdown.Root>
    </TextInput.Slot>
</TextInput.Root>
```


packages/components/MIGRATING.md
```diff
@@ -58,3 +61,99 @@ Changes:
...
+        ```tsx
+        <TextInput.Root>
+            {/* The following replace `decorator` prop */}
+            <TextInput.Slot name="left">
+                <IconIcon />
+            </TextInput.Slot>
+
+            {/* The following replace `clearable`, `copyable` and `extraActions` props */}
+            <TextInput.Slot name="right">
+                <Button onClick={(event) => {}}>
```
noahwaldner, line 91: onPress

packages/components/MIGRATING.md
```diff
@@ -58,3 +61,99 @@ Changes:
...
+                    <IconClear />
+                </Button>
+
+                <Button onClick={(event) => {}}>
```
noahwaldner, line 95: onPress

packages/components/src/components/TextInput/TextInput.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+const TextFieldRoot = forwardRef<HTMLInputElement, TextInputProps>(
+    ({ children, className, status = 'neutral', 'data-test-id': dataTestId, ...inputProps }, ref) => {
+        return (
+            <div
+                className={cn(rootStyles, className)}
+                data-disabled={inputProps.disabled}
+                data-read-only={inputProps.readOnly}
+                data-status={status}
+                data-test-id={dataTestId ?? 'fondue-text-input'}
```
noahwaldner, line 52: can we define the default during the prop spreading?

packages/components/src/components/TextInput/TextInput.tsx
```diff
@@ -0,0 +1,107 @@
...
+            >
+                <div className={loadingStatusStyles} data-test-id="fondue-text-input-loader" />
```
noahwaldner, line 54: plese use string concatination in case a custom test id is passed

packages/components/src/components/TextInput/TextInput.tsx
```diff
@@ -0,0 +1,107 @@
...
+    'aria-labelledby'?: string;
+    'aria-describedby'?: string;
+};
+
+const TextFieldRoot = forwardRef<HTMLInputElement, TextInputProps>(
+    ({ children, className, status = 'neutral', 'data-test-id': dataTestId, ...inputProps }, ref) => {
+        return (
+            <div
+                className={cn(rootStyles, className)}
+                data-disabled={inputProps.disabled}
```
noahwaldner, line 49: why is this needed? couldn't this be determined by has:input:disabled (pseudo code)

If we wanna explicitly define those things, we should use tailwind-variants

packages/components/src/components/TextInput/TextInput.tsx
```diff
@@ -0,0 +1,107 @@
...
+                data-read-only={inputProps.readOnly}
```
noahwaldner, line 50: why is this needed? couldn't this be determined by has:input:read-only (pseudo code)

If we wanna explicitly define those things, we should use tailwind-variants

packages/components/src/components/TextInput/styles/textInputStyles.ts
```diff
@@ -0,0 +1,37 @@
...
+    // Success and error status styles
+    'data-[status="success"]:tw-border-text-positive data-[status="error"]:tw-border-text-negative ';
+
+export const inputStyles =
+    'tw-w-full tw-bg-transparent tw-flex tw-items-center [text-align:inherit] tw-indent-3 tw-outline-none tw-rounded-[calc(var(--radius)_-_var(--line-width))] ' +
+    // Remove border-radius and text-indent/padding on the left if there’s a left-side slot
+    'has-[~_div:not([data-name="right"])]:tw-indent-0 has-[~_:not([data-name="right"])]:tw-pl-0 has-[~_:not([data-name="right"])]:tw-rounded-l-[0px] ' +
+    // Remove border-radius and padding on the right if there’s a right-side slot
+    'has-[~_[data-name="right"]]:tw-pr-0 has-[~_[data-name="right"]]:tw-rounded-r-[0px] ' +
+    '[~_:not([data-name="right"])_~_:not([data-name="left"])]:tw-pr-0 [~_:not([data-name="right"])_~_:not([data-name="left"])]:tw-rounded-r-[0px]' +
```
noahwaldner, line 18: use tw-rounded-r-none instead of arbitary value

packages/components/src/components/TextInput/styles/textInputStyles.ts
```diff
@@ -0,0 +1,37 @@
...
+    // Readonly or disabled styles
+    'read-only:tw-cursor-text disabled:tw-cursor-text read-only:tw-text-text-disabled disabled:tw-text-text-disabled ';
+
+export const slotStyles =
+    'tw-peer tw-shrink-0 tw-flex tw-items-center tw-cursor-text tw-gap-2 tw-pl-3 tw-pr-2 ' +
+    // Left slot
+    '[&:not([data-name="right"])]:-tw-order-1 [&:not([data-name="right"])]:tw-ml-px [&:not([data-name="right"])]:tw-mr-0 ' +
```
noahwaldner, line 25: why not target data-name="left" ?

packages/components/src/components/TextInput/styles/textInputStyles.ts
```diff
@@ -0,0 +1,37 @@
...
+    // Right slot
+    'data-[name="right"]:tw-order-[0] data-[name="right"]:tw-ml-0 data-[name="right"]:tw-mr-px ' +
+    // 2 slots without any name should be implicitly ordered
+    '[&:not([data-name="right"])_~_:not([data-name="left"])]:tw-order-[0] [&:not([data-name="right"])_~_:not([data-name="left"])]:tw-ml-0 [&:not([data-name="right"])_~_:not([data-name="left"])]:tw-mr-px ' +
+    // Readonly or disabled color styles
+    'group-data-[read-only="true"]:tw-text-text-disabled group-data-[disabled="true"]:tw-text-text-disabled ';
+
+export const loadingStatusStyles =
+    'group-data-[status="loading"]:tw-flex tw-hidden tw-absolute tw-size-6 -tw-top-3 -tw-right-3 tw-pointer-events-none ' +
```
noahwaldner, line 34: why not load these styles conditionally in JS instead of setting the data attribute?

noahwaldner:
I think we need to discuss if the benefits of the data-attribute + css selector are worth it to take some reduced readability in. I feel in some areas this could be a bit prone to bugs and reduce development speed in the future

noahwaldner (changes requested):

packages/components/src/components/TextInput/TextInput.tsx
SamuelAlev: - div shouldn't have `disabled`
- Side-effect of the wrapper based on the input instead of the other way around
- Easier debugging

packages/components/src/components/TextInput/TextInput.tsx
SamuelAlev: Same as previous comment

packages/components/src/components/TextInput/styles/textInputStyles.ts
SamuelAlev: Could be `div data-name="left"` and `div` (implicit left), `:not(data-name="right")` is easier

packages/components/src/components/TextInput/TextInput.stories.tsx
```diff
@@ -0,0 +1,158 @@
...
+    },
+    args: {
+        placeholder: 'Placeholder',
+        defaultValue: 'Hello from Frontify',
+        onChange: action('onChange'),
+        onBlur: action('onBlur'),
+        onFocus: action('onFocus'),
+    },
+    render: (args) => {
+        // Used to get the correct component in the Storybook for the simple cases (`TextInput` instead of `TextInput.Root`)
```
noahwaldner, line 34: this is only used once, why not define this in the story and use one of the more commonly used ones as the component default?

packages/components/src/components/TextInput/TextInput.tsx
noahwaldner: i thought during our discussion yesterday we decided that for these specifically defined attributes we decided to use tailwindVariants?

I'm abolutely okay with using the css selectors for the things that are there anyways or only possible that way (disabled input or slot name) but for the ones specifically defined, pls use variants or let's discuss again.

packages/components/src/components/TextInput/TextInput.tsx
noahwaldner: see previous comment

packages/components/src/components/TextInput/TextInput.tsx
```diff
@@ -42,16 +43,20 @@ type TextInputProps = {
...
             <div
                 className={cn(rootStyles, className)}
                 data-disabled={inputProps.disabled}
                 data-read-only={inputProps.readOnly}
                 data-status={status}
-                data-test-id={dataTestId ?? 'fondue-text-input'}
+                data-test-id={dataTestId}
             >
-                <div className={loadingStatusStyles} data-test-id="fondue-text-input-loader" />
+                {status === 'loading' ? <div className={loadingStatusStyles} data-test-id="loader" /> : null}
```
noahwaldner, line 58: wouldn't it make sense to use `${dataTestId}-loader`?

noahwaldner (changes requested):

packages/components/src/components/TextInput/TextInput.tsx
SamuelAlev: Internal use only, external usage shouldn't know about this test id.

packages/components/src/components/TextInput/TextInput.stories.tsx
SamuelAlev: Prefer to have the "simple" and most used case as the default `render`

packages/components/src/components/TextInput/TextInput.tsx
SamuelAlev: Fixed, pushed in the wrong branch

packages/components/src/components/TextInput/TextInput.tsx
SamuelAlev: same

noahwaldner (approved):

PR #1856, opened 2024-05-06
ci: should use pnpm version defined in package.json

ragi96:
So [this](https://github.com/Frontify/fondue/actions/runs/8968679284/job/24628597757?pr=1855) (the ci) should not fail anymore, as it should use the pnpm version defined in the `package.json`

After discussion with @noahwaldner also bumping `pnpm` to `9.1.0`


noahwaldner (approved):

noahwaldner:
LGTM! @ragi96

PR #1857, opened 2024-05-07
chore: bump immer to 10

ragi96:
CU-869483phw

If we bump `immer` to latest within fondue we can finally remove the override of it within `web-app`

[Release Notes with Breaking Changes](https://github.com/immerjs/immer/releases/tag/v10.0.0)
- No breaking change is related to how we use `immer` within the tree component, please confirm @anxobotana 

Dependant Components
- RTE ✅ 
- Tree ✅  thanks @anxobotana
- LinkChooser ✅


SamuelAlev:
@ragi96 Doesn't it breaks the link chooser? (it uses immer with xstate)

ragi96:
> @ragi96 Doesn't it breaks the link chooser? (it uses immer with xstate)

Good catch didn't checked that one 😓 

In storybook it's okay imo, will check it next tuesday directly in web-app if everything is fine 🤞

ragi96:
> @ragi96 Doesn't it breaks the link chooser? (it uses immer with xstate)

Works, just noticed one UI issue, but its not related to the immer bump.

SamuelAlev (approved):

PR #1867, opened 2024-05-13
fix: loading bar 0%

SamuelAlev:
Task CU-8694h0bzt

### DIFF
--- a/packages/components/src/components/LoadingBar/LoadingBar.tsx
+++ b/packages/components/src/components/LoadingBar/LoadingBar.tsx
@@ -55,7 +55,13 @@ export const LoadingBar = forwardRef<ElementRef<typeof ProgressRadixPrimitive.Ro
                 aria-busy={value !== max}
                 value={value}
                 max={max}
-                style={value ? ({ '--loading-bar-value': `${value}%` } as CSSProperties) : {}}
+                style={
+                    {
+                        '--loading-bar-value': value,
+                        '--loading-bar-max': max,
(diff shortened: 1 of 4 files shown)


packages/components/src/components/LoadingBar/LoadingBar.tsx
```diff
@@ -55,7 +55,13 @@ export const LoadingBar = forwardRef<ElementRef<typeof ProgressRadixPrimitive.Ro
                 aria-busy={value !== max}
                 value={value}
                 max={max}
-                style={value ? ({ '--loading-bar-value': `${value}%` } as CSSProperties) : {}}
+                style={
+                    {
+                        '--loading-bar-value': value,
+                        '--loading-bar-max': max,
+                        '--loading-bar-proportion': 'calc(var(--loading-bar-value) / var(--loading-bar-max))',
```
noahwaldner, line 62: could that be in the styles file instead of being a separate css variable

noahwaldner (approved):

packages/components/src/components/LoadingBar/LoadingBar.tsx
SamuelAlev: Could be easily in the TW arbitrary class, but I use it to test it is the correct value we are passing in, else we don't have a way to test it "for real"

SamuelAlev: If it bother you, we can make another PR to inline it and see how we can change the tests

PR #1884, opened 2024-05-28
fix: text input text color

SamuelAlev:
- fix `TextInput` text color
- add `showCloseButton` to `Dialog`
- delete unused class in `Flyout` and `Dialog`


noahwaldner (approved):

noahwaldner:
@SamuelAlev this just came to my mind:

Is it a smart idea to have `showCloseButton` default to true? Isn't this somehow counter intuitive?

Also in the Flyout Header it defaults to false

PR #1886, opened 2024-05-29
fix: tailwind config: override the entire screens key for the right order of breakpoints

bojangles-m:
If you want to add an additional small breakpoint, you can’t use extend because the small breakpoint would be added to the end of the breakpoint list
https://tailwindcss.com/docs/screens#adding-smaller-breakpoints

### DIFF
--- a/.changeset/seven-actors-greet.md
+++ b/.changeset/seven-actors-greet.md
@@ -1,0 +1,6 @@
+---
+"@frontify/fondue-components": minor
+"@frontify/fondue": minor
+---
+
+tailwind config: override the entire `screens` key to add smaller xs key at the beginning of the list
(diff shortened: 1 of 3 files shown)


.changeset/seven-actors-greet.md
```diff
@@ -0,0 +1,6 @@
+---
+"@frontify/fondue-components": minor
+"@frontify/fondue": minor
+---
+
+tailwind config: override the entire `screens` key to add smaller xs key at the beginning of the list
```
SamuelAlev, line 6:
```suggestion
fix: make the `xs:` breakpoint work correctly
```

SamuelAlev (approved):

PR #1891, opened 2024-05-31
fix: combine thirdpartystyles into one file

noahwaldner:
This change combines the thirdpartystyles into one file so it can be parsed correctly by postcss-import

postcss-import requires the @import to be at the beginning of the file


.changeset/fuzzy-impalas-tease.md
```diff
@@ -0,0 +1,5 @@
+---
+"@frontify/fondue": patch
+---
+
+fix: improve style exports for thirdpartystyles
```
SamuelAlev, line 5:
```suggestion
fix: combine third party styles into one file to prevent PostCSS warning
```

SamuelAlev (approved):

PR #1899, opened 2024-06-04
feat(RTE): add custom class to column break plugin

fulopdaniel:
This PR adds `customClass` to the `ColumnBreakPlugin`.

With this addition, you can create the plugin like this:

`new BreakAfterPlugin({ customClass: 'tw-columns-1 @sm:tw-columns-2' })`

This allows more fine-grained controls when needed while keeping the original implementation as well.

### DIFF
--- a/packages/fondue/src/components/RichTextEditor/Plugins/ColumnBreakPlugin/types.ts
+++ b/packages/fondue/src/components/RichTextEditor/Plugins/ColumnBreakPlugin/types.ts
@@ -1,0 +1,12 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+import { type PluginProps } from '@components/RichTextEditor/Plugins/Plugin';
+
+export type BreakAfterPluginProps = PluginProps & {
+    columns?: number;
+    gap?: string | number;
+    /**
+     * If provided the `columns` and `gap` properties will be ignored
+     */
+    customClass?: string;
+};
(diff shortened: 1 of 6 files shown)


packages/fondue/src/components/RichTextEditor/Plugins/ColumnBreakPlugin/types.ts
```diff
@@ -0,0 +1,12 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+import { type PluginProps } from '@components/RichTextEditor/Plugins/Plugin';
+
+export type BreakAfterPluginProps = PluginProps & {
+    columns?: number;
+    gap?: string | number;
```
SamuelAlev, line 7: What about having a type like this?

```tsx
import type tailwindConfig from '<root>/tailwind.config';

type Breakpoint = keyof typeof tailwindConfig.theme.screens;
export type Responsive<TValue> = { [key in Breakpoint]?: TValue } | TValue;

export type BreakAfterPluginProps = PluginProps & {
    columns?: Responsive<number>;
    gap?: Responsive<string | number>;
```

fulopdaniel: What we aim to have now is container queries, that are responsive based on the container, not the traditional viewport. For that we will [add the container query plugin to brand-sdk](https://github.com/Frontify/brand-sdk/pull/949) as that is considered block-related logic.

noahwaldner (approved):

PR #1906, opened 2024-06-07
refactor: refactor dialog to use scss

noahwaldner:
- refactor styles to scss
- add mixin for sm media query


packages/components/src/components/Dialog/styles/dialog.module.scss
SamuelAlev, line 1: I just opened the first story (https://deploy-preview-1906--fondue-components.netlify.app/?path=/story/current_components-dialog--with-full-layout) and there are inconsistencies with main (https://fondue-components.frontify.com/?path=/story/current_components-dialog--with-full-layout)

Please check over one by one.

packages/components/package.json
```diff
@@ -71,6 +71,7 @@
         "@storybook/addon-links": "^8.1.5",
         "@storybook/blocks": "^8.1.5",
         "@storybook/manager-api": "^8.1.5",
+        "@storybook/preset-scss": "^1.0.3",
```
SamuelAlev, line 74: vite support scss by default and the plugin was necessary when storybook was using webpack instead of vite

noahwaldner: removed, it was a chache issue

noahwaldner:
Styles adjusted to match tw version

packages/components/package.json
SamuelAlev, line 74: Still there

packages/components/package.json
```diff
@@ -93,7 +94,6 @@
         "postcss-import": "^16.1.0",
         "prettier": "^3.3.0",
         "react": "^18.3.1",
-        "react-dom": "^18.3.1",
```
SamuelAlev, line 96: Please revert back

PR #1931, opened 2024-06-25
fix: add missing prop showCloseButton

noahwaldner:
CU-8694wyg23


packages/components/src/components/Dialog/Dialog.tsx
```diff
@@ -68,6 +68,10 @@ export type DialogContentProps = {
 export type DialogTriggerProps = { children?: ReactNode; 'data-test-id'?: string };
 
 export type DialogHeaderProps = {
+    /**
+     * Show a close button in the header
```
SamuelAlev, line 72: default true?

SamuelAlev (approved):

PR #1932, opened 2024-06-25
docs: add story for external control

noahwaldner:
CU-8694wx7en


packages/components/src/components/Dialog/Dialog.stories.tsx
```diff
@@ -291,3 +292,30 @@ export const MobileView: Story = {
         children: 'Hello World',
     },
 };
+
+export const ExternalControl: Story = {
```
SamuelAlev, line 296: To match Dropdown and Select story name
```suggestion
export const ControlledComponent: Story = {
```

SamuelAlev (approved):

PR #1953, opened 2024-07-12
fix: prevent firefox from overriding outline class with global pseudo class def

jose-costa-frontify:
image

image

This here global style is given priority over the class declaration that assigns the proper outline. This is a firefox specific declaration that sets `outline:auto` and seems to be coming from [tailwind's normalization styles](https://cdn.frontify.com/css/spa-frontify-index.7dcb291c092af233ee86.css). The `currentcolor` definition basically inherits from the button's text color, which in some cases will match the background color where the button is located (ie. if the button is black with white text on white background) giving the impression that no outline is present. By making the class declaration more specific the issue is fixed.

[CU-869519gyp](https://app.clickup.com/t/869519gyp)

### DIFF
--- a/packages/components/src/utilities/focusStyle.ts
+++ b/packages/components/src/utilities/focusStyle.ts
@@ -1,4 +1,4 @@
 /* (c) Copyright Frontify Ltd., all rights reserved. */
 
 export const FOCUS_OUTLINE =
-    'focus-visible:tw-outline has-[[data-show-focus-ring=true]]:tw-outline tw-outline-4 tw-outline-offset-2 tw-outline-blue';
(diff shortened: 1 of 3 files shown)


packages/components/src/utilities/focusStyle.ts
```diff
@@ -1,4 +1,4 @@
 /* (c) Copyright Frontify Ltd., all rights reserved. */
 
 export const FOCUS_OUTLINE =
-    'focus-visible:tw-outline has-[[data-show-focus-ring=true]]:tw-outline tw-outline-4 tw-outline-offset-2 tw-outline-blue';
+    'focus-visible:tw-outline has-[[data-show-focus-ring=true]]:tw-outline tw-outline-4 tw-outline-offset-2 tw-outline-blue focus-visible:tw-outline-blue'; // second declaration of tw-outline-blue is to assure that in firefox the outline isn't overriden by a global definition of :-moz-focusring which is coming from tailwinds normalization styling
```
SamuelAlev, line 4: IMO comment not necessary.
Also as we are moving to SCSS, could you add it to `packages/components/src/utilities/focusStyle.module.scss`?

noahwaldner (approved):

PR #1963, opened 2024-07-22
feat: export focus outline util

jose-costa-frontify:
Clickup task: [86952w1jf](https://app.clickup.com/t/86952w1jf)


packages/components/src/index.ts
SamuelAlev, line 1: We are going the SCSS way, meaning this file will be deleted when all components are SCSS instead of TW

SamuelAlev (changes requested):

PR #1965, opened 2024-07-22
feat: table - add column align props and custom title node for column headers

joseazevedo88:
### Table columns now have a `titleNode` prop that can be used to pass components as the column header.

Also updated stories and tests

### DIFF
--- a/packages/fondue/src/components/Table/Table.cy.tsx
+++ b/packages/fondue/src/components/Table/Table.cy.tsx
@@ -13,6 +13,18 @@ const TABLE_COLUMNS: Column[] = [
     { name: 'Regions', key: 'regions' },
 ];
 
+const TABLE_COLUMNS_WITH_ALIGNMENT: Column[] = [
+    { name: 'User', key: 'user', align: 'right' },
+    { name: 'Active Sessions', key: 'activeSessions', sortable: true, align: 'right' },
+    { name: 'Regions', key: 'regions', align: 'right' },
+];
+
+const TABLE_COLUMNS_WITH_TITLE_NODE: Column[] = [
+    { name: 'User', key: 'user', titleNode: <span>User</span> },
+    { name: 'Active Sessions', key: 'activeSessions', sortable: true, titleNode: <span>Active sessions</span> },
+    { name: 'Regions', key: 'regions', titleNode: <span className="tw-bg-red-10">Regions</span> },
+];
+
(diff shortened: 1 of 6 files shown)


packages/fondue/src/components/Table/Table.tsx
```diff
@@ -73,7 +78,11 @@ const mapToTableAriaProps = (columns: Column[], rows: Row[], hasSort = false): T
             <TableHeader key="table-header" columns={columns}>
                 {(column) => {
                     const allowsSorting = !!(column.sortable && hasSort);
-                    return <AriaColumn allowsSorting={allowsSorting}>{column.name}</AriaColumn>;
+                    return (
+                        <AriaColumn allowsSorting={allowsSorting} {...{ align: column.align }}>
```
jose-costa-frontify, line 82: is this different than
```suggestion
                        <AriaColumn allowsSorting={allowsSorting} align={column.align}>
```
?

packages/fondue/src/components/Table/Table.cy.tsx
```diff
@@ -115,7 +127,7 @@ describe('Table Component', () => {
         }
     });
 
-    it('should render single-select table', () => {
+    it.skip('should render single-select table', () => {
```
jose-costa-frontify, line 130: leftover from testing?

packages/fondue/src/components/Table/Table.tsx
joseazevedo88: Only in terms of types. That component(imported from a library) has a determinate set of props, in this case like `allowSorting`. But if I want to pass custom props to it it only worked via spread

packages/fondue/src/components/Table/Table.cy.tsx
joseazevedo88: Before, only a single test was running inside this test file with `it.only`. I ran them all and saw that some of them were always failing and decided to skip those as I don't have time to fix them now. In the end, the coverage is greater now because some good tests were being skipped

jose-costa-frontify: understood, thank you!

packages/fondue/src/components/Table/Table.tsx
jose-costa-frontify: interesting, cool TS loophole

jose-costa-frontify (approved):
LGTM

PR #1970, opened 2024-07-24
fix: remove accordion's empty aria-labelledby, add aria-hidden to icon generator

jose-costa-frontify:
Clickup task: [2523021/TASK-11732](https://app.clickup.com/t/2523021/TASK-11732)

This should fix an array of issues for screen readers stumbling into icons and confusing users but it was done in light of an issue reported in respect to the accordion.


SamuelAlev (approved):
Missing changeset but ✅

PR #1971, opened 2024-07-25
fix: tabs component not allowing users to focus on non selected tabs

ghost:
Clickup id: [869572jd3](https://app.clickup.com/t/869572jd3)

This must've been a misunderstanding of what `tabIndex` does


SamuelAlev (approved):

SamuelAlev:
Approved but make sure it is correct according to https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tab_role#keyboard_interactions

PR #1973, opened 2024-07-25
feat: refactor color picker

noahwaldner:
TASK-9413


packages/components/src/components/ColorPicker/ColorPicker.tsx
```diff
@@ -0,0 +1,100 @@
...
+export const ColorPickerRoot = ({
+    children,
+    currentColor = DEFAULT_COLOR,
+    onColorChange = () => {},
+    defaultFormat = DEFAULT_FORMAT,
+    ...props
+}: ColorPickerProps) => {
+    const [currentFormat, setCurrentFormat] = useState<ColorFormat>(defaultFormat);
+
+    const handleColorChange = useCallback(
```
jose-costa-frontify, line 42: if this function received a `onColorChange` callback as well you could pull it out of this closure entirely and avoid having to memoize it. Since you have to receive `currentFormat` anyway it might not be a bad idea to add that 3rd param

packages/components/src/components/ColorPicker/ColorValueInput.tsx
```diff
@@ -0,0 +1,188 @@
...
+    setCurrentFormat?: (format: ColorFormat) => void;
+};
+
+export const ColorValueInput = ({
+    currentColor = DEFAULT_COLOR,
+    onColorChange = () => {},
+    currentFormat = DEFAULT_FORMAT,
+    setCurrentFormat = () => {},
+}: ColorValueInputProps) => {
+    const [hexColorValue, setHexColorValue] = useState<string>(rgbColorToHex(currentColor));
```
jose-costa-frontify, line 43: mayhaps?

```suggestion
    const hexColorValue = useMemo(()=> rgbColorToHex(currentColor), [currentColor]);
```

packages/components/src/components/ColorPicker/ColorPicker.tsx
```diff
@@ -0,0 +1,91 @@
...
+import { type ColorFormat, type RgbaColor } from './types';
+import { DEFAULT_COLOR, DEFAULT_FORMAT, getColorWithName } from './utils';
+
+type ColorPickerProps = {
+    /**
+     * The children of the color picker component. This can contain multiple `ColorPicker.Values` or `ColorPicker.Gradient` components.
+     */
+    children?: ReactNode;
+    /**
+     * The active color in the color picker
```
jose-costa-frontify, line 19: maybe we can add a default here?

packages/components/src/components/ColorPicker/ColorValueInput.tsx
```diff
@@ -0,0 +1,177 @@
...
+
+    useEffect(() => {
+        setHexColorValue(rgbColorToHex(currentColor));
+    }, [currentColor]);
+
+    return (
+        <div className={styles.inputs} data-test-id={dataTestId} ref={forwardedRef}>
+            <div className={styles.colorFormatInput}>
+                <Select
+                    aria-label="Select an item"
```
jose-costa-frontify, line 68: if multiple of these are on the page using the same `aria-label` could be problematic, this should probably come from a prop

packages/components/src/components/ColorPicker/styles/customColorPicker.module.scss
```diff
@@ -0,0 +1,59 @@
...
+.gradientInput {
+    width: 100%;
+
+    .reactColorful {
+        width: 100%;
+    }
+}
+
+//remove arrows from number inputs
+.valueInput input::-webkit-outer-spin-button,
```
jose-costa-frontify, line 51: why not nest these?

packages/components/src/components/ColorPicker/utils.ts
```diff
@@ -0,0 +1,73 @@
...
+    }
+    return `rgba(${color.red}, ${color.green}, ${color.blue}, ${color.alpha || 1})`;
+};
+
+export const isValidHexColor = (color: string): boolean => {
+    const hexRegex = /^([\dA-Fa-f]{3}){1,2}$/;
+    return color.length > 2 && hexRegex.test(color);
+};
+
+export const hexColorToRgba = (hex: string): RgbaColor => {
```
jose-costa-frontify, line 20: idk about you guys but I'm a big fan of annotating functions like mad

```suggestion
/**
 * Converts a hexadecimal color code to an RGBA color object. This function accepts
 * both three-digit and six-digit hex codes without the '#' symbol.
 * @param {string} hex - The hex code, either 3 or 6 characters long. Characters should be from the set [0-9A-Fa-f].
 * @returns {RgbaColor} 
 * @example
 * hexColorToRgba('FFF'); // { red: 255, green: 255, blue: 255, alpha: 1 }
 * @example
 * hexColorToRgba('FF5733'); // { red: 255, green: 87, blue: 51, alpha: 1 }
 * @example
 * hexColorToRgba('XYZ'); // { red: 0, green: 0, blue: 0, alpha: 0 }
 */
export const hexColorToRgba = (hex: string): RgbaColor => {
```

jose-costa-frontify (changes requested):
Looking good just a couple potential issues

packages/components/src/components/ColorPicker/ColorPicker.tsx
noahwaldner: not necessary, as we still need the defaults in the subcomponents. If we make the prop required in the subcomponents, typescript will complain because it doesn't know about the slot that injects these.

packages/components/src/components/ColorPicker/ColorPickerInput.tsx
```diff
@@ -0,0 +1,69 @@
...
+                />
+            ) : (
+                <IconDroplet size={16} />
+            )}
+
+            <span id={colorNameId} className={styles.colorName}>
+                {currentColor?.name}
+            </span>
+            {onClear && (
+                <button type="button" aria-label="Clear color" className={styles.clear}>
```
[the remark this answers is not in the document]
noahwaldner: added

packages/components/src/components/ColorPicker/styles/customColorPicker.module.scss
noahwaldner: adjusted!

packages/components/src/components/ColorPicker/utils.ts
noahwaldner: added jsdoc

packages/components/src/components/ColorPicker/ColorValueInput.tsx
jose-costa-frontify: are you sure? i feel like using a screen reader on a form with a few of this would yield many readouts of the same text. Also, it might be worth considering adding `aria-labelledby` if a dev wants to pair this with his own label

packages/components/src/components/ColorPicker/ColorValueInput.tsx
```diff
@@ -0,0 +1,177 @@
...
+                    placeholder="Select an item"
```
jose-costa-frontify, line 69: I feel like this should be coming from a prop set by the dev so as to allow him to pass custom placeholders, specially translations

jose-costa-frontify (approved):
LGTM

PR #1982, opened 2024-07-29
feat(bar chart): tooltip changes [CU-8695881ux]

joseazevedo88:
### Design
image

### Strorybook
image

#### Also fixed an issue where toggling the horizontal field in the bar chart stories wasn't working

### DIFF
--- a/packages/charts/src/components/BarChart/components/Axes.tsx
+++ b/packages/charts/src/components/BarChart/components/Axes.tsx
@@ -1,10 +1,11 @@
 /* (c) Copyright Frontify Ltd., all rights reserved. */
 
+import { Axis as VisxAxis } from '@visx/xychart';
+import { type Dispatch, type SetStateAction } from 'react';
+
 import { useBandTicks } from '@components/BarChart/components/hooks/useBandTicks';
 import { useRotatedLabel } from '@components/BarChart/components/hooks/useRotatedLabel';
 import { type LabelFormatter, type ValueFormatter } from '@components/common/types';
-import { Axis as VisxAxis } from '@visx/xychart';
-import { type Dispatch, type SetStateAction } from 'react';
 
 import { truncateTextLabel } from '../helpers';
 
--- a/packages/charts/src/components/BarChart/components/Axes.tsx
+++ b/packages/charts/src/components/BarChart/components/Axes.tsx
@@ -38,7 +38,7 @@ export const Axes = ({
     valueFormatter,
     labelFormatter,
 }: AxesProps) => {
-    const formatLabel = (label: string) => truncateTextLabel(labelFormatter(label).toString());
+    const formatLabel = (label: string | number) => truncateTextLabel(labelFormatter(label.toString()));
     const angle = useRotatedLabel(horizontal, updateMaxLabelHeight, updateFirstLabelOverflowsBy, formatLabel);
     const filteredTicks = useBandTicks(horizontal, updateBandScaleTicks);
     const textAnchor = getTextAnchorProp(horizontal, angle);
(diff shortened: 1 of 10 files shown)


joewoodson (changes requested):
Code looks good, and the stories are working as expected, nice work! But I wonder if the name of the new prop could be made clearer -- what about `valueContext`, as the new content helps to contextualize the value and clarify its meaning?

joewoodson (approved):

joseazevedo88:
@Frontify/frontend-platform I need a review/approval here, please. Thanks

packages/charts/src/components/BarChart/components/Axes.tsx
```diff
@@ -38,7 +38,7 @@ export const Axes = ({
     valueFormatter,
     labelFormatter,
 }: AxesProps) => {
-    const formatLabel = (label: string) => truncateTextLabel(labelFormatter(label));
+    const formatLabel = (label: string) => truncateTextLabel(labelFormatter(label).toString());
```
noahwaldner, line 41: Why is this necessary? 

The typing of laybelFormatter is `export type LabelFormatter = (label: string) => string;` the value is already a string

packages/charts/src/components/common/components/Tooltip/Tooltip.spec.tsx
```diff
@@ -0,0 +1,66 @@
...
+                },
+            },
+            tooltipOpen: true,
+        });
+        vi.mocked(usePositions).mockReturnValue({
+            tooltipPosition: { tooltipLeft: 20, tooltipTop: 30 },
+            glyphProps: [],
+        });
+        const { getByText } = render(
+            <Tooltip crossHairStyle="line" colorAccessor={(str) => str} valueContextBySeries={['%', 'ºC', 'F']} />,
```
noahwaldner, line 59: if there is a case where no proper `colorAccessor` function is passed in here, it would make sense to make this optional and define a default in the Tooltip.

packages/charts/src/components/common/components/Tooltip/Tooltip.tsx
```diff
@@ -91,8 +93,11 @@ export const Tooltip = ({
         tooltipContext.tooltipData?.datumByKey,
         childSumLabel,
     );
+    const entriesWithFormattedSeries = valueContextBySeries
```
noahwaldner, line 96: You could move this into getTooltipEntries and potentially even memoize that function. That way you have one function that returns your formatted tooltip content

joseazevedo88: Moved, didn't do the memo though, no reason to do it unless it's slow as it also has a overhead cost.

packages/charts/src/components/common/components/Tooltip/Tooltip.spec.tsx
joseazevedo88: Made it optional with a fallback

packages/charts/src/components/BarChart/components/Axes.tsx
joseazevedo88: Fixed the typing, the `formatLabel` param is of type `string | number` in reality because of the axis toggle

packages/charts/src/components/common/components/Tooltip/Tooltip.tsx
jose-costa-frontify: nothing is slow on a powerful enough cpu 😛 `.reverse` is efficient because it is performed in-place but i would suggest the memory cost of `useMemo` are more than surpassed by the memory spent spreading that array every time the component re-renders. Then it's off waiting for the garbage collector, potentially piles of lonely arrays with nothing pointing to them

jose-costa-frontify (approved):
LGTM

PR #1983, opened 2024-07-30
fix: remove outline on hover in dropdown

noahwaldner:


packages/components/src/components/Dropdown/styles/dropdown.module.scss
```diff
@@ -53,6 +49,10 @@
         order: 2;
         margin-left: auto;
     }
+
+    &:not(&:hover) {
```
SamuelAlev, line 53: If we leave the mouse on the item then switch to keyboard navigation, the outline isn't present.

https://github.com/user-attachments/assets/22981af1-baad-476c-b460-fbd5f8a72b77

jose-costa-frontify, line 53: I think this is not valid but once you remove that second `&` you might end up with the focus outline styling applying to the element any time it isn't hovered. I think this declaration might not be necessary since currently it's probably not doing anything

jose-costa-frontify (changes requested):

noahwaldner:
Let's push it to the next fondue release

packages/components/src/components/Dropdown/styles/dropdown.module.scss
jose-costa-frontify: This might do it

```suggestion
    &:hover:not(:focus-visible) {
        outline: none;
    }
```

noahwaldner: yes, i also noticed that, but i believe that its not a big issue.

The alternative would be to manually keep track of a separate selected state for the background and the outline. This would add quite a bit of hacky code which i believe is not worth it.

As there is no a11y guideline about focus indicators in select/dropdown elements this does not create an accessibility issue but is just a nice to have.

noahwaldner: fixed

SamuelAlev: The suggestion of José isn't applied, it would fix the issue with the mouse being on top of an element not adding the outline when switching to keyboard

noahwaldner: No it wouldn't because the elements are indeed "focus-visible" once the mouse hovers over them.
So this rule would not do anything.

jose-costa-frontify: yep, you're right. been reading a bit more on this and it seems like [it was quite contentious](https://github.com/radix-ui/primitives/issues/1164) but apparently inline with a11y patterns to have the focus be the same as the hover visually. Then they introduced the [data-highlighted](https://github.com/radix-ui/primitives/issues/1289) to allow devs to get around this limitation but it can't be done without the hover excluding the outline I don't think

jose-costa-frontify (approved):
LGTM

PR #1988, opened 2024-07-30
fix: missing scroll on select & dropdown

jose-costa-frontify:
CU-86954bwq8


packages/components/src/components/Select/__tests__/Select.ct.tsx
```diff
@@ -316,3 +318,30 @@ test('should render custom clear slot', async ({ mount, page }) => {
...
+        </Select>,
+    );
+
+    await expect(component).toBeVisible();
+    await component.click();
+
+    const dialog = page.getByTestId('fondue-select-menu');
+    await expect(dialog).toBeVisible();
+
+    const boundingBox = await dialog.boundingBox();
```
jose-costa-frontify, line 342: I was thinking this could be an util function so we can easily test other components that need the same feature. Testing this generically, like testing the hook or the util function, seems cumbersome and leaves open the possibility that some quirk of a particular implementation of this will generate an unintended result that is not captured by a test

packages/components/src/components/Dropdown/Dropdown.tsx
```diff
@@ -47,12 +50,22 @@ export const DropdownTrigger = (
 };
 DropdownTrigger.displayName = 'Dropdown.Trigger';
 
-export type DropdownContentProps = { children?: ReactNode; 'data-test-id'?: string };
+export type DropdownContentProps = {
```
noahwaldner, line 53: maybe add jsdoc now that it actually has props and not just children

packages/components/src/components/Dropdown/Dropdown.tsx
```diff
@@ -47,12 +50,22 @@ export const DropdownTrigger = (
...
+};
 
 export const DropdownContent = (
-    { children, 'data-test-id': dataTestId = 'fondue-dropdown-content' }: DropdownContentProps,
+    { onOpen, onClose, children, 'data-test-id': dataTestId = 'fondue-dropdown-content' }: DropdownContentProps,
     ref: ForwardedRef<HTMLDivElement>,
 ) => {
+    const localRef = useRef(null);
+    const dropdownIsOpen = useRef(false);
+
```
noahwaldner, line 66: unneccessary linebreak?

noahwaldner (approved):

packages/components/src/components/Dropdown/Dropdown.tsx
jose-costa-frontify: just separating react hooks from custom hooks, idk why I do it 🤣

PR #1990, opened 2024-07-31
fix: combobox filtering fix

noahwaldner:
- fix filtering in combobox

- exclude storybook caches from vite dependency optimisation to fix startup


jose-costa-frontify (approved):
🚀

PR #1997, opened 2024-08-07
refactor: tooltip value context approach [CU-8695b2cv6]

joseazevedo88:
A more granular approach was needed to fill the requirements of the new tooltip, dropped the `valueContextBySeries` and passed the value along with the data point, which is rendered by the tooltip.

Examples: 
image

Design: 
image

### DIFF
--- a/packages/charts/src/components/LineChart/LineChart.stories.tsx
+++ b/packages/charts/src/components/LineChart/LineChart.stories.tsx
@@ -1,9 +1,8 @@
 /* (c) Copyright Frontify Ltd., all rights reserved. */
 
+import { ChartHeading } from '@components/ChartHeading';
 import { type Meta, type StoryFn } from '@storybook/react';
 import { cityTemperature } from '@visx/mock-data';
-
-import { ChartHeading } from '@components/ChartHeading';
 
 import { IconRocket } from '../../utils/components/IconRocket';
 
@@ -147,10 +146,12 @@ const formattedTemperatureData = ((): LineChartSeries[] => {
         newYork.push({
             value: parseInt(obj['New York'], 10),
             timestamp: new Date(obj.date),
+            valueContext: `${Math.round(parseInt(obj['New York'], 10) / parseInt(obj.Austin, 10))}%`,
         });
         sanFrancisco.push({
             value: parseInt(obj['San Francisco'], 10),
             timestamp: new Date(obj.date),
+            valueContext: `(CTR ${Math.round(parseInt(obj['San Francisco'], 10) / parseInt(obj['New York'], 10))}%)`,
         });
     }
 
--- a/packages/charts/src/components/LineChart/LineChart.stories.tsx
+++ b/packages/charts/src/components/LineChart/LineChart.stories.tsx
@@ -1,9 +1,8 @@
 /* (c) Copyright Frontify Ltd., all rights reserved. */
 
+import { ChartHeading } from '@components/ChartHeading';
 import { type Meta, type StoryFn } from '@storybook/react';
 import { cityTemperature } from '@visx/mock-data';
-
-import { ChartHeading } from '@components/ChartHeading';
 
 import { IconRocket } from '../../utils/components/IconRocket';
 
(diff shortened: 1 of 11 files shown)


packages/charts/src/components/LineChart/LineChart.stories.tsx
```diff
@@ -147,10 +147,12 @@ const formattedTemperatureData = ((): LineChartSeries[] => {
         newYork.push({
             value: parseInt(obj['New York'], 10),
             timestamp: new Date(obj.date),
+            valueContext: `${Math.round(parseInt(obj['New York'], 10) / parseInt(obj.Austin, 10))}%`,
```
jose-costa-frontify, line 150: If the idea is to get a percentage, aren't we missing the `* 100`?

```suggestion
            valueContext: `${Math.round((parseInt(obj['New York'], 10) / parseInt(obj.Austin, 10)) * 100)}%`,
```

also would be nice if we were using an util for this to keep it dry

```suggestion
            valueContext: `${getPercentageFromNumStrings(obj['New York'], obj.Austin)}%`,
```

we also have to be careful because we might get a 0 as the denominator. I know this is just a story but it will set the example for other devs

packages/charts/src/components/common/components/Tooltip/helpers/getTooltipEntries.ts
```diff
@@ -35,8 +34,7 @@ export const getTooltipEntries = (
                 title: key,
                 value: getDataPointValue(missingValueLabel, datumByKey[key]?.datum.value, valueFormatter),
                 color: colorAccessor(key),
-
-                ...(valueContextBySeries?.[index] ? { valueContextBySeries: valueContextBySeries?.[index] } : {}),
+                ...(datumByKey[key]?.datum.valueContext ? { valueContext: datumByKey[key]?.datum.valueContext } : {}),
```
jose-costa-frontify, line 37: is this not functionally equivalent to simply doing

```suggestion
                valueContext: datumByKey[key]?.datum.valueContext,
```
since if `datumByKey[key]?.datum.valueContext` is undefined it results in that key of the object being undefined if it is assigned directly, just like it would be if we assign nothing at all by spreading an empty object.

jose-costa-frontify (changes requested):
looks good but I think we can make it a bit more readable for future devs

packages/charts/src/components/common/components/Tooltip/helpers/getTooltipEntries.ts
joseazevedo88: No difference, I just find it cleaner to pass the property if it needs the value, because it's a common component shared across charts. But I'll change it, since you guys are now responsible for this code 😄

packages/charts/src/components/LineChart/LineChart.stories.tsx
joseazevedo88: Can be anything really, updated the stories to make it clear

joseazevedo88:
@jose-costa-frontify we seem to use a different set of linter rules than web app. I always have to run `lint:fix` to fix the github action because I have formatting on save. Could be an interesting idea to unify this with web app. What do you think?

packages/charts/src/components/common/components/Tooltip/helpers/getTooltipEntries.ts
jose-costa-frontify: Don't do it because we're responsible, do it because it's cleaner (**if you agree**). 

If you prefer not passing anything at all, which I can understand, you can always move the object to a separate variable excluding the `valueContext` upon its initialization and then conditionally adding a the `valueContext` key before pushing the object into the array. I think we can both agree that combining ternaries with the spread operator and optional chaining results in some pretty esoteric code and dummies like myself will have a harder time reading it than plain KISS-style programming 😅

jose-costa-frontify:
> @jose-costa-frontify we seem to use a different set of linter rules than web app. I always have to run `lint:fix` to fix the github action because I have formatting on save. Could be an interesting idea to unify this with web app. What do you think?

There may be good reason for the disparity, I'm not sure. I haven't faced this issue though, is there any rule in particular that is conflicting?

jose-costa-frontify (approved):

joseazevedo88:
> > @jose-costa-frontify we seem to use a different set of linter rules than web app. I always have to run `lint:fix` to fix the github action because I have formatting on save. Could be an interesting idea to unify this with web app. What do you think?
> 
> There may be good reason for the disparity, I'm not sure. I haven't faced this issue though, is there any rule in particular that is conflicting?

Yup, the import sorting rule

PR #1998, opened 2024-08-08
feat: add dropdown migration guide

jose-costa-frontify:
CU-8695bwruq


packages/components/MIGRATING.md
```diff
@@ -246,7 +249,6 @@ Changes:
 
 ```tsx
 const [isOpen, setIsOpen] = useState(false);
-
```
SamuelAlev, line 249: Line break can be added back

SamuelAlev (approved):

PR #2001, opened 2024-08-09
feat: add custom value support to combobox

jose-costa-frontify:
## Definition of Done


packages/components/src/components/Select/useSelectData.tsx
```diff
@@ -53,42 +108,60 @@ export const useSelectData = (children: ReactNode) => {
...
+    }, [children, filterText, shouldAddCustomItem]);
 
     const getItemByValue = useCallback(
         (value?: string) => (value ? itemValues.find((item) => item.value === value) : undefined),
         [itemValues],
     );
 
+    const items = useMemo(() => {
+        const filteredItems = itemValues.filter(
+            (item) => filterText === '' || item.label.toLowerCase().includes(filterText.toLowerCase()),
```
jose-costa-frontify, line 139: could be nice to improve this search such that it splits each words and looks for instances of that word rather than inclusions

packages/components/src/components/Select/useSelectData.tsx
```diff
@@ -27,12 +42,52 @@ export const getRecursiveOptionValues = (children: ReactNode): { value: string;
...
+ * @property {boolean} shouldAddCustomItem - Whether a custom item should be added based on the current state.
+ * @property {boolean} valueInvalid - Whether the current value is invalid (when custom values are not allowed).
+ * @property {function} setFilterText - Function to update the filter text.
+ * @property {function} getItemByValue - Function to get an item by its value.
+ */
+export const useSelectData = (children: ReactNode, allowCustomValue = false) => {
     const [filterText, setFilterText] = useState('');
-    const { inputSlots, menuSlots, itemValues, clearButton } = useMemo(() => {
+
+    const itemValues = useMemo(() => getRecursiveOptionValues(children), [children]);
```
noahwaldner, line 68: this will most likely cause issues, as `children` also include the Input Slots. If slots are present, we only pass menuSlots

packages/components/src/components/Select/useSelectData.tsx
```diff
@@ -53,42 +108,60 @@ export const useSelectData = (children: ReactNode) => {
...
+
+            menuSlots.push(customItem);
+        }
+
+        return {
+            inputSlots,
+            menuSlots,
+            clearButton,
+        };
+    }, [children, filterText, shouldAddCustomItem]);
```
noahwaldner, line 130: By adding this, on every key stroke we will be parsing the slots again, even though they shouldn't ever change.

packages/components/src/components/Select/useSelectData.tsx
noahwaldner: could be done in the future, for now it's out of scope

packages/components/src/components/Select/useSelectData.tsx
```diff
@@ -53,42 +108,60 @@ export const useSelectData = (children: ReactNode) => {
...
+        );
+
+        if (shouldAddCustomItem) {
+            filteredItems.push({ label: filterText, value: filterText });
+        }
+
+        return filteredItems;
+    }, [itemValues, shouldAddCustomItem, filterText]);
+
+    const valueInvalid = useMemo(
```
noahwaldner, line 149: whats the benefit of moving this here? this only applies to the combobox

packages/components/src/components/Select/Combobox.tsx
```diff
@@ -135,14 +168,15 @@ export const SelectCombobox = (
                             {clearButton}
                         </RadixSlot>
                     )}
+
                     <button
+                        {...getToggleButtonProps()}
```
noahwaldner, line 173: reordering this might cause issues because overriding some props will be overridden that should not be

packages/components/src/components/Select/Combobox.tsx
```diff
@@ -52,77 +57,105 @@ export const SelectCombobox = (
...
+                        })}
+                        placeholder={placeholder}
+                        className={styles.input}
+                        disabled={disabled}
                         onMouseDown={(mouseEvent) => {
                             wasClicked.current = true;
                             mouseEvent.currentTarget.dataset.showFocusRing = 'false';
                         }}
-                        placeholder={placeholder}
-                        {...getInputProps({
```
noahwaldner, line 106: reordering this might cause issues with overrides that are not intended.

packages/components/src/components/Select/Combobox.tsx
```diff
@@ -52,77 +57,105 @@ export const SelectCombobox = (
...
-        () => inputValue && !items.find((item) => item.label.toLowerCase().includes(inputValue.toLowerCase())),
-        [inputValue, items],
-    );
+
+    const onBlurHandler = (blurEvent: FocusEvent<HTMLInputElement, Element>) => {
+        blurEvent.target.dataset.showFocusRing = 'false';
+        wasClicked.current = false;
+
+        if (valueExists) {
+            if (existingValueItem?.label) {
```
noahwaldner, line 112: out of scope, this doesn't happen as of now, unrelated from the custom value

if we change this, please create a separete pr

packages/components/src/components/Select/Select.tsx
```diff
@@ -34,6 +34,11 @@ export type SelectComponentProps = {
      * The placeholder in the select component.
      */
     placeholder?: string;
+    /**
+     * Allow custom value in the combobox component.
+     * @default false
+     */
+    allowCustomValue?: boolean;
```
noahwaldner, line 41: Out of scope! Also Select doesn't have a text field, no custom values are possible

noahwaldner (changes requested):

packages/components/src/components/Select/Combobox.tsx
jose-costa-frontify: will revert

packages/components/src/components/Select/Combobox.tsx
jose-costa-frontify: will revert

packages/components/src/components/Select/useSelectData.tsx
jose-costa-frontify: true, shouldn't be there

packages/components/src/components/Select/Combobox.tsx
jose-costa-frontify: I was about to copy the code from this over to a new branch when it occurred to me this will cause conflicts down the line if I implement it separately. For now, it's just a suggestion, we might get rid of it entirely but if we do decide to keep it, doesn't it make sense that it would be included alongside a new feature that makes it necessary?

packages/components/src/components/Select/Select.tsx
jose-costa-frontify: for some reason, if I remove this I get a type error on the story. I added it as a temporary fix before I spent some time figuring out what's actually happening

packages/components/src/components/Select/useSelectData.tsx
jose-costa-frontify: for sure, just a thought

packages/components/src/components/Select/useSelectData.tsx
```diff
@@ -53,42 +77,39 @@ export const useSelectData = (children: ReactNode) => {
...
     }, [children]);
 
-    const filteredItems = useMemo(
-        () =>
-            itemValues.filter(
-                (item) => filterText === '' || item.label.toLowerCase().includes(filterText.toLowerCase()),
-            ),
-        [itemValues, filterText],
-    );
+    const itemValues = useMemo(() => getRecursiveOptionValues(menuSlots), [menuSlots]);
```
noahwaldner, line 91: good change! makes it more readable 👍🏻

packages/components/src/components/Select/Combobox.tsx
noahwaldner: This "select on blur" behaviour is unrelated to the custom value change.

This only changes the behaviour when choosing a option thats pre existing. 
There is no change on this behaviour when `allowCustomOptions` is passed or not.

We're basically bundling two interaction changes together. 

Let's look at it together later

packages/components/src/components/Select/Combobox.tsx
noahwaldner: still has changed order

noahwaldner (changes requested):

packages/components/src/components/Select/useSelectData.tsx
jose-costa-frontify: reverted it though, I did this because when I moved the new logic into this hook I needed this to exist prior to the initialization of the slots but now there is no reason to keep it there as I improved the structure to make this not be necessary

PR #2003, opened 2024-08-11
feat: add select migration guide

jose-costa-frontify:


packages/components/MIGRATING.md
```diff
@@ -149,7 +152,6 @@ Changes:
 
 ```tsx
 <ColorPicker
-    {...args}
```
jose-costa-frontify, line 152: removed these as I think they add noise to the guide

packages/components/MIGRATING.md
```diff
@@ -608,6 +610,94 @@ Changes:
...
+            <Select.Item value="small-item">
+                <span class="tw-text-sm">Small item</span>
+            </Select.Item>
+        </Select.Slot>
+    </Select>
+)
+
+// Or, for a simpler setup you don't have to use the slots:
+
+<Select>
```
SamuelAlev, line 693: A bit strange decision we have made as `Dialog`, `Dropdown`, `Tooltip`, `TextInput` (depending on the usage) and `SegmentedControl` have `.Root` but not the `Select` 🤔

jose-costa-frontify: personally I prefer the way the `Select` does it. But we should normalize it one way or the other

PR #2011, opened 2024-08-15
refactor: styles of segmented controls to SCSS

SamuelAlev:
Task CU-8695dzecz


packages/components/src/components/SegmentedControl/styles/segmentedControl.module.scss
```diff
@@ -0,0 +1,193 @@
...
+    pointer-events: none;
+
+    // Outline of active item
+    &::before {
+        content: '';
+        position: absolute;
+        inset: 0;
+        border-radius: calc(var(--radius) - var(--line-width));
+        background-color: #fff;
+
```
jose-costa-frontify, line 140:
```suggestion
```

packages/components/src/components/SegmentedControl/styles/segmentedControl.module.scss
```diff
@@ -0,0 +1,193 @@
...
+.activeIndicator {
+    @include transition-transform;
+
+    display: none;
+    position: absolute;
+    z-index: -1;
+    top: 0;
+    left: 0;
+    height: 100%;
+
```
jose-costa-frontify, line 130:
```suggestion
```

packages/components/src/components/SegmentedControl/SegmentedControl.tsx
```diff
@@ -52,7 +44,7 @@ export const SegmentedControlRoot = (
         <ToggleGroupPrimitive.Root
             ref={ref}
             {...rootProps}
-            className={segmentedControlRootStyles}
+            className={styles.root}
```
jose-costa-frontify, line 47: sometimes we use `.root,` others we use `.elementName`, might be good to standardize. I don't have a preference but since `.root` is used most often, I suggest we normalize to that

jose-costa-frontify (changes requested):
Just a few house keeping things

packages/components/src/components/SegmentedControl/SegmentedControl.tsx
SamuelAlev: https://app.clickup.com/t/8695ec0ae

packages/components/src/components/SegmentedControl/styles/segmentedControl.module.scss
jose-costa-frontify: that's reasonable but since it's just one that is left out of the party maybe we can find a different grouping so that the space does a bit more heavy lifting?

packages/components/src/components/SegmentedControl/SegmentedControl.tsx
jose-costa-frontify: https://github.com/Frontify/fondue/pull/2012

jose-costa-frontify (approved):
LGTM

packages/components/src/components/SegmentedControl/styles/segmentedControl.module.scss
SamuelAlev: We can have a look at StyleLint now that we are with scss to make it automatic, wdyt?

jose-costa-frontify: sounds good!

PR #2013, opened 2024-08-16
fix(Flyout): margin top is overriden

fulopdaniel:
I was trying to fix this issue in https://github.com/Frontify/web-app/pull/5078 which got reverted, so this is the other approach.


SamuelAlev:
Will be fixed in https://github.com/Frontify/fondue/pull/2015, can I close the PR?

PR #2016, opened 2024-08-16
feat: handle focusing of interactive elements inside of a Flyout

noahwaldner:
handle focusing of interactive elements inside of a Flyout

- Don't show focus visible when opened with mouse

### DIFF
--- a/packages/components/src/components/Flyout/Flyout.tsx
+++ b/packages/components/src/components/Flyout/Flyout.tsx
@@ -80,7 +80,18 @@ export const FlyoutTrigger = (
     ref: ForwardedRef<HTMLButtonElement>,
 ) => {
     return (
-        <RadixPopover.Trigger data-test-id={dataTestId} asChild ref={ref}>
+        <RadixPopover.Trigger
+            onMouseDown={(mouseEvent) => {
+                console.log('mouse');
+
+                mouseEvent.currentTarget.dataset.autoFocusVisible = 'false';
+            }}
+            data-auto-focus-visible="true"
+            data-flyout-trigger
+            data-test-id={dataTestId}
+            asChild
+            ref={ref}
+        >
             {children}
         </RadixPopover.Trigger>
     );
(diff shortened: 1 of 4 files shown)


packages/components/src/components/Flyout/__tests__/Flyout.ct.tsx
```diff
@@ -364,3 +368,81 @@ test('should trigger callback on open and close', async ({ mount, page }) => {
...
+        'box-shadow',
+        'rgb(255, 255, 255) 0px 0px 0px 2px, rgb(65, 150, 251) 0px 0px 0px 6px',
+    );
+    await page.keyboard.press('Tab');
+    await expect(textInput2).toHaveCSS(
+        'box-shadow',
+        'rgb(255, 255, 255) 0px 0px 0px 2px, rgb(65, 150, 251) 0px 0px 0px 6px',
+    );
+    await textInput3.click();
+    await expect(textInput3).not.toHaveCSS(
```
jose-costa-frontify, line 405: use these thingies https://github.com/Frontify/fondue/blob/main/packages/components/src/helpers/constants.ts

packages/components/src/components/Flyout/Flyout.tsx
```diff
@@ -113,6 +122,16 @@ export const FlyoutContent = (
                 className={flyoutContentStyles({ ...props })}
                 data-flyout-spacing={padding}
                 data-test-id={dataTestId}
+                onFocus={(e) => {
```
jose-costa-frontify, line 125: I think we don't use initials for variable names

jose-costa-frontify (changes requested):
really good stuff 🚀

packages/components/src/components/Flyout/Flyout.tsx
noahwaldner: fixed

packages/components/src/components/Flyout/__tests__/Flyout.ct.tsx
noahwaldner: adjusted

jose-costa-frontify (approved):
LGTM

PR #2029, opened 2024-09-02
fix: prevent box-sizing override from normalization scripts in web app

jose-costa-frontify:
CU-8695kx55k


SamuelAlev (approved):

SamuelAlev:
Probably affect more components without tailwind, let's have an eye on it

PR #2033, opened 2024-09-04
chore: update deps

SamuelAlev:


packages/charts/src/components/LineChart/LineChart.spec.tsx
```diff
@@ -37,7 +38,9 @@ describe('LineChart', () => {
 
         const svgContainer = container.getElementsByTagName('svg')[0];
 
-        expect(svgContainer?.getAttribute('width')).toBe('400');
+        console.log(container.attributes);
```
jose-costa-frontify, line 41: still debugging?

SamuelAlev: you are just too fast ;p

jose-costa-frontify: image

jose-costa-frontify (approved):

PR #2035, opened 2024-09-04
fix: prevent dialog input autofocus

jose-costa-frontify:
CU-8695mu8rt


packages/components/src/components/Dialog/Dialog.stories.tsx
```diff
@@ -313,6 +314,29 @@ export const WithCustomElementAnnouncements: Story = {
     },
 };
 
+export const WithFocusableContent: Story = {
+    args: {
+        children: 'Hello World',
+    },
+    render: ({ ...args }) => {
+        return (
+            <Flyout.Root>
```
SamuelAlev, line 323: Dialog component not Flyout

packages/components/src/utilities/domUtilities.ts
```diff
@@ -75,3 +75,38 @@ export function syncRefs(localRef: RefObject<HTMLDivElement>, forwardedRef: Forw
...
 }
+
+/**
+ * Sets the 'autoFocusVisible' data attribute to 'false' on the current target of a mouse event.
+ * This function is typically used to indicate that an element has been interacted with via mouse,
+ * which should prevent the focus ring from being displayed when the element receives focus.
+ *
+ * @param mouseEvent - The mouse event object.
+ */
+export function addAutoFocusAttribute(mouseEvent: ReactMouseEvent<HTMLButtonElement, MouseEvent>) {
```
SamuelAlev, line 86: Can be added to Flyout component too

jose-costa-frontify: yep I did!

packages/components/src/components/Dialog/Dialog.stories.tsx
jose-costa-frontify: no more pasta for me today

PR #2040, opened 2024-09-06
fix(CollapsibleWrap): overflow is kept hidden after expansion

fulopdaniel:
To avoid scenarios like this I set the overflow to visible after the animation is done.

image


jose-costa-frontify (approved):
strange... shouldn't the container take the height of its content in that situation?

PR #2049, opened 2024-09-10
fix: align dropdown and select styles with figma

noahwaldner:
Dropdown:
- Align styling with figma

Select:
- Align Styling with Figma
- Add highlight for selected option
- add test for selected option

Task: CU-8695prr3h


packages/components/src/components/Select/SelectMenu.tsx
```diff
@@ -52,7 +73,13 @@ export const SelectMenu = ({ highlightedIndex, getMenuProps, getItemProps, child
     return (
         <RadixPopover.Portal>
             <RadixPopover.Content onOpenAutoFocus={handleOnOpenAutoFocus} className={styles.portal}>
-                <ul className={styles.menu} {...getMenuProps()} ref={ref} data-test-id="fondue-select-menu">
+                <ul
+                    className={styles.menu}
+                    {...getMenuProps()}
+                    ref={ref}
+                    data-has-interacted={hasInteractedSinceOpening ? 'true' : 'false'}
```
jose-costa-frontify, line 80: since `hasInteractedSinceOpening` is a boolean, why not
```suggestion
                    data-has-interacted={hasInteractedSinceOpening}
```

packages/components/src/components/Select/SelectMenu.tsx
```diff
@@ -37,9 +37,30 @@ export type SelectMenuProps = {
      * The filter text shown in the combobox input element.
      */
     filterText?: string;
+    /**
+     * @internal
+     * The type of the menu.
+     */
+    selectedItem?: {
+        value: string;
+    } | null;
```
jose-costa-frontify, line 46: why both `| null` and optional?

packages/components/src/components/Select/SelectMenu.tsx
```diff
@@ -64,15 +91,18 @@ export const SelectMenu = ({ highlightedIndex, getMenuProps, getItemProps, child
...
                                         index,
                                         ...(child.ref ? { ref: child.ref } : {}),
                                     });
+
                                     return (
                                         <RadixSlot
                                             className={styles.item}
                                             data-highlighted={highlightedIndex === index}
+                                            data-selected={selectedItem?.value === optionData.value}
                                             key={`${child.props.value}`}
```
jose-costa-frontify, line 106:
```suggestion
                                            key={child.props.value}
```

packages/components/src/components/Select/__tests__/Select.ct.tsx
```diff
@@ -429,8 +429,45 @@ test('render border and no focus ring when mouse focused', async ({ mount, page
...
+
+    await expect(firstItem).toHaveAttribute('data-selected', 'true');
+    await expect(firstItem).toHaveCSS('background-color', 'rgb(66, 71, 71)');
+    await expect(chekmarkIcon).toBeVisible();
+
+    await page.keyboard.press('ArrowDown');
+
+    await expect(firstItem).toHaveCSS('background-color', 'rgb(240, 234, 250)');
+    await expect(chekmarkIcon).toBeVisible();
+    await expect(chekmarkIcon).toBeVisible();
```
jose-costa-frontify, line 472:
```suggestion
```

jose-costa-frontify (changes requested):
all good, just some cosmetic thingies

packages/components/src/components/Select/SelectMenu.tsx
noahwaldner: data attributes need to be a string afaik? am i wrong?

packages/components/src/components/Select/SelectMenu.tsx
noahwaldner: For whatever reason `useCombobox` uses null and `useSelect` uses undefined (downshift hooks)

packages/components/src/components/Select/SelectMenu.tsx
jose-costa-frontify: don't think so, if memory serves it ends up as a string on the DOM anyway

jose-costa-frontify (approved):
LGTM

PR #2053, opened 2024-09-11
fix: add full width wrapper in the header to allow more content

SamuelAlev:
Allow header to expand full width

### DIFF
--- a/packages/components/src/components/Dialog/styles/dialog.module.scss
+++ b/packages/components/src/components/Dialog/styles/dialog.module.scss
@@ -133,7 +133,6 @@
     }
 
     > div {
-        display: flex;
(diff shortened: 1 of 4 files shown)


packages/components/src/components/Dialog/styles/dialog.module.scss
```diff
@@ -131,6 +131,11 @@
         grid-row: 1;
         grid-column: 2;
     }
+
+    > div {
+        display: flex;
```
jose-costa-frontify: `display: flex` seems to be unnecessary, the 100% width should suffice

jose-costa-frontify (changes requested):

PR #2055, opened 2024-09-12
fix(charts): pie chart label overlapping [CU-86930eter]

joseazevedo88:
### In some edge cases the pie chart label and percentage were overlapping:
image

### Should be:
image

#### Made the percentage be a span in line with the value, so it doesn't need to be placed with the coordinates, which were somehow(couldn't figure it out why) miscalculated sometimes.

### DIFF
--- a/.changeset/rotten-poems-guess.md
+++ b/.changeset/rotten-poems-guess.md
@@ -2,4 +2,4 @@
 "@frontify/fondue-charts": patch
 ---
 
-Pie chart label overlap fix
+fix: prevent overlap of label and percent on PieChart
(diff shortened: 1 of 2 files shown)


SamuelAlev (approved):

.changeset/rotten-poems-guess.md
```diff
@@ -0,0 +1,5 @@
+---
+"@frontify/fondue-charts": patch
+---
+
+Pie chart label overlap fix
```
SamuelAlev, line 5: fix: prevent overlap of label and percent on PieChart

PR #2056, opened 2024-09-12
fix: tooltip data-state conflicts with other components

jose-costa-frontify:
CU-8695t5fjm

more on this issue:

https://github.com/radix-ui/primitives/discussions/560
https://github.com/radix-ui/primitives/issues/2096

### DIFF
--- a/packages/components/src/components/Tooltip/Tooltip.tsx
+++ b/packages/components/src/components/Tooltip/Tooltip.tsx
@@ -24,13 +24,6 @@ export type TooltipRootProps = {
     children: Array<ReactElement<TooltipTriggerProps | TooltipContentProps>>;
 };
 export type TooltipTriggerProps = {
-    /**
-     * Whether to wrap the child in a trigger element.
-     * When false, add a wrapper with the trigger behavior around the child.
-     * When true, applies trigger behavior directly to the child.
-     * @default false
-     */
-    asChild?: boolean;
     children: ReactNode;
     'data-test-id'?: string;
 };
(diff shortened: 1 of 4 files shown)


packages/components/src/components/Tooltip/Tooltip.tsx
```diff
@@ -55,7 +55,7 @@ export const TooltipTrigger = (
     ref: ForwardedRef<HTMLButtonElement>,
 ) => {
     return (
-        <RadixTooltip.Trigger data-test-id={dataTestId} asChild ref={ref}>
+        <RadixTooltip.Trigger data-test-id={dataTestId} ref={ref}>
```
jose-costa-frontify, line 58: I haven't seen any side effects on this change but we can expose the `asChild` prop if we want to be on the safe side

packages/components/src/components/Tooltip/Tooltip.tsx
```diff
@@ -23,7 +23,17 @@ export type TooltipRootProps = {
     enterDelay?: number;
     children: Array<ReactElement<TooltipTriggerProps | TooltipContentProps>>;
 };
-export type TooltipTriggerProps = { children: ReactNode; 'data-test-id'?: string };
+export type TooltipTriggerProps = {
+    /**
+     * Whether the trigger should be applied to the child of a wrapper element.
+     * If disabled, the margins around the element will not affect the tooltip positioning.
+     * Enable this if you have styling or behavioral issues with the child element.
```
noahwaldner, line 30:
```suggestion
     * Whether to wrap the child in a trigger element.
     * When true, add a wrapper with the trigger behavior around the child.
     * When false, applies trigger behavior directly to the child.
     * Enable when direct application causes styling or behavioral issues.
```

packages/components/src/components/Tooltip/Tooltip.tsx
```diff
@@ -51,11 +62,11 @@ export const TooltipRoot = ({ children, enterDelay = 700, open, onOpenChange }:
 TooltipRoot.displayName = 'Tooltip.Root';
 
 export const TooltipTrigger = (
-    { children, 'data-test-id': dataTestId = 'fondue-tooltip-trigger' }: TooltipTriggerProps,
+    { wrapChild = false, children, 'data-test-id': dataTestId = 'fondue-tooltip-trigger' }: TooltipTriggerProps,
     ref: ForwardedRef<HTMLButtonElement>,
 ) => {
     return (
-        <RadixTooltip.Trigger data-test-id={dataTestId} asChild ref={ref}>
+        <RadixTooltip.Trigger asChild={!wrapChild} data-test-id={dataTestId} ref={ref}>
```
SamuelAlev, line 69: Why did we have to rename?

SamuelAlev: Render delegation pattern usually use the `as` cc @noahwaldner

jose-costa-frontify: Personally I'd go with `asChild`, I think it's as descriptive as `wrapChild` plus we didn't come up with it

SamuelAlev (approved):

PR #2060, opened 2024-09-17
feat(RTE): portal the link modal

fulopdaniel:
Use portal for the modals.

### DIFF
--- a/packages/fondue/src/components/RichTextEditor/Plugins/LinkPlugin/FloatingLink/CustomFloatingLink.tsx
+++ b/packages/fondue/src/components/RichTextEditor/Plugins/LinkPlugin/FloatingLink/CustomFloatingLink.tsx
@@ -57,7 +57,7 @@ export const CustomFloatingLink = () => {
             {insertState.isOpen &&
                 !editState.isOpen &&
                 createPortal(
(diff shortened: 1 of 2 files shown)


packages/fondue/src/components/RichTextEditor/Plugins/LinkPlugin/FloatingLink/CustomFloatingLink.tsx
```diff
@@ -55,21 +56,30 @@ export const CustomFloatingLink = () => {
...
-                    {input}
-                </div>
-            )}
+            {insertState.isOpen &&
+                !editState.isOpen &&
+                createPortal(
+                    <div
+                        ref={insertRef}
+                        {...insertProps}
+                        style={{ ...insertProps.style, zIndex: zIndexLayers.floatingModal }}
```
SamuelAlev, line 65: if you are using Portal, let's leverage the DOM order layering instead of adding `zIndex`

fulopdaniel: Fair point, adjusted 👍

SamuelAlev (approved):

PR #2062, opened 2024-09-17
feat: add asChild prop to tooltip

jose-costa-frontify:


packages/components/src/components/Tooltip/Tooltip.tsx
```diff
@@ -24,6 +24,11 @@ export type TooltipRootProps = {
     children: Array<ReactElement<TooltipTriggerProps | TooltipContentProps>>;
 };
 export type TooltipTriggerProps = {
+    /**
+     * If true, the trigger will not be wrapped in a button element.
```
SamuelAlev, line 28: Radix UI description of `asChild`: Change the default rendered element for the one passed as a child, merging their props and behavior.

PR #2067, opened 2024-09-19
feat: add workflow for generating icons

noahwaldner:


.github/workflows/icon-generation.yaml
```diff
@@ -0,0 +1,67 @@
...
+
+      - name: Build icons
+        run: pnpm build:icons
+        env:
+          FIGMA_ACCESS_TOKEN: ${{ secrets.FIGMA_ACCESS_TOKEN }}
+
+      - name: Generate icons
+        run: pnpm --stream --filter {packages/fondue} generate:icons
+
+      - name: Create Unique Changeset
```
jose-costa-frontify, line 33: since this is created, won't it mean that `git-check.outputs.changes` is always true?

jose-costa-frontify (changes requested):

.github/workflows/icon-generation.yaml
```diff
@@ -0,0 +1,67 @@
+name: Update Icons
+
+on:
+  workflow_dispatch:
```
SamuelAlev, line 4: let’s add a cron job every week just in case there are changes without us being aware?

.github/workflows/icon-generation.yaml
SamuelAlev: Yup José is right, we need create the changeset file if « git status » returns you any changes (the build script will add/modify/delete files if there are changes)

.github/workflows/icon-generation.yaml
noahwaldner: sounds good, will add it

.github/workflows/icon-generation.yaml
noahwaldner: sounds good, will reorder

jose-costa-frontify (approved):

PR #2077, opened 2024-10-04
feat: refactor tabs component

noahwaldner:


packages/components/src/components/Dropdown/Dropdown.tsx
```diff
@@ -101,6 +107,8 @@ export const DropdownContent = (
                     dropdownIsOpen.current = false;
                 }}
                 onFocus={() => {
+                    console.log('onFocus');
+
```
SamuelAlev, line 111:
```suggestion
```

packages/components/src/components/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,194 @@
...
+     * The default active tab
+     * Used for uncontrolled components
+     */
+    defaultActiveTab?: string;
+    /**
+     * The controlled value of the active tab
+     */
+    activeTab?: string;
+    /**
+     * The height of the tabs
```
SamuelAlev, line 29: Missing `@default`

packages/components/src/components/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,194 @@
...
+     */
+    size?: 'default' | 'large';
```
SamuelAlev, line 31: No other component has `default` as size let's align it

packages/components/src/components/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,194 @@
...
+
+    const localRef = useRef<HTMLButtonElement>(null);
+
+    useEffect(() => {
+        addTrigger({
+            ref: localRef || ref,
+            value: value || '',
+            disabled,
+            element: (
+                <button ref={localRef || ref} {...props} data-bla="ttt">
```
SamuelAlev, line 163:
```suggestion
                <button ref={localRef || ref} {...props}>
```

packages/components/src/components/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,194 @@
...
+
+type TabsContentProps = {
+    children: ReactNode;
+};
+
+export const TabsContent = ({ children, ...itemProps }: TabsContentProps, ref: ForwardedRef<HTMLDivElement>) => {
+    const { value } = useContext(TabConfigContext);
+
+    return (
+        <RadixTabs.Content ref={ref} {...itemProps} className={styles.content} value={value || ''}>
```
SamuelAlev, line 182: Prefer `??` to `||` when possible
```suggestion
        <RadixTabs.Content ref={ref} {...itemProps} className={styles.content} value={value ?? ''}>
```

packages/components/src/components/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,194 @@
...
+TabsRoot.displayName = 'Tabs.Root';
+
+type TabsTabProps = {
+    children: ReactNode;
+    value: string;
+    disabled?: boolean;
+};
+
+export const TabsTab = ({ children, value, disabled }: TabsTabProps) => (
+    <TabConfigContext.Provider value={{ value, disabled: !!disabled }}>{children}</TabConfigContext.Provider>
```
SamuelAlev, line 142: `disabled` is already a boolean, if not possible prefer the fallback approach `disabled ?? false/true`
```suggestion
    <TabConfigContext.Provider value={{ value, disabled }}>{children}</TabConfigContext.Provider>
```

packages/components/src/components/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,194 @@
...
+                                </Button>
+                            </Dropdown.Trigger>
+                        )}
+                        <Dropdown.Content align="end" data-test-id="overflow-items-dropdown-content">
+                            {triggersOutOfView.map((trigger) => (
+                                <Dropdown.Item
+                                    disabled={trigger.disabled}
+                                    onSelect={() => {
+                                        setActiveTab(trigger.value);
+                                    }}
```
SamuelAlev, line 114: As it is in `.map`, move it out and add a `useCallback`

packages/components/src/components/Tabs/constants.ts
SamuelAlev, line 1: No usage found, file can be deleted
If I am misstaken, you can directly inline it

packages/components/src/components/Tabs/hooks/useTabTriggers.tsx
```diff
@@ -0,0 +1,139 @@
...
+        const activeTriggerElement = triggerListRef.current?.querySelector('[data-state="active"]');
+        if (activeTriggerElement) {
+            moveActiveIndicator(triggerListRef, activeIndicatorRef);
+            activeTriggerElement.scrollIntoView({ behavior: 'smooth', block: 'nearest', inline: 'end' });
+        }
+    }, [activeTab, activeIndicatorRef]);
+
+    useLayoutEffect(() => {
+        const activeTriggerElement = triggerListRef.current?.querySelector('[data-state="active"]') as HTMLElement;
+        const triggerListElement = triggerListRef.current as HTMLDivElement;
```
SamuelAlev, line 91: Use proper runtime assertion
`if (activeTriggerElement && triggerListElement)` (and even some `instanceof`)

Because it could fail to find with `querySelector` in case:
- there is no `data-state`
- `triggerListRef.current` is `null`

```suggestion
        const activeTriggerElement = triggerListRef.current?.querySelector('[data-state="active"]');
        const triggerListElement = triggerListRef.current;
```

packages/components/src/components/Tabs/hooks/useTabTriggers.tsx
```diff
@@ -0,0 +1,139 @@
...
+    activeTab?: string;
+}): {
+    triggerListRef: RefObject<HTMLDivElement>;
+    activeIndicatorRef: RefObject<HTMLSpanElement>;
+    triggers: TabTrigger[];
+    triggersOutOfView: TabTrigger[];
+    addTrigger: (trigger: TabTrigger) => void;
+} => {
+    const triggerListRef = useRef<HTMLDivElement>(null);
+    const activeIndicatorRef = useRef<HTMLSpanElement>(null);
```
SamuelAlev, line 76:
```suggestion
    const triggerListRef = useRef<HTMLDivElement | null>(null);
    const activeIndicatorRef = useRef<HTMLSpanElement | null>(null);
```

packages/components/src/components/Tabs/hooks/useTabTriggers.tsx
```diff
@@ -0,0 +1,139 @@
...
+                moveActiveIndicator(triggerListRef, activeIndicatorRef);
+            },
+            {
+                root: triggerListElement,
+                threshold: 1,
+            },
+        );
+        for (const trigger of triggers) {
+            if (trigger.ref?.current) {
+                intersectionObserver.observe(trigger.ref.current);
```
SamuelAlev, line 111: One interaction observer per trigger seems not so great, can't you have it on the parent instead?

packages/components/src/components/Tabs/types.ts
```diff
@@ -0,0 +1,10 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+import { type RefObject } from 'react';
+
+export type TabTrigger = {
+    ref?: RefObject<HTMLButtonElement>;
+    element: JSX.Element;
```
SamuelAlev, line 7:
```suggestion
    element: ReactElement;
```
?

packages/components/src/index.ts
```diff
@@ -22,3 +22,4 @@ export { Slider } from './components/Slider/Slider';
 export { Switch } from './components/Switch/Switch';
 export { TextInput } from './components/TextInput/TextInput';
 export { Tooltip } from './components/Tooltip/Tooltip';
+export { Tabs } from './components/Tabs/Tabs';
```
SamuelAlev, line 25:
```suggestion
export { Tabs } from './components/Tabs/Tabs';
export { TextInput } from './components/TextInput/TextInput';
export { Tooltip } from './components/Tooltip/Tooltip';
```

packages/components/src/components/Tabs/__tests__/Tabs.ct.tsx
```diff
@@ -0,0 +1,516 @@
...
+        await component.getByTestId(DROPDOWN_TRIGGER_TEST_ID).click();
+        const dropdownContent = page.getByTestId(DROPDOWN_CONTENT_TEST_ID);
+        await expect(dropdownContent).toBeVisible();
+        await expect(dropdownContent.getByText(SECOND_TAB_TRIGGER_TEXT)).toBeVisible();
+        await expect(dropdownContent.getByText(SECOND_TAB_TRIGGER_TEXT)).toBeDisabled();
+        await expect(dropdownContent.getByText(THIRD_TAB_TRIGGER_TEXT)).toBeVisible();
+
+        await dropdownContent.getByText(THIRD_TAB_TRIGGER_TEXT).click();
+
+        await page.waitForTimeout(500);
```
SamuelAlev, line 504: Is it necessary now that you have the reduce motion?

if so still prefer a less "fixed" approach and use `waitForSelector`

packages/components/src/components/Tabs/Tabs.tsx
noahwaldner: added

packages/components/src/components/Tabs/Tabs.tsx
noahwaldner: fixed

packages/components/src/components/Tabs/Tabs.tsx
noahwaldner: cleaned

packages/components/src/components/Tabs/Tabs.tsx
noahwaldner: fixed

packages/components/src/components/Tabs/Tabs.tsx
noahwaldner: fixed

packages/components/src/components/Tabs/Tabs.tsx
noahwaldner: good catch!

packages/components/src/components/Tabs/hooks/useTabTriggers.tsx
noahwaldner: added runtime assertion

packages/components/src/components/Tabs/hooks/useTabTriggers.tsx
noahwaldner: added

packages/components/src/components/Tabs/hooks/useTabTriggers.tsx
noahwaldner: there is actually only one `intersectionObserver` 

the `observe()` function adds targets to it and is not creating new observers

packages/components/src/components/Tabs/types.ts
noahwaldner: good catch! changed

packages/components/src/index.ts
noahwaldner: reordered

packages/components/src/components/Tabs/__tests__/Tabs.ct.tsx
noahwaldner: playwright is super flaky with these transitions and animations, there are several issues open about it.

With the waitFor i can get the test to pass but it isn't stable (~1 out of 5 tries just fails and it reads the wrong position) 
I really wanna avoid having a test in here thats not 100% stable

packages/components/src/components/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,205 @@
...
+import { useControllableState } from '#/hooks/useControllableState';
+
+import { Button } from '../Button/Button';
+import { Dropdown } from '../Dropdown/Dropdown';
+
+import { useTabTriggers } from './hooks/useTabTriggers';
+import styles from './styles/tabs.module.scss';
+import { type TabTrigger } from './types';
+
+export type TabsRootProps = {
```
jose-costa-frontify, line 25: how do you feel about having the types next to the components they belong to rather than separating with all types first and then all components below? I think it makes for an easier read, wdyt?

noahwaldner: imo because the types are all very short, it is very easy to read this way.

I'm also okay with having them at the top if you feel that would bring us a improvement

packages/components/src/components/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,205 @@
...
+                                disabled={trigger.disabled}
+                                className={styles.trigger}
+                                asChild
+                            >
+                                {trigger.element}
+                            </RadixTabs.Trigger>
+                        ))}
+                    </RadixTabs.List>
+                    <Dropdown.Root>
+                        {triggersOutOfView.length > 0 && (
```
jose-costa-frontify, line 114: shouldn't this condition wrap the entire dropdown?

packages/components/src/components/Tabs/hooks/useTabTriggers.tsx
```diff
@@ -0,0 +1,140 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+import { cloneElement, type RefObject, useEffect, useLayoutEffect, useRef, useState } from 'react';
+
+import { type TabTrigger } from '../types';
+
+const getOverflowingTriggers = (triggers: TabTrigger[], triggerListElement: HTMLDivElement) => {
+    const triggersOutOfView: TabTrigger[] = [];
+    if (triggerListElement) {
```
jose-costa-frontify, line 9: to avoid the pyramid we could do 
```suggestion
    if (!triggerListElement) {
      return triggersOutOfView;
    }
```

packages/components/src/components/Tabs/hooks/useTabTriggers.tsx
```diff
@@ -0,0 +1,140 @@
...
+        for (const trigger of triggers) {
+            const triggerElement = trigger.ref?.current;
+            if (triggerElement) {
```
jose-costa-frontify, line 12: same here

packages/components/src/components/Tabs/hooks/useTabTriggers.tsx
```diff
@@ -0,0 +1,140 @@
...
+                if (
+                    triggerElement.offsetLeft + triggerElement.offsetWidth >
+                        triggerListElement?.scrollLeft + triggerListElement.offsetWidth ||
```
jose-costa-frontify, line 15: can `triggerListElement` be undefined?

jose-costa-frontify (changes requested):
I noticed some inconsistent behavior with the hidden tabs

As soon as the container width intersects with the tabs width, the last tab gets hidden, show no decorators:
Screenshot 2024-10-04 at 14 07 32

Then as we continue to decrease the container size, the second tab is not hidden:
Screenshot 2024-10-04 at 14 07 44

If we continue decreasing it, then when the first tab is intersected, both the first and second tab get hidden and they include the decorators: 

Screenshot 2024-10-04 at 14 08 02

Then if we expand the container past the point of the second tab, despite it becoming visible, it stays within the hidden tabs until we start revealing the third tab by increasing the container even further:

image

packages/components/src/components/Tabs/Tabs.tsx
jose-costa-frontify: yes in this case it's not a big deal but it's one of those things where consistency helps a bit. I think this is a good topic of discussion for our focus week, that and the use of implicit returns which sometimes I feel we do and other times not and I'm not 100% clear on what our philosophy is on that

packages/components/src/components/Tabs/styles/tabs.module.scss
```diff
@@ -0,0 +1,83 @@
...
+.triggerList {
+    position: relative;
+    flex-shrink: 0;
+    display: flex;
+    gap: sizeToken(1);
+    height: sizeToken(12);
+    flex: 1;
+    min-width: 0;
+    overflow-x: hidden;
+    &[data-size='large'] {
```
jose-costa-frontify, line 25: to avoid the override of height we could perhaps assign the default height also to its respective data attribute

packages/components/src/components/Tabs/hooks/useTabTriggers.tsx
noahwaldner: fixed

packages/components/src/components/Tabs/hooks/useTabTriggers.tsx
noahwaldner: fixed

packages/components/src/components/Tabs/hooks/useTabTriggers.tsx
noahwaldner: fixed

packages/components/src/components/Tabs/styles/tabs.module.scss
noahwaldner: what would be the benefit of that?

noahwaldner:
@jose-costa-frontify i improved the overflowing a bit, some of your cases were not reprodcible though

jose-costa-frontify (approved):
image

I'm not 100% on board with the way the indicator behaves when it is on an item which is overflowing, I would rather it stay below the item, also overflowing. It might be misleading at certain times where the indicator seems to be below some other item. 

The issues I mentioned previously are gone so from a technical stand point I have no issues, all looks good!

packages/components/src/components/Tabs/hooks/useTabTriggers.tsx
```diff
@@ -0,0 +1,140 @@
...
+                        element: cloneElement(trigger.element, { ref: null }),
+                    });
+                }
+            }
+        }
+    }
+    return triggersOutOfView;
+};
+
+const moveActiveIndicator = (
```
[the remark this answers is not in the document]
noahwaldner: not really as it is very different because of the scroll

packages/components/src/components/Tabs/hooks/useTabTriggers.tsx
```diff
@@ -18,17 +18,20 @@ const getOverflowingTriggers = (triggers: TabTrigger[], triggerListElement: HTML
...
+
+    if (
+        !triggerListElement ||
+        !(activeTriggerElement instanceof HTMLButtonElement) ||
+        !(activeIndicatorElement instanceof HTMLSpanElement)
+    ) {
         return;
     }
 
+    console.log(activeTriggerElement);
```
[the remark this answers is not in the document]
jose-costa-frontify: it appeared to me in the files, maybe a fluke

PR #2081, opened 2024-10-24
feat(Flyout): add rounded and shadow options

jose-costa-frontify:
CU-8696apu1y
CU-8696apu8n

### DIFF
--- a/packages/components/src/components/Flyout/Flyout.tsx
+++ b/packages/components/src/components/Flyout/Flyout.tsx
@@ -63,10 +63,15 @@ FlyoutTrigger.displayName = 'Flyout.Trigger';
 
 export type FlyoutContentProps = {
     /**
-     * Add rounded corners to the flyout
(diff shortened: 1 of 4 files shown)


packages/components/src/components/Flyout/Flyout.tsx
```diff
@@ -62,11 +62,16 @@ export const FlyoutTrigger = (
...
 
 export type FlyoutContentProps = {
+    /**
+     * Add a shadow to the flyout
+     * @default "medium"
+     */
+    shadow?: 'none' | 'medium' | 'large';
     /**
      * Add rounded corners to the flyout
-     * @default true
```
jose-costa-frontify, line 67: the default used to say true but it was actually undefined previously, I assume this was an oversight and I now set the default to medium (same as true before)

noahwaldner (approved):

PR #2082, opened 2024-10-24
feat(Flyout): add width prop

jose-costa-frontify:
CU-8696apw43


packages/components/src/components/Flyout/Flyout.tsx
```diff
@@ -93,9 +98,10 @@ export type FlyoutContentProps = {
 
 export const FlyoutContent = (
     {
-        align = 'start',
```
noahwaldner, line 96: why nor alphabetical

jose-costa-frontify: to not separate `width` from `maxWidth` but don't have a strong preference

noahwaldner (approved):

PR #2090, opened 2024-10-31
chore: deprecation housekeeping

noahwaldner:
- deprecated legacy components that were replaced 
- set new components to 'released'

Feature additions:
- added "danger" style to `DropdownItem` to be able to replace the old `ActionMenu`
- added "showShadow" to the `ScrollArea` to be able to replace the old `ScrollWrapper`

### DIFF
--- a/packages/components/src/components/Dropdown/Dropdown.tsx
+++ b/packages/components/src/components/Dropdown/Dropdown.tsx
@@ -215,7 +215,7 @@ export type DropdownItemProps = {
      * The style of the item.
      * @default "default"
      */
-    style?: 'default' | 'danger';
+    emphasis?: 'default' | 'danger';
     /**
      * Callback that is called when the item is selected.
      */
@@ -229,7 +229,7 @@ export const DropdownItem = (
         disabled,
         textValue,
         onSelect,
-        style = 'default',
+        emphasis = 'default',
         'data-test-id': dataTestId = 'fondue-dropdown-subtrigger',
         ...props
     }: DropdownItemProps,
@@ -241,7 +241,7 @@ export const DropdownItem = (
             className={styles.item}
             textValue={textValue}
             data-test-id={dataTestId}
-            data-style={style}
+            data-emphasis={emphasis}
             ref={ref}
             disabled={disabled}
             {...props}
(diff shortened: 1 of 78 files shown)


packages/components/src/components/Dropdown/Dropdown.tsx
```diff
@@ -211,6 +211,11 @@ export type DropdownItemProps = {
      * The text value of the item that is passed to the onSelect callback.
      */
     textValue?: string;
+    /**
+     * The style of the item.
+     * @default "default"
+     */
+    style?: 'default' | 'danger';
```
jose-costa-frontify, line 218: feels like this stuff could've been a separate PR

packages/components/src/components/ScrollArea/styles/scrollArea.module.scss
```diff
@@ -13,6 +13,10 @@
     width: 100%;
     height: 100%;
     padding-right: var(--scrollbar-size);
+
+    &[data-show-shadow="true"] {
+        box-shadow: inset 0px -11px 8px 0px var(--box-neutral-color);
```
jose-costa-frontify, line 18: this one looks a bit odd because it also appears on the right side even if there is no horizontal scroll. idk how we could target the horizontal scroll when it exists, but since usually horizontal scrolling is a bad thing, we could stick to only showing the shadow on the bottom. 

```suggestion
        box-shadow: inset 0 -15px 8px -10px var(--box-neutral-color);
```

Also, the shadow appears even when there is no scroll present... it'd be cool if radix exposed some data attrs to let us know when the content is overflowing. Say we use this in the brand chooser, people with 6 or less brands will end up seeing a weird shadow... maybe we can figure out if there is overflow ourselves and apply this programmatically?

jose-costa-frontify (approved):

packages/components/src/components/Dropdown/Dropdown.tsx
noahwaldner: agree

packages/components/src/components/ScrollArea/styles/scrollArea.module.scss
noahwaldner: let's iterate on this!

PR #2096, opened 2024-11-06
feat(charts): add support for image rendering inside tooltip

joseazevedo88:
#### Example story
image

#### Design
image

### DIFF
--- a/packages/charts/src/components/common/components/Tooltip/TooltipContent.tsx
+++ b/packages/charts/src/components/common/components/Tooltip/TooltipContent.tsx
@@ -22,7 +22,11 @@ export const TooltipContent = ({ title, description, imageUrl, entries }: Toolti
 
     return (
         <div className="tw-bg-[var(--text-color)] tw-p-3 tw-rounded tw-border tw-border-button-border">
-            {imageUrl && <img src={imageUrl} alt={description} className="tw-w-48 tw-h-28 tw-object-cover tw-mb-3" />}
+            {imageUrl && (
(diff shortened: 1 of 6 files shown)


packages/charts/src/components/common/components/Tooltip/TooltipContent.tsx
```diff
@@ -13,18 +13,24 @@ type TooltipContentV2Props = {
     entries: TooltipEntry[];
     title?: string;
     description?: string;
+    imageUrl?: string;
 };
 
-export const TooltipContent = ({ title, description, entries }: TooltipContentV2Props) => {
+export const TooltipContent = ({ title, description, imageUrl, entries }: TooltipContentV2Props) => {
```
SamuelAlev, line 19: Why not go with a composition approach?
```tsx
<TooltipContent.Root>
  <TooltipContent.Title>My title</TooltipContent.Title>
  <TooltipContent.Body>
    <TooltipContent.Image src="" alt="" />
    <Text>Some content under the image</Text>
  </TooltipContent.Body>
</TooltipContent.Root>
```

joewoodson (approved):
LGTM, tested and working well 👍🏽

packages/charts/src/components/common/components/Tooltip/TooltipContent.tsx
joseazevedo88: That's a good pattern(although I don't like the dot notation), but it's out of the scope of this PR

PR #2104, opened 2024-11-14
feat(Table): refactor component

jose-costa-frontify:
CU-8696na6d4

## Definition of Done

### DIFF
--- a/packages/components/src/components/Table/Table.tsx
+++ b/packages/components/src/components/Table/Table.tsx
@@ -19,33 +19,13 @@ type TableRootProps = {
      * Whether header should stick to the top when scrolling
      */
     sticky?: 'head' | 'col' | 'both';
-    /**
-     * Optional caption text for the table that appears above it
-     */
-    caption?: string;
-    /**
-     * Whether to make the table take full width of its container
-     * @default true
-     */
-    fullWidth?: boolean;
     children: ReactNode;
     'aria-label'?: string;
     'aria-describedby'?: string;
 };
 
 export const TableRoot = forwardRef<HTMLTableElement, TableRootProps>(
-    (
-        {
-            layout = 'auto',
-            fullWidth = true,
-            sticky,
-            caption,
-            children,
-            'aria-label': ariaLabel,
-            'aria-describedby': ariaDescribedBy,
-        },
-        ref,
-    ) => {
+    ({ layout = 'auto', sticky, children, 'aria-label': ariaLabel, 'aria-describedby': ariaDescribedBy }, ref) => {
         return (
             <div onKeyDown={handleKeyDown} role="grid" tabIndex={-1}>
                 <table
@@ -53,11 +33,9 @@ export const TableRoot = forwardRef<HTMLTableElement, TableRootProps>(
                     className={styles.table}
                     data-layout={layout}
                     data-sticky={sticky}
-                    data-full-width={fullWidth}
                     aria-label={ariaLabel}
                     aria-describedby={ariaDescribedBy}
                 >
-                    {caption && <caption>{caption}</caption>}
                     {children}
                 </table>
             </div>
@@ -65,6 +43,15 @@ export const TableRoot = forwardRef<HTMLTableElement, TableRootProps>(
     },
 );
 TableRoot.displayName = 'Table.Root';
+
+export const TableCaption = forwardRef<HTMLTableCaptionElement, { children: ReactNode }>(({ children }, ref) => {
+    return (
+        <caption ref={ref} className={styles.caption}>
+            {children}
+        </caption>
+    );
+});
+TableCaption.displayName = 'Table.Caption';
 
 type TableHeaderProps = {
     children: ReactNode;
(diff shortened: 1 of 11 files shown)


packages/components/src/components/Table/Table.tsx
```diff
@@ -0,0 +1,369 @@
...
+    },
+);
+TableRow.displayName = 'Table.Row';
+
+type TableRowCellProps = {
+    /**
+     * Whether to truncate content with ellipsis when it overflows
+     * @default false
+     */
+    truncate?: boolean;
```
noahwaldner, line 341: can we please add a title when truncated? so the full text shows on hover

packages/components/src/components/Table/styles/table.module.scss
```diff
@@ -0,0 +1,157 @@
...
+
+    &:focus-visible {
+        @include focus-outline;
+    }
+}
+
+.body {
+    padding: 0 sizeToken(6);
+
+    &[data-sticky-first-column='true'] {
```
noahwaldner, line 89: it doesn't work on the header at the moment. The first column header is still scrolling

noahwaldner (changes requested):

packages/components/src/components/Table/styles/table.module.scss
jose-costa-frontify: it's not meant to, only the content

packages/components/src/components/Table/Table.tsx
```diff
@@ -0,0 +1,395 @@
...
+     */
+    layout?: 'auto' | 'fixed';
+    /**
+     * Whether header should stick to the top when scrolling
+     */
+    sticky?: 'head' | 'col' | 'both';
+    /**
+     * Optional caption text for the table that appears above it
+     */
+    caption?: string;
```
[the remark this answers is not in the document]
jose-costa-frontify: ideally this goes within the table https://www.w3schools.com/tags/tag_caption.asp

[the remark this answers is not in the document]
noahwaldner: caption needs to be inside the <table> so it technically cannot be handled outside. The alternative would be a Slot, but thats most likely overkill

packages/components/src/components/Table/utils.ts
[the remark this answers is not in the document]
jose-costa-frontify: ofc, completely forgot 🤦

packages/components/src/components/Table/Table.tsx
```diff
@@ -0,0 +1,395 @@
...
+    /**
+     * Whether to make the table take full width of its container
+     * @default true
+     */
+    fullWidth?: boolean;
+    children: ReactNode;
+    'aria-label'?: string;
+    'aria-describedby'?: string;
```
noahwaldner, line 33: what abot using `.. & ({ 'aria-label': string } | { 'aria-labelledby': string });` like in the other components so at leas one needs to be added

packages/components/src/components/Table/Table.tsx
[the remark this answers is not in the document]
jose-costa-frontify: Come to think of it, with this API devs can put wtv they want inside the table. I just need to expose a Table.Caption component. Will do that

packages/components/src/components/Table/Table.tsx
```diff
@@ -0,0 +1,395 @@
...
+    sticky?: 'head' | 'col' | 'both';
+    /**
+     * Optional caption text for the table that appears above it
+     */
+    caption?: string;
+    /**
+     * Whether to make the table take full width of its container
+     * @default true
+     */
+    fullWidth?: boolean;
```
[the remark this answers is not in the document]
jose-costa-frontify: Probably yeah

noahwaldner (approved):

packages/components/src/components/Table/Table.tsx
```diff
@@ -0,0 +1,373 @@
...
+                className={styles.headerCell}
+                scope={scope}
+                colSpan={colSpan}
+                data-align={align}
+                data-truncate={truncate}
+                data-no-shrink={noShrink}
+            >
+                {onSortChange ? (
+                    <button className={styles.sortButton} aria-label={sortLabel} onClick={handleSortChange}>
+                        {children}
```
noahwaldner, line 173: when the icon appears and disappears it can cause layout shifts, can you make sure the icon has a reserved space?

noahwaldner: Also: Can we show something then the col is sortable? otherwise the use doesn't know he can click

packages/components/src/components/Table/Table.tsx
```diff
@@ -0,0 +1,373 @@
...
+                style={{ width }}
+                className={styles.headerCell}
+                scope={scope}
+                colSpan={colSpan}
+                data-align={align}
+                data-truncate={truncate}
+                data-no-shrink={noShrink}
+            >
+                {onSortChange ? (
+                    <button className={styles.sortButton} aria-label={sortLabel} onClick={handleSortChange}>
```
noahwaldner, line 172: please move the padding from the th to the button, else not the full cell is clickable

PR #2125, opened 2024-12-04
feat: dropdown improvements

noahwaldner:
- allow for link items in dropdown
- added option to prevent focusing the trigger on close

### DIFF
--- a/packages/components/src/components/Dropdown/Dropdown.tsx
+++ b/packages/components/src/components/Dropdown/Dropdown.tsx
@@ -21,6 +21,7 @@ export type DropdownRootProps = {
      * Callback that is called when the open state of the dropdown changes.
      */
     onOpenChange?: (open: boolean) => void;
+
     'data-test-id'?: string;
 };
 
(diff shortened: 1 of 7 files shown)


packages/components/src/components/Dropdown/Dropdown.tsx
```diff
@@ -93,6 +101,7 @@ export const DropdownContent = (
         padding = 'comfortable',
         align = 'start',
         children,
+        preventTriggerFocusOnClose,
```
SamuelAlev, line 104: Why? it is kinda an accessibility anti-pattern

packages/components/src/components/Dropdown/styles/dropdown.module.scss
```diff
@@ -79,15 +78,11 @@
         color: var(--text-color-disabled);
     }
 
-    .itemContent {
+    span {
```
SamuelAlev, line 81: What if the children is a `p` or `div` tag?

noahwaldner: that would be a custom element, we shouldn't interfere there. This is affecting the text passed in like suggested in our examples / the migration guide

SamuelAlev (approved):

PR #2128, opened 2024-12-05
feat(Flyout): enable trigger positioning in mobile

jose-costa-frontify:


packages/components/src/components/Flyout/Flyout.stories.tsx
```diff
@@ -264,6 +264,39 @@ export const MobileView: Story = {
     },
 };
 
+export const MobileViewBottom: Story = {
```
SamuelAlev, line 267: There is already a mobile view https://fondue-components.frontify.com/?path=/story/current_components-flyout--mobile-view&globals=viewport:mobile

noahwaldner (approved):

packages/components/src/components/Flyout/Flyout.stories.tsx
jose-costa-frontify: the button is at the bottom, not strictly necessary it's just to show what happens when we position the trigger differently

PR #2145, opened 2024-12-17
fix(Dialog): missing granular padding control

jose-costa-frontify:
CU-8697231g6

### DIFF
--- a/packages/components/src/components/Dialog/Dialog.tsx
+++ b/packages/components/src/components/Dialog/Dialog.tsx
@@ -79,7 +79,7 @@ export type DialogTriggerProps = {
 
 export type DialogHeaderProps = {
     /**
-     * Define the padding of the dialog
+     * Define the padding for the dialog header
      * @default "compact"
      */
     padding?: 'none' | 'tight' | 'compact' | 'comfortable' | 'spacious';
--- a/packages/components/src/components/Dialog/Dialog.tsx
+++ b/packages/components/src/components/Dialog/Dialog.tsx
@@ -99,7 +99,7 @@ export type DialogHeaderProps = {
 
 export type DialogFooterProps = {
     /**
-     * Define the padding of the dialog
+     * Define the padding for the dialog footer
      * @default "compact"
      */
     padding?: 'none' | 'tight' | 'compact' | 'comfortable' | 'spacious';
--- a/packages/components/src/components/Dialog/Dialog.tsx
+++ b/packages/components/src/components/Dialog/Dialog.tsx
@@ -33,6 +33,11 @@ export type DialogContentProps = {
      */
     rounded?: boolean;
     /**
+     * Define the padding of the dialog
+     * @default "compact"
+     */
+    padding?: 'none' | 'tight' | 'compact' | 'comfortable' | 'spacious';
+    /**
      * The vertical alignment of the divider
      * @default "center"
      */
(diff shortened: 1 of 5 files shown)


packages/components/src/components/Dialog/Dialog.tsx
```diff
@@ -78,6 +78,11 @@ export type DialogTriggerProps = {
 };
 
 export type DialogHeaderProps = {
+    /**
+     * Define the padding of the dialog
```
noahwaldner, line 82:
```suggestion
     * Define the padding for the dialog header
```

packages/components/src/components/Dialog/Dialog.tsx
```diff
@@ -93,6 +98,11 @@ export type DialogHeaderProps = {
 };
 
 export type DialogFooterProps = {
+    /**
+     * Define the padding of the dialog
```
noahwaldner, line 102:
```suggestion
     * Define the padding for the dialog footer
```

packages/components/src/components/Dialog/Dialog.tsx
```diff
@@ -102,7 +112,15 @@ export type DialogFooterProps = {
     'data-test-id'?: string;
 };
 
-export type DialogBodyProps = { children?: ReactNode; 'data-test-id'?: string };
+export type DialogBodyProps = {
+    /**
+     * Define the padding of the dialog
```
noahwaldner, line 117:
```suggestion
     * Define the padding for the dialog body
```

packages/components/src/components/Dialog/Dialog.tsx
```diff
@@ -206,9 +224,10 @@ DialogContent.displayName = 'Dialog.Content';
 
 export const DialogHeader = (
     {
-        children,
+        padding = 'compact',
```
noahwaldner, line 227: if we set a default value here, wouldn't it will always override the dialog-wide padding?

packages/components/src/components/Dialog/Dialog.tsx
```diff
@@ -251,11 +277,17 @@ export const DialogFooter = (
 DialogFooter.displayName = 'Dialog.Footer';
 
 export const DialogBody = (
-    { children, 'data-test-id': dataTestId = 'fondue-dialog-body' }: DialogBodyProps,
+    { padding = 'compact', children, 'data-test-id': dataTestId = 'fondue-dialog-body' }: DialogBodyProps,
```
noahwaldner, line 280: if we set a default value here, wouldn't it will always override the dialog-wide padding?

packages/components/src/components/Dialog/Dialog.tsx
```diff
@@ -233,14 +253,20 @@ export const DialogHeader = (
 DialogHeader.displayName = 'Dialog.Header';
 
 export const DialogFooter = (
-    { showBorder = true, children, 'data-test-id': dataTestId = 'fondue-dialog-footer' }: DialogFooterProps,
+    {
+        padding = 'compact',
```
noahwaldner, line 257: if we set a default value here, wouldn't it will always override the dialog-wide padding

noahwaldner (changes requested):

packages/components/src/components/Dialog/Dialog.tsx
jose-costa-frontify: very true, last night i just added the global one and went on my way, nice catch

PR #2146, opened 2024-12-17
feat(Tabs): add padding control

jose-costa-frontify:
CU-8697232rc

### DIFF
--- a/packages/components/src/components/Tabs/Tabs.stories.tsx
+++ b/packages/components/src/components/Tabs/Tabs.stories.tsx
@@ -216,4 +216,175 @@ export const WithDecorator: Story = {
         </Tabs.Root>
     ),
 };
+
+export const WithNoPadding: Story = {
+    render: (args) => (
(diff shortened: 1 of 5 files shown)


packages/components/src/components/Tabs/Tabs.stories.tsx
```diff
@@ -216,4 +216,175 @@ export const WithDecorator: Story = {
         </Tabs.Root>
     ),
 };
+
```
noahwaldner, line 219: the storybook controls are not working as the hardcoded padding always overrides the args.

if you switch the order it should work

noahwaldner (changes requested):

packages/components/src/components/Tabs/Tabs.stories.tsx
jose-costa-frontify: fixed

noahwaldner (approved):

PR #2148, opened 2024-12-18
feat(Dropdown): add support for active items

jose-costa-frontify:


SamuelAlev:
Won't do for now, closing
Let's reopen when necessary

PR #2160, opened 2025-01-15
fix: explicitly set pointer-events in datepicker to auto

noahwaldner:
in web-app the body has `pointer-events` set to `none` in some cases. By explicitly setting pointer events to `auto`we ensure the date picker is clickable


syeo66 (approved):
LGTM

PR #2162, opened 2025-01-15
fix(ColorPicker): allow empty alpha field

syeo66:
This way we can delete the field without immediately getting a '0'.


packages/components/src/components/ColorPicker/ColorValueInput.tsx
SamuelAlev, line 1: Please add test coverage

SamuelAlev (changes requested):

PR #2169, opened 2025-01-21
fix: catch null value in onSelect

noahwaldner:


noahwaldner:
Force Merging due to incident

PR #2172, opened 2025-01-23
feat: add `side` to Select component

SamuelAlev:
Add the `side` prop as we offer it in the `Dropdown` already


noahwaldner (commented):
when using side "left" or "right" there is no margin. If we can add this we're good to go

PR #2177, opened 2025-01-27
chore: update deps

SamuelAlev:


SamuelAlev:
Aggregation of Renovate PRs into one, will merge

PR #2178, opened 2025-01-27
chore: update Sass notation for breaking changes in next version

SamuelAlev:
Update the notation to please the Sass compiler as the next major version will break these.


syeo66 (approved):
LGTM

PR #2179, opened 2025-01-27
fix(Dropdown): add missing modal prop (and set to false by default)

jose-costa-frontify:
CU-8697pkzu5

### DIFF
--- a/packages/components/src/components/Dropdown/Dropdown.tsx
+++ b/packages/components/src/components/Dropdown/Dropdown.tsx
@@ -11,6 +11,11 @@ import styles from './styles/dropdown.module.scss';
 export type DropdownRootProps = {
     children?: ReactNode;
     /**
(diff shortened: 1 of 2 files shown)


noahwaldner (approved):

packages/components/src/components/Dropdown/Dropdown.tsx
```diff
@@ -10,6 +10,11 @@ import styles from './styles/dropdown.module.scss';
 
 export type DropdownRootProps = {
     children?: ReactNode;
+    /**
+     * When set to true, interaction with outside elements will be disabled and only menu content will be visible to screen readers.
+     * @default false
```
SamuelAlev, line 15: Why is it set to `false`? Default is `true` in Radix, why would you need to allow interaction with the outside elements?

SamuelAlev (changes requested):
Just blocking to get a better idea why we do that change

SamuelAlev (approved):

PR #2180, opened 2025-01-28
fix(RTE): multi column layout is off in safari

fulopdaniel:
# Safari
#### Before
image

#### After
image

# Chrome
#### Before
image

#### After (unchanged)
image


fulopdaniel:
Seems like an RTE test is failing, gonna investigate and reopen.

PR #2191, opened 2025-02-06
fix: remove leftover index file

noahwaldner:


syeo66 (approved):
I guess this is okay.

PR #2193, opened 2025-02-06
feat: add customizable trigger spacing to `Dropdown`

noahwaldner:


packages/components/src/components/Dropdown/Dropdown.tsx
```diff
@@ -94,20 +99,26 @@ export const DropdownContent = (
...
         preventTriggerFocusOnClose,
         'data-test-id': dataTestId = 'fondue-dropdown-content',
     }: DropdownContentProps,
     ref: ForwardedRef<HTMLDivElement>,
 ) => {
     const theme = useFondueTheme();
+    const spacingMap = {
+        compact: 8,
+        spacious: 12,
+    };
```
SamuelAlev, line 113: Extract out of the component
```suggestion
    const SPACING_MAP: Record<DropdownContentProps['padding'], number> = {
        compact: 8,
        spacious: 12,
    };
    
    export const Dropdown // ....
```

packages/components/src/components/Dropdown/Dropdown.tsx
```diff
@@ -94,20 +99,26 @@ export const DropdownContent = (
     {
         side = 'bottom',
         align = 'start',
+        distance = 'compact',
```
SamuelAlev, line 102: We use `padding` everywhere else
```suggestion
        padding = 'compact',
```

packages/components/src/components/Dropdown/Dropdown.tsx
```diff
@@ -84,30 +86,42 @@ export type DropdownContentProps = {
...
     const theme = useFondueTheme();
+
     return (
         <RadixDropdown.Portal>
             <ThemeProvider theme={theme}>
                 <RadixDropdown.Content
                     align={align}
                     collisionPadding={8}
-                    sideOffset={8}
+                    sideOffset={SPACING_MAP[spacing]}
```
jose-costa-frontify, line 124: very neat

packages/components/src/components/Dropdown/Dropdown.tsx
jose-costa-frontify: as we discussed I was confused by the naming, since this seems to be a very niche configuration that I expect will only be used once in a blue moon I wouldn't be opposed to giving it a braindead name like `triggerOffset` or `triggerGap`

SamuelAlev (approved):

PR #2197, opened 2025-02-07
feat: Add loading state to Table Header Cell

noahwaldner:
- Refactored `LoadingCircle` component
- Added loading state to Table Header Component

### DIFF
--- a/packages/components/src/components/LoadingCircle/LoadingCircle.tsx
+++ b/packages/components/src/components/LoadingCircle/LoadingCircle.tsx
@@ -1,0 +1,41 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+import { forwardRef, type ForwardedRef } from 'react';
+
+import styles from './styles/loadingCircleStyles.module.scss';
+
+export type LoadingCircleProps = {
+    /**
+     * The  variant of the loading circle.
+     * @default 'progress'
+     */
+    variant?: 'progress' | 'success' | 'danger';
+    /**
+     * The size of the loading circle
+     * @default "medium"
+     */
+    size?: 'small' | 'medium' | 'large' | 'x-large';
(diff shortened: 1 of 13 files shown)


packages/components/src/components/LoadingCircle/LoadingCircle.tsx
```diff
@@ -0,0 +1,41 @@
...
+            variant = 'progress',
+            size = 'medium',
+            'data-test-id': dataTestId = 'fondue-loading-circle-content',
+        }: LoadingCircleProps,
+        ref: ForwardedRef<HTMLDivElement>,
+    ) => {
+        return (
+            <div
+                ref={ref}
+                data-test-id={dataTestId}
```
jose-costa-frontify, line 33: we should prolly enhance this for a11y with aria-live and some screen reader text that can be customized via props no?

packages/components/src/components/Table/Table.stories.tsx
```diff
@@ -323,6 +323,82 @@ export const Sortable: Story = {
     },
 };
 
+export const WithLoadingState: Story = {
+    render: ({ ...args }) => {
+        const [sortField, setSortField] = useState<string | null>(null);
+        const [sortDirection, setSortDirection] = useState<'ascending' | 'descending'>();
+
+        const sortedUsers = [...TABLE_DATA].sort((a, b) => {
```
jose-costa-frontify, line 331: we already have a sortable story maybe we can keep this lean and only have the loading prop?

jose-costa-frontify (approved):
Couple nits, not blocking

packages/components/src/components/Table/Table.stories.tsx
noahwaldner: can do

PR #2202, opened 2025-02-10
feat(Text): migrate text component

syeo66:
Screenshot 2025-02-10 at 09 05 33
CU-8695khwkg


packages/components/src/components/Text/Text.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+export type TextProps<TTag extends TagType = 'span'> = CommonAriaProps &
+    DetailedHTMLProps<HTMLAttributes<TextElementType<TTag>>, TextElementType<TTag>> & {
+        size?: TextSize;
+        weight?: TextWeight;
+        as?: TTag;
+        color?: TextColor;
+        /** @description optional color prop that uses the inverse box color when accessibility contrast is needed */
+        boxColor?: BoxColor;
+        children?: ReactNode;
```
SamuelAlev, line 25: Have a look at all other components, they have some kind of docs in the type, please add it too (same in heading)

packages/components/src/components/Text/Text.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+type TextWeight = 'default' | 'strong' | 'x-strong';
+type TextSize = 'x-small' | 'small' | 'medium' | 'large';
+type TextColor = 'default' | 'weak' | 'x-weak' | 'disabled' | 'negative' | 'positive' | 'warning' | 'interactive';
+type BoxColor = 'neutral' | 'selected' | 'disabled' | 'positive' | 'negative' | 'warning';
+
+type TagType = 'a' | 'abbr' | 'address' | 'em' | 'label' | 'li' | 'span' | 'strong' | 'time' | 'p';
+
+export type TextProps<TTag extends TagType = 'span'> = CommonAriaProps &
+    DetailedHTMLProps<HTMLAttributes<TextElementType<TTag>>, TextElementType<TTag>> & {
```
SamuelAlev, line 18: After second thought, not too much of a fan to expose _all_ the props available as it could make maintenance a pain, could we restrict it to only what we need at the moment (same with Heading)

packages/components/src/components/Text/styles/text.module.scss
```diff
@@ -0,0 +1,43 @@
+.root {
```
SamuelAlev, line 1: copyright

SamuelAlev (changes requested):

packages/components/src/components/Text/Text.tsx
syeo66: Will go back to the Header component in a separate PR.

packages/components/src/components/Text/Text.tsx
```diff
@@ -0,0 +1,130 @@
...
+type TextSize = 'x-small' | 'small' | 'medium' | 'large';
+type TextColor = 'default' | 'weak' | 'x-weak' | 'disabled' | 'negative' | 'positive' | 'warning' | 'interactive';
+type BoxColor = 'neutral' | 'selected' | 'disabled' | 'positive' | 'negative' | 'warning';
+
+type TagType = 'a' | 'abbr' | 'address' | 'em' | 'label' | 'li' | 'span' | 'strong' | 'time' | 'p';
+
+export type TextProps<TTag extends TagType = 'span'> = CommonAriaProps &
+    TagProps<TTag> & {
+        /**
+         * `id` represents the id of the element
```
SamuelAlev, line 20: Keep the same structure to the docs 😉 
We know it is the `id`, we are hovering it.

```suggestion
         * The id of the element
```

packages/components/src/components/Text/Text.tsx
```diff
@@ -0,0 +1,130 @@
...
+         */
+        id?: string;
+        /**
+         * `size` is the size of the text
+         *
+         * Values are 'x-small', 'small', 'medium'  or 'large'
+         *
```
SamuelAlev, line 27: Same with values, they are part of the type, unecessary to repeat.

The comment might seems unecessary too, but later on we will be able to add link to the doc/storybook to show the different possibilities.
```suggestion
         * Size of the text
         *
```

packages/components/src/components/Text/Text.tsx
syeo66: Well, it's called id... duh.😂

SamuelAlev (approved):

PR #2207, opened 2025-02-13
fix(Dialog): first input not focused when header contains close button

jose-costa-frontify:
CU-8697y4upp

We must bypass the close button otherwise that gets the focus


syeo66 (approved):
LGTM

PR #2218, opened 2025-02-17
fix: fix broken flyout width on mobile

noahwaldner:
Adding the `ThemeProvider` makes this not be a direct child of the `body`


syeo66 (approved):
LGTM

PR #2222, opened 2025-02-18
feat(Select): add id prop to Select

syeo66:
CU-8697zwv0k

### DIFF
--- a/packages/components/src/components/Label/Label.stories.tsx
+++ b/packages/components/src/components/Label/Label.stories.tsx
@@ -88,8 +88,9 @@ export const WithSelect: Story = {
     render: (args) => {
         return (
             <div className="tw-flex tw-flex-col tw-gap-2">
-                <Label {...args} id="select-label" />
-                <Select placeholder="Select..." labelId="select-label">
+                {/* The select is using a 'div' as a trigger, this means we have to trigger it manually. */}
+                <Label {...args} htmlFor="select" onClick={() => document.getElementById('select')?.click()} />
+                <Select id="select" placeholder="Select...">
                     <Select.Item value="1">Option 1</Select.Item>
                     <Select.Item value="2">Option 2</Select.Item>
                 </Select>
--- a/packages/components/src/components/Label/Label.stories.tsx
+++ b/packages/components/src/components/Label/Label.stories.tsx
@@ -5,6 +5,7 @@ import { type Meta, type StoryObj } from '@storybook/react';
 import { type ComponentProps } from 'react';
 
 import { Checkbox } from '../Checkbox/Checkbox';
+import { Select } from '../Select/Select';
 import { TextInput } from '../TextInput/TextInput';
 import { Tooltip } from '../Tooltip/Tooltip';
 
@@ -37,8 +38,8 @@ export const Mandatory: Story = {
 
         return (
             <div className="tw-flex tw-flex-col tw-gap-2">
-                <Label {...args} htmlFor="input" />
-                <STextInput id="input" placeholder="Enter your name" />
+                <Label {...args} htmlFor="input-mandatory" />
+                <STextInput id="input-mandatory" placeholder="Enter your name" />
             </div>
         );
     },
@@ -55,8 +56,8 @@ export const WithTextInput: Story = {
 
         return (
             <div className="tw-flex tw-flex-col tw-gap-2">
-                <Label {...args} htmlFor="input" />
-                <STextInput id="input" placeholder="Enter your name" />
+                <Label {...args} htmlFor="input-text" />
+                <STextInput id="input-text" placeholder="Enter your name" />
             </div>
         );
     },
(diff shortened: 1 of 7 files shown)


packages/components/src/components/Select/Select.tsx
SamuelAlev, line 1: Can you add a story in the `Label.stories.tsx`?

packages/components/src/components/Select/Select.tsx
```diff
@@ -135,6 +140,7 @@ export const SelectInput = (
                               'aria-label': ariaLabel,
                               ...(forwardedRef ? { ref: forwardedRef } : {}),
                           }))}
+                    {...(id ? { id } : {})}
```
jose-costa-frontify, line 143: I think we don't need to spread the id this way here we can just do 
```suggestion
                    id={id}
```
no?

syeo66: No, I only want to override the id if it is actually set outside (and I even feel bad about that)

syeo66: Screenshot 2025-02-19 at 10 20 34

If you look at the generated html you'll notice the Radix component already take quite a lot of control.

packages/components/src/components/Label/Label.stories.tsx
```diff
@@ -73,8 +74,26 @@ export const WithDisabledInput: Story = {
...
+
+export const WithSelect: Story = {
+    args: {
+        children: 'Hello World',
+    },
+    render: (args) => {
+        return (
+            <div className="tw-flex tw-flex-col tw-gap-2">
+                {/* The select is using a 'div' as a trigger, this means we have to trigger it manually. */}
+                <Label {...args} htmlFor="select" onClick={() => document.getElementById('select')?.click()} />
```
SamuelAlev, line 92: Would be awesome to not have to manually do it (not intuitive) but rather find a way to communicate it to downshift

packages/components/src/components/Select/Select.tsx
SamuelAlev, line 1: Can be added to the Combobox component (same folder) too

packages/components/src/components/Label/Label.stories.tsx
syeo66: No idea how this would work.

syeo66: It's not intuitive either way. But reversed not so you can leverage an id on the label from the select.

packages/components/src/components/Select/Select.tsx
```diff
@@ -110,6 +114,29 @@ export const SelectInput = (
...
+        if (!labelId) {
+            return;
+        }
+
+        document.getElementById(labelId)?.addEventListener('click', labelClickHandler);
+
+        return () => {
+            document.getElementById(labelId)?.removeEventListener('click', labelClickHandler);
+        };
+    }, [labelId, toggleButtonId]);
```
SamuelAlev, line 141: Pls add it in the label component
```tsx
data-test-id={dataTestId}
onClick={(event) => {
    // Add support of Select component, Radix only allows native `select`
    // but we use `div[role=combobox]` because of downshift
    // https://github.com/radix-ui/primitives/blob/6e75e117977c9e6ffa939e6951a707f16ba0f95e/packages/react/label/src/label.tsx#L22
    const targetId = (event.target as HTMLElement).getAttribute('for');
    const target = targetId ? document.getElementById(targetId) : null;

    if (target && target.getAttribute('role') === 'combobox') {
        event.preventDefault();
        event.stopPropagation();
        target.click();
    }
}}
{...props}
```

packages/components/src/components/Select/Combobox.tsx
```diff
@@ -57,6 +66,14 @@ export type ComboboxProps = {
...
      */
     'data-test-id'?: string;
+    /**
+     * id of the select component
+     */
+    id?: string;
+    /**
+     * id of a label which should trigger the combobox
+     */
+    labelId?: string;
```
SamuelAlev, line 76:
```tsx
& (
    | {
          /**
           * The ID of the label element.
           */
          'aria-labelledby': string;
      }
    | {
          /**
           * The aria label of the combobox component.
           */
          'aria-label': string;
      }
)
```

packages/components/src/components/Select/Select.tsx
```diff
@@ -61,6 +61,14 @@ export type SelectComponentProps = {
...
      */
     'data-test-id'?: string;
+    /**
+     * id of the select component
+     */
+    id?: string;
+    /**
+     * id of a label which should trigger the select
+     */
+    labelId?: string;
```
SamuelAlev, line 71:
```tsx
& (
    | {
          /**
           * The ID of the label element.
           */
          'aria-labelledby': string;
      }
    | {
          /**
           * The aria label of the select component.
           */
          'aria-label': string;
      }
)
```

packages/components/src/components/Select/Combobox.tsx
```diff
@@ -91,6 +110,8 @@ export const SelectCombobox = (
         highlightedIndex,
         inputValue,
     } = useCombobox({
+        labelId,
+        id,
```
SamuelAlev, line 114:
```suggestion
        toggleButtonId: id,
```
Then you can remove the `id={toggleButtonId}` and put back the spread

SamuelAlev (changes requested):

packages/components/src/components/Select/Select.tsx
syeo66: Not really a fan of either solution. But okay, this makes it closest to the native behaviour for the developers.

packages/components/src/components/Select/Combobox.tsx
syeo66: Typing it like that would make at least one of the aria props mandatory. Had to do a bit more to make them both optional but exclusive.

jose-costa-frontify: we have the `CommonAriaAttrs`

syeo66: Which just sets both props optional. I don't know... I don't care... @SamuelAlev ?

syeo66: Okay, the fancy typing causes more trouble than it is worth. So, I just went with the CommonAriaProps route

PR #2224, opened 2025-02-19
fix: label and select/combobox

SamuelAlev:


SamuelAlev:
Applied in the main PR, closing

PR #2230, opened 2025-02-25
feat: add customizable triggerOffset for `Flyout` + improve accordion deprecation

noahwaldner:
- add triggerOffset prop to `Flyout`Component
- improve deprecation notices for `Accordion`Component


packages/components/src/components/Flyout/Flyout.tsx
```diff
@@ -98,11 +99,22 @@ export type FlyoutContentProps = {
      * Define the maximum width of the flyout
      * @default "360px"
      */
+    /**
```
syeo66, line 102: The comment above should probable be moved down.

syeo66 (approved):

PR #2237, opened 2025-02-27
feat: add scrollbar gutter

noahwaldner:
- add scrollbar gutter prop

### DIFF
--- a/packages/components/src/components/ScrollArea/styles/scrollArea.module.scss
+++ b/packages/components/src/components/ScrollArea/styles/scrollArea.module.scss
@@ -14,7 +14,7 @@
 .viewport {
     width: 100%;
     height: 100%;
-    padding-right: var(--scrollbar-size);
+    background-color: red;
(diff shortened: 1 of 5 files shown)


packages/components/src/components/ScrollArea/styles/scrollArea.module.scss
```diff
@@ -14,7 +14,7 @@
 .viewport {
     width: 100%;
     height: 100%;
-    padding-right: var(--scrollbar-size);
+    background-color: red;
```
syeo66, line 17: Hm, debugging leftover?

syeo66 (approved):
I guess you should check the background color. Other than that it looks goot to me.

packages/components/src/components/ScrollArea/styles/scrollArea.module.scss
noahwaldner: yep, removed

PR #2239, opened 2025-02-28
feat(Select): Add size prop to Select

syeo66:
Add more padding. Used on the error page builder:
Screenshot 2025-02-28 at 13 01 26

Screenshot 2025-02-28 at 13 07 26

CU-86982zvzh


syeo66:
According to the communication with Joanna we no longer need/support this.

PR #2242, opened 2025-02-28
refactor(Modal): remove default z-index from legacy modal

syeo66:
See https://github.com/Frontify/fondue/pull/2240#discussion_r1975472680

CU-86982zvzh

### DIFF
--- a/.changeset/red-plums-repeat.md
+++ b/.changeset/red-plums-repeat.md
@@ -1,0 +1,5 @@
+---
+"@frontify/fondue": minor
+---
+
+refactor(Modal): remove default z-index from legacy modal
--- a/packages/fondue/src/components/Modal/Modal.tsx
+++ b/packages/fondue/src/components/Modal/Modal.tsx
@@ -72,7 +72,7 @@ const ModalComponent = memo((props: ModalProps): ReactElement => {
             exit="exit"
             style={{
                 background: 'rgba(0, 0, 0, .5)',
-                ...(zIndex ? { zIndex } : {}),
+                zIndex,
             }}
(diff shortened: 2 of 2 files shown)


.changeset/red-plums-repeat.md
syeo66, line 1: No, I did not choose the filename. 🤣

packages/fondue/src/components/Modal/Modal.tsx
```diff
@@ -46,13 +46,11 @@ const widthMap: Record<ModalWidth, string> = {
     [ModalWidth.Large]: 'tw-max-w-[1200px]',
 };
 
-const DEFAULT_ZINDEX = 50;
```
SamuelAlev, line 49: Really cool! I remember few places where I had to hardcode `zIndex=0` in the web-app

packages/fondue/src/components/Modal/Modal.tsx
```diff
@@ -74,7 +72,7 @@ const ModalComponent = memo((props: ModalProps): ReactElement => {
             exit="exit"
             style={{
                 background: 'rgba(0, 0, 0, .5)',
-                zIndex,
+                ...(zIndex ? { zIndex } : {}),
```
SamuelAlev, line 75: Is there a reason why we change that line?
if not set, it will be `undefined` and then it won't be added

syeo66: I usually don't trust in magic. But sure. 👍

SamuelAlev (approved):

PR #2245, opened 2025-03-03
fix(Accordion): add box sizing to accordion trigger

syeo66:


packages/components/src/components/Accordion/Accordion.tsx
```diff
@@ -121,6 +121,7 @@ AccordionHeader.displayName = 'Accordion.Header';
 
 export type AccordionTriggerProps = {
     'data-test-id'?: string;
+    asChild?: boolean;
```
SamuelAlev, line 124: JSDoc missing + `@default <true/false>`

SamuelAlev (approved):

PR #2248, opened 2025-03-05
feat(SegmentedControl): add tooltip capability

syeo66:
Screenshot 2025-03-05 at 13 48 57
(Note: Keyboard navigation, thus border... was easier to screenshot)

### DIFF
--- a/packages/components/src/components/SegmentedControl/SegmentedControl.stories.tsx
+++ b/packages/components/src/components/SegmentedControl/SegmentedControl.stories.tsx
@@ -45,6 +45,22 @@ export const TextOnly: Story = {
     ),
 };
 
+export const WithTooltip: Story = {
+    render: (args) => (
+        <SegmentedControl.Root {...args}>
+            <SegmentedControl.Item tooltip="First One" value="first">
+                First
+            </SegmentedControl.Item>
+            <SegmentedControl.Item tooltip="Second One" value="second">
+                Second
+            </SegmentedControl.Item>
+            <SegmentedControl.Item tooltip="Third One" value="third">
+                Third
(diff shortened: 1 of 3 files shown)


packages/components/src/components/SegmentedControl/styles/segmentedControl.module.scss
```diff
@@ -164,7 +164,8 @@ $radius-without-border: calc(var(--radius) - var(--line-width));
     }
 
     // Show the active indicator when the state is on
-    .item[data-state='on'] ~ & {
+    .item[data-state='on'] ~ &,
```
syeo66, line 167: Since the tooltip and the segmented control both set 'data-state' I had to rely on aria-checked for the segmented-control state in those cases.

packages/components/src/components/SegmentedControl/SegmentedControl.stories.tsx
```diff
@@ -45,6 +45,22 @@ export const TextOnly: Story = {
     ),
 };
 
+export const WithTooltip: Story = {
+    render: (args) => (
+        <SegmentedControl.Root {...args}>
+            <SegmentedControl.Item tooltip="First One" value="first">
+                First
+            </SegmentedControl.Item>
```
SamuelAlev, line 53: Not much in the mindset of all the components we have
```suggestion
            <Tooltip.Root>
              <Tooltip.Trigger asChild>
                <SegmentedControl.Item tooltip="First One" value="first">
                  First
                </SegmentedControl.Item>
              </Tooltip.Trigger>
              ....
```

Let's fix it the other way around (aka in Tooltip comp) if there is an issue

PR #2251, opened 2025-03-06
fix(Select): adjacent groups are missing a divider line

syeo66:
CU-8698725jf


jose-costa-frontify (approved):
If you have time plz fix the inconsistencies with the margins

PR #2252, opened 2025-03-06
feat: add scrollable flyout body

noahwaldner:
- add prop to allow scrolling in body

CU-86985rx47

### DIFF
--- a/packages/components/src/components/Flyout/Flyout.tsx
+++ b/packages/components/src/components/Flyout/Flyout.tsx
@@ -203,7 +203,15 @@ export const FlyoutFooter = (
 };
 FlyoutFooter.displayName = 'Flyout.Footer';
 
-export type FlyoutBodyProps = { children?: ReactNode; 'data-test-id'?: string; scrollable?: boolean };
+export type FlyoutBodyProps = {
+    children?: ReactNode;
+    'data-test-id'?: string;
+    /**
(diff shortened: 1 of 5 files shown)


syeo66 (approved):
LGTM

packages/components/src/components/Flyout/Flyout.tsx
```diff
@@ -203,14 +203,28 @@ export const FlyoutFooter = (
...
 
-export type FlyoutBodyProps = { children?: ReactNode; 'data-test-id'?: string };
+export type FlyoutBodyProps = {
+    children?: ReactNode;
+    'data-test-id'?: string;
+    /**
+     * Allow the body to scroll if the max height of the flyout is reached
+     * @default false
+     */
+    scrollable?: boolean;
```
SamuelAlev, line 213: Shouldn't it just be the default? Also why does `ScrollArea` doesn't work in the body?

noahwaldner: not necessarily, there are cases where teams want the whole to scroll instead of having sticky headers and footers

If we want to use scroll area here then we would need to conditionally add a scrolll area. This solution is css only and does the job just fine, as we do not need horizontal scroll, control over scrollbar rendering, etc.
I would not make this more complex than it needs to be

PR #2261, opened 2025-03-14
feat: add customizable collision padding

noahwaldner:
allow for customizing the collision padding

CU-86985rr4t


packages/components/src/components/Flyout/Flyout.tsx
```diff
@@ -105,6 +106,11 @@ export type FlyoutContentProps = {
      * @default "360px"
      */
     maxWidth?: string;
+    /**
+     * Define the minimum distance between the flyout and the viewport edge
+     * @default 8
```
syeo66, line 111:
```suggestion
     * @default 'default'
```

packages/components/src/components/Dropdown/Dropdown.tsx
```diff
@@ -94,6 +94,11 @@ export type DropdownContentProps = {
      * Prevents the focus from being set on the trigger when the dropdown is closed.
      */
     preventTriggerFocusOnClose?: boolean;
+    /**
+     * Define the minimum distance between the dropdown and the viewport edge
+     * @default 8
```
syeo66, line 99: Default 8 from 'default' | 'large'?

```suggestion
     * @default 'default'
```

syeo66 (approved):

packages/components/src/components/Select/SelectMenu.tsx
```diff
@@ -73,19 +80,25 @@ export const SelectMenu = ({
...
+
     const theme = useFondueTheme();
     return (
         <RadixPopover.Portal>
             <ThemeProvider theme={theme}>
                 <RadixPopover.Content
                     align={align}
                     side={side}
-                    collisionPadding={16}
+                    collisionPadding={VIEWPORT_COLLISION_PADDING_MAP[viewportCollisionPadding] + 8}
```
jose-costa-frontify, line 101: why the additional 8px here?

jose-costa-frontify (approved):

packages/components/src/components/Flyout/Flyout.stories.tsx
```diff
@@ -413,3 +413,22 @@ export const SpaciousTriggerOffset: Story = {
...
+        viewportCollisionPadding: 'spacious',
+    },
+    render: ({ ...args }) => {
+        return (
+            <Flyout.Root>
+                <Flyout.Trigger>
+                    <Button>Open flyout</Button>
+                </Flyout.Trigger>
+                <Flyout.Content {...args}>
+                    <Flyout.Body {...args} />
```
jose-costa-frontify, line 429: atm the effect of this prop is not visible in the stories, maybe we could make it a bit more clear by having 2 stories for each padding and a big lorem ipsum inside the flyout, dropdown etc so that we can see at the right edge the difference this prop makes

packages/components/src/components/Select/SelectMenu.tsx
noahwaldner: required because of the padding addid in css, radix doesn't know about it when calculating the size of the popover

packages/components/src/components/Flyout/Flyout.stories.tsx
noahwaldner: Screenshot 2025-03-19 at 07 29 56

Verified that its vidible

PR #2279, opened 2025-04-04
feat: add utility for variable font italics

noahwaldner:
for italics, use variable font features instead of broiwser font-style

### DIFF
--- a/packages/tokens/src/formatters/tailwind/template.ts
+++ b/packages/tokens/src/formatters/tailwind/template.ts
@@ -13,10 +13,19 @@ export const template = ({ theme, plugin }: TemplateOptions) => {
       prefix: "tw-",
       theme: ${JSON.stringify(theme)},
       plugins: [
-      plugin(function({ addComponents}) {
-        const components = ${JSON.stringify(plugin)};
-        addComponents(components);
-        })
+        plugin(function({ addComponents, addUtilities}) {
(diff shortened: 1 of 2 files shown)


packages/tokens/src/formatters/tailwind/template.ts
```diff
@@ -13,10 +13,19 @@ export const template = ({ theme, plugin }: TemplateOptions) => {
...
-        const components = ${JSON.stringify(plugin)};
-        addComponents(components);
-        })
+        plugin(function({ addComponents, addUtilities}) {
+            const components = ${JSON.stringify(plugin)};
+            addComponents(components);
+        }),
+        plugin(({ addUtilities }) => {
+            addUtilities({
+                '.italic, i, em': {
```
SamuelAlev, line 22: Does it apply to every `i` & `em`? Quite dangerous, better to apply that config in web-app styles

noahwaldner: this is already done at the moment in `reset.css`

Screenshot 2025-04-04 at 10 03 16

SamuelAlev: That's a resetter, it doesn't apply font specific settings like `font-feature-settings: '"ss01"';`

SamuelAlev (approved):

PR #2281, opened 2025-04-04
fix: switch handle animates from outside

noahwaldner:
Container width calculation on load caused a glitch in the position

- removed usage of container query

CU-8698hhujv

### DIFF
--- a/packages/components/src/components/Switch/styles/switch.module.scss
+++ b/packages/components/src/components/Switch/styles/switch.module.scss
@@ -6,8 +6,7 @@
 
 .root {
     position: relative;
-    container-type: size;
-    border-radius: 9999px;
+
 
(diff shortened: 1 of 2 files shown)


packages/components/src/components/Switch/styles/switch.module.scss
```diff
@@ -6,36 +6,42 @@
 
 .root {
     position: relative;
-    container-type: size;
-    border-radius: 9999px;
+
```
SamuelAlev, line 9: Could you remove all the extra line breaks?

noahwaldner: removed

jose-costa-frontify (approved):

PR #2283, opened 2025-04-04
fix(RTE): fix a11y issues with the checkbox not having a proper label

peter-tudosa:


ragi96 (commented):
Not fully sure but probably the serializer is missing @fulopdaniel

peter-tudosa:
> Not fully sure but probably the serializer is missing @fulopdaniel

Thanks for the heads up @ragi96! We have to tweak that as well as far as I can tell, I'm on it!

.changeset/solid-worlds-dance.md
```diff
@@ -0,0 +1,5 @@
+---
+"@frontify/fondue": minor
```
ragi96, line 2:
```suggestion
"@frontify/fondue": patch
```

.changeset/solid-worlds-dance.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+a11y fixes for RTE
```
ragi96, line 5:
```suggestion
fix(RTE): improved a11y for the CheckboxListPlugin
```

fulopdaniel:
Couple things I noticed:
1. cursor is now default instead of text, when hovering label text.
2. in edit mode it is difficult to select text without ticking the checkbox accidentally.

To fix these I would actually avoid wrapping the whole thing in `<label>` and would just rather use an `aria-labelledby` or `aria-label` attribute on the input, as that also fixes the a11y issue we have.

What do you think?

packages/fondue/src/components/RichTextEditor/serializer/utils/makeIdGenerator.ts
```diff
@@ -0,0 +1,10 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+export const makeIdGenerator = (prefixToken?: string) => {
+    let seedId = 0;
+    const prefix = prefixToken ?? +new Date();
```
SamuelAlev, line 5: https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID

PR #2284, opened 2025-04-07
feat(RTE): expose placeholder opacity

fulopdaniel:


packages/fondue/src/components/RichTextEditor/RichTextEditor.tsx
```diff
@@ -50,6 +51,7 @@ export const RichTextEditor = ({
     border = true,
     toolbarWidth,
     hideExternalFloatingModals,
+    placeholderOpacity = 0.333,
```
jose-costa-frontify, line 54: why do we need this level of control? idk why this is being added but I assume we should at most have something like 

```suggestion
    transparency = 'medium', // medium, low, high
```

packages/fondue/src/components/RichTextEditor/components/RenderPlaceholder/RenderPlaceholder.tsx
```diff
@@ -8,12 +8,13 @@ const PLACEHOLDER_STYLES: CSSProperties = {
     height: '0',
 };
 
-export const RenderPlaceholder = ({ children, attributes }: RenderPlaceholderProps) => {
+export const RenderPlaceholder = ({ children, attributes, opacity }: RenderPlaceholderProps & { opacity: number }) => {
```
jose-costa-frontify, line 11: if we have a new prop in `RenderPlaceholder` we should add it to `RenderPlaceholderProps`

packages/fondue/src/components/RichTextEditor/components/RenderPlaceholder/RenderPlaceholder.tsx
```diff
@@ -8,12 +8,13 @@ const PLACEHOLDER_STYLES: CSSProperties = {
...
     const mergedAttributes = {
         ...attributes,
         style: {
             ...attributes.style,
```
jose-costa-frontify, line 15: does this not already accept any style? why send opacity separately?

fulopdaniel: `attributes.style` comes directly from the 3rd party package, we have no control over that.

packages/fondue/src/components/RichTextEditor/RichTextEditor.tsx
fulopdaniel: Well, there is a bit of a trial and error going on trying to make it compatible with different color schemes. I figured it will be easier to give full control, and then we don't need to touch fondue if we decide to change.

Right now we want it to be `0.7` instead of the default `0.333`, but since RTE is used other places as well, not just guidelines, I did not want to just override directly.

I can add the low/medium/high options if you worry it's too much freedom.

fulopdaniel: Gotcha, added

PR #2288, opened 2025-04-09
fix: revert vite 

jose-costa-frontify:


SamuelAlev:
Also need in `icons` package, done in #2289 (also aligns all deps)

PR #2302, opened 2025-05-12
WIP: feat revamped token generation

noahwaldner:
**Work in progress**


packages/components/.storybook/components/StoryWithTeme.tsx
syeo66, line 1: File name looks funny

packages/tokens/.gitignore
```diff
@@ -86,7 +86,6 @@ typings/
 
 # Nuxt.js build / generate output
 .nuxt
-dist
```
SamuelAlev, line 89: Could you add back the dist folder to the ignored files?

packages/components/src/components/Tokens/previews/Utility.tsx
```diff
@@ -0,0 +1,25 @@
...
+import { type Utility } from '../components/Utilities/types';
+
+export const getUtilityClassName = ({ path }: Utility) => {
+    return `${path.join('-').replace('-default', '')}`;
+};
+
+export const UtilityPreview = ({ properties }: Utility) => {
+    const { textContent } = usePreviewConfig();
+
+    const style = Object.entries(properties).reduce(
```
syeo66, line 13:
```suggestion
    const style = Object.entries(properties).reduce<Record<string, string>>(
```

packages/components/src/components/Tokens/previews/Utility.tsx
```diff
@@ -0,0 +1,25 @@
...
+        (acc, [key, value]) => {
+            acc[key.replaceAll(/-./g, (match) => match.slice(-1).toUpperCase())] = `var(--${value.name})`;
+            return acc;
+        },
+        {} as Record<string, string>,
```
syeo66, line 18:
```suggestion
        {},
```

packages/fondue/src/subpackages/tokens.ts
```diff
@@ -0,0 +1,3 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+/* (c) Copyright Frontify Ltd., all rights reserved. */
```
syeo66, line 3:
```suggestion

```

syeo66 (approved):

packages/tokens/.gitignore
noahwaldner: This is an intentional change. 

Previously we manually copied the tokens from figma to the tokens package.

As we now automatically fetch tokens when building the package, there would be no way way of knowing which tokens were included in a previous state. Additionally, everyone who wants to run fondue locally would need a Figma API key and access to our figma file to be able to build the tokens package, which is required to run the other packages.
By adding the dist folder to version control, we can track what tokens are actually exported at any time.
The alternative would be to version some intermediate format of the tokens fetched from figma.

This also allows us to have a ci script that automatically fetches figma and creates PRs with the updated tokens

PR #2308, opened 2025-05-23
feat: deprecate drag delay

joelharkes:
We had a bug report that was related to this.

See PR in web-app to fix it: https://github.com/Frontify/web-app/pull/9055

I think the delaySeconds will only confuse people and you have to manually disable it, as you already have a 5px check they delay is just confusing people: https://app.clickup.com/t/2523021/TASK-14618

I deprecated the feature and removed the usage for it.


jcosta33 (approved):

syeo66:
Hm, actually it seems like you are deprecating and removing it at the same time. Wouldn't deprecating mean there is a grace period before removing it?

packages/fondue/src/components/Tree/types.ts
```diff
@@ -139,6 +139,7 @@ export type TreeProps = {
     selectedIds?: string[];
     expandedIds?: string[];
     dragHandlerPosition?: DragHandlerPosition;
+    /** @deprecated unused, property was used to at 150 ms delay when dragHandlerPosition is 'none'. but could stop registering trags. */
     enableDragDelay?: boolean;
```
SamuelAlev, line 143: Tbh it is way more confusing to keep it but unused.
Since it might be used externally, keep the behavior as it was. But you can change the default behavior when we use the Tree component and it is unset (`enableDragDelay === undefined`).

SamuelAlev (changes requested):

PR #2309, opened 2025-05-26
feat: Add `disabled` property to color picker input

jprats89:
Add a `disabled` property to the `ColorPickerInput` so color picker is not actionable for users with no permission.

**Why?**
For the backend library view sidebar, we need to "disable" all the fields for viewers.

TASK-14613

### DIFF
--- a/.changeset/smooth-hats-sleep.md
+++ b/.changeset/smooth-hats-sleep.md
@@ -1,0 +1,5 @@
+---
+"@frontify/fondue-components": minor
+---
+
+Add `disabled` property to ColorPickerInput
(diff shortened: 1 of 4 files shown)


.changeset/smooth-hats-sleep.md
```diff
@@ -0,0 +1,5 @@
+---
+"@frontify/fondue-components": minor
+---
+
+Add `disabled` property to ColorPickerInput
```
noahwaldner, line 5:
```suggestion
Add `disabled` property to `ColorPicker.Input`
```

PR #2318, opened 2025-06-05
fix: Clicking outside flyout not triggering other elements

joelharkes:
CU-86994t2ye

bug reported in LibraryList

Clicking outside the Flyout should not trigger things outside, instead it should only close the Flyout. 

(having this issue right now in Library list.)

### DIFF
--- a/.changeset/upset-glasses-rule.md
+++ b/.changeset/upset-glasses-rule.md
@@ -1,0 +1,5 @@
+---
+"@frontify/fondue-components": minor
+---
+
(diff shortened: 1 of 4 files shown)


joelharkes:
I couldn't find anything working. only an example of adding a backdrop like this might work:
https://github.com/radix-ui/primitives/issues/1679#issuecomment-2531180521

noahwaldner (approved):

joelharkes:
Double check if the click might leak through to but it doesn't seem the case :)

PR #2320, opened 2025-06-24
chore(Tabs): add trigger click indicator

syeo66:
CU-8699aep44

### DIFF
--- a/.changeset/clever-symbols-bet.md
+++ b/.changeset/clever-symbols-bet.md
@@ -1,0 +1,5 @@
+---
+"@frontify/fondue-components": patch
+---
+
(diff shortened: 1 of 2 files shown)


.changeset/clever-symbols-bet.md
```diff
@@ -0,0 +1,5 @@
+---
+"@frontify/fondue-components": patch
+---
+
+chore(Tabs): add trigger click indicator aligned with webapp navigation
```
SamuelAlev, line 5: web app doesn't exist in the context of Fondue
```suggestion
chore(Tabs): add trigger click indicator aligned with [...]
```

Same with PR title

SamuelAlev (approved):

PR #2323, opened 2025-06-25
fix(Tabs): changing trigger content did not rerender

syeo66:
CU-8699g8pd0


packages/components/src/components/Tabs/hooks/useTabTriggers.tsx
```diff
@@ -120,13 +120,23 @@ export const useTabTriggers = ({
         }
     }, [triggers, triggerListRef, activeIndicatorRef]);
 
+    const addTrigger = (trigger: TabTrigger) => {
```
SamuelAlev, line 123: Test case to prevent regression?

PR #2331, opened 2025-06-30
feat(Select): add heading to select groups

syeo66:
Screenshot 2025-06-30 at 13 07 33

CU-8699kkzgp

### DIFF
--- a/packages/components/src/components/Select/SelectItem.tsx
+++ b/packages/components/src/components/Select/SelectItem.tsx
@@ -73,7 +73,7 @@ export const SelectItemGroup = (
         <div data-test-id={dataTestId} className={styles.group} ref={forwardedRef} key={groupId}>
             {heading ? (
                 <div className={styles.groupHeading}>
-                    <span aria-label={heading}>{heading}</span>
+                    <span>{heading}</span>
                 </div>
             ) : null}
             {children}
--- a/packages/components/src/components/Select/SelectItem.tsx
+++ b/packages/components/src/components/Select/SelectItem.tsx
@@ -56,17 +56,26 @@ export type SelectItemGroupProps = {
      */
     groupId: string;
     /**
+     * The groups heading
+     */
+    heading?: string;
+    /**
      * The data test id of the select item group.
      */
     'data-test-id'?: string;
 };
 
 export const SelectItemGroup = (
-    { children, groupId, 'data-test-id': dataTestId = 'fondue-select-item-group' }: SelectItemGroupProps,
+    { children, groupId, heading, 'data-test-id': dataTestId = 'fondue-select-item-group' }: SelectItemGroupProps,
     forwardedRef?: ForwardedRef<HTMLDivElement>,
 ) => {
     return (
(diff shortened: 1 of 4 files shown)


packages/components/src/components/Select/SelectItem.tsx
```diff
@@ -55,18 +55,27 @@ export type SelectItemGroupProps = {
...
 export const SelectItemGroup = (
-    { children, groupId, 'data-test-id': dataTestId = 'fondue-select-item-group' }: SelectItemGroupProps,
+    { children, groupId, heading, 'data-test-id': dataTestId = 'fondue-select-item-group' }: SelectItemGroupProps,
     forwardedRef?: ForwardedRef<HTMLDivElement>,
 ) => {
     return (
         <div data-test-id={dataTestId} className={styles.group} ref={forwardedRef} key={groupId}>
+            {heading ? (
+                <div className={styles.groupHeading}>
+                    <span aria-label={heading}>{heading}</span>
```
jcosta33, line 76: I think we don't need the aria label here

```suggestion
                    <span>{heading}</span>
```

packages/components/src/components/Select/SelectItem.tsx
```diff
@@ -55,18 +55,27 @@ export type SelectItemGroupProps = {
...
 
 export const SelectItemGroup = (
-    { children, groupId, 'data-test-id': dataTestId = 'fondue-select-item-group' }: SelectItemGroupProps,
+    { children, groupId, heading, 'data-test-id': dataTestId = 'fondue-select-item-group' }: SelectItemGroupProps,
     forwardedRef?: ForwardedRef<HTMLDivElement>,
 ) => {
     return (
         <div data-test-id={dataTestId} className={styles.group} ref={forwardedRef} key={groupId}>
+            {heading ? (
+                <div className={styles.groupHeading}>
```
jcosta33, line 75:
```suggestion
                <div id={labelId} className={styles.groupHeading}>
```

packages/components/src/components/Select/SelectItem.tsx
```diff
@@ -55,18 +55,27 @@ export type SelectItemGroupProps = {
...
     'data-test-id'?: string;
 };
 
 export const SelectItemGroup = (
-    { children, groupId, 'data-test-id': dataTestId = 'fondue-select-item-group' }: SelectItemGroupProps,
+    { children, groupId, heading, 'data-test-id': dataTestId = 'fondue-select-item-group' }: SelectItemGroupProps,
     forwardedRef?: ForwardedRef<HTMLDivElement>,
 ) => {
     return (
         <div data-test-id={dataTestId} className={styles.group} ref={forwardedRef} key={groupId}>
```
jcosta33, line 73:
```suggestion
        <div role="group" aria-labelledby={labelId} data-test-id={dataTestId} className={styles.group} ref={forwardedRef} key={groupId}>
```

jcosta33 (approved):
just some a11y stuff to consider

packages/components/src/components/Select/SelectItem.tsx
jcosta33: probably don't need the `span` even

PR #2334, opened 2025-06-30
fix: prevent button and link clicks inside <Table.Row> from triggering row click

mbgspcii:
**What:**
This PR fixes how clicks work in the `<Table.Row>` component.
If the `click` prop is defined for <Table.Row>:
- clicking a button or link inside a cell will only trigger that button or link’s click
- clicking on a cell without a button or link will trigger the row’s click.

**Why:**
Previously, clicks on interactive elements like buttons or links inside a table row would also fire the row’s click handler.

### DIFF
--- a/packages/components/src/components/Table/Table.tsx
+++ b/packages/components/src/components/Table/Table.tsx
@@ -12,15 +12,15 @@ import {
     type ReactNode,
 } from 'react';
 
+import { useSyncRefs } from '#/hooks/useSyncRefs';
+import { useTextTruncation } from '#/hooks/useTextTruncation';
+import { type CommonAriaAttrs } from '#/utilities/types';
+
 import { Box } from '../Box/Box';
 import { LoadingCircle } from '../LoadingCircle/LoadingCircle';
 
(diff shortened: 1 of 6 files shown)


packages/components/src/components/Table/Table.tsx
```diff
@@ -321,11 +322,16 @@ export const TableRow = forwardRef<HTMLTableRowElement, TableRowProps>(
     ) => {
         const isInteractive = onClick !== undefined && !disabled;
 
-        const handleClick = () => {
+        const handleClick = (event?: MouseEvent<HTMLTableRowElement>) => {
             if (disabled) {
                 return;
             }
 
+            const clickedCell = event?.target instanceof Element ? event.target.closest('td') : null;
```
syeo66, line 330: Why not use event.target directly and check if it is a button or a link? As it is implemented now we'd prevent handling of clicks in the whitespace of a cell just because the cell contains any of those things.

mbgspcii: Yes you're right gg.

PR #2344, opened 2025-07-14
feat: add basic `RouterProvider` and `Link`

SamuelAlev:
Preview: https://fondue-components.frontify.com/pr-preview/pr-2344/?path=/docs/current_typography-link--documentation

Adding first version of a `RouterProvider` and `Link` component (~there is no final design as of today, so I copy/pasted the design of `Text`~).
Agnostic of the router we use, it can be integrated like this:

#### React Router example
```tsx
	import { useHref, useNavigate } from 'react-router-dom';

	const navigate = useNavigate();

	return (
		<RouterProvider navigate={navigate} useHref={useHref}>
			// [...]
			<Link href="/foo">Hello world</Link>
			// [...]
		</RouterProvider>
	);
```

#### Vanilla history example
```tsx
	function useHref(to, basename = '') {
		const base = basename.endsWith('/') ? basename.slice(0, -1) : basename;
		const path = to.startsWith('/') ? to : '/' + to;
		return base + path;
	}

	return (
		<RouterProvider navigate={(url)=> (history.pushState(null, null, url))} useHref={useHref}>
			// [...]
			<Link href="/foo">Hello world</Link>
			// [...]
		</RouterProvider>
	);
```

CU-8699kkh61

### DIFF
--- a/.github/workflows/continuous-deployment.yaml
+++ b/.github/workflows/continuous-deployment.yaml
@@ -137,7 +137,7 @@ jobs:
     component-tests:
         name: Component Tests
         runs-on: ubuntu-latest
-        timeout-minutes: 10
+        timeout-minutes: 25
 
         steps:
             - name: Checkout current commit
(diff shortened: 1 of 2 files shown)


syeo66 (approved):
LGTM

PR #2391, opened 2025-08-26
feat: accordion allow setting header padding and adding action

noahwaldner:
- Add `Accordion.Slot` to allow a action in the header
- Combine `Accordion.Trigger` and `Accordion.Header`
- Allow passing a padding prop on `Accordion.Root` to set the paddings for Header & Content

CU-869a3phev
CU-869a1qhfm


syeo66 (approved):
LGTM

PR #2395, opened 2025-08-26
feat(charts): update tokens [CU-869a8984f]

joseazevedo88:
### Branched from `v13` - Update the tokens according to [design](https://www.figma.com/design/jAwm8whm4UZrKn6ogLB88b/Analytics---Access-Management-Color-Token-Adaption?node-id=38-10558&t=BX2QufMd43y8X409-4)

image

image

image

image

image


joseazevedo88:
@noahwaldner do you know why in the deployed preview in the url above the text is not white? Seems the class `tw-text-base` is not working, even though it's working locally, as seen in the screenshots above.

image

image

PR #2398, opened 2025-08-28
feat: implement new RadioList component

syeo66:


noahwaldner:
https://github.com/user-attachments/assets/598aee0e-a088-4eef-b8d6-38d926a49b5b

Changing the text weight makes the entire list shift a bit to the left and right. could we possibly add some opacity trickery and render a hidden text to avoid this?


Also, can you please mark the old component as deprecated?

packages/components/src/components/Label/Label.tsx
```diff
@@ -53,7 +57,14 @@ export const LabelComponent = (
...
             }}
             {...props}
-        />
+        >
+            {/* Hidden version with medium font weight to reserve space */}
+            <span className="tw-font-medium tw-opacity-0 tw-pointer-events-none" aria-hidden="true">
+                {children}
+            </span>
+            {/* Visible version (inherits all styling from parent) */}
+            <span className="tw-absolute tw-inset-0">{children}</span>
```
SamuelAlev, line 66: Display grid with stacking would be a more robust solution than absolute IMO: https://x.com/wesbos/status/1866895899554926912

Feel free to skip it but let's add tests then

noahwaldner: prefferably we would just use scss for the entirety of the components, then we can at some point remove tailwind from the prod bundle

syeo66: Yes, that's what I'm saying. I'll move the Label away from Tailwind. But not in this PR.

PR #2434, opened 2025-09-10
feat(RTE): remove cmd+e shortcut as it clashes with toggle view/edit keyboard shortcut

peter-tudosa:
Ticket: https://app.clickup.com/t/869abwnja


packages/fondue/src/components/RichTextEditor/Plugins/MarkPlugin/CodePlugin/index.ts
```diff
@@ -19,7 +19,13 @@ export class CodePlugin extends Plugin {
     }
 
     plugins(): PlatePlugin[] {
-        return [createCodePlugin()];
+        return [
+            createCodePlugin({
+                options: {
+                    hotkey: [],
```
SamuelAlev, line 25: `cmd+shift+c` / `ctrl+shift+c` instead, so we are not just removing the shortcut

PR #2464, opened 2025-10-07
feat(Dialog): implement RTL support

syeo66:
Screenshot 2025-10-07 at 10 48 53


packages/components/src/components/Dialog/Dialog.stories.tsx
```diff
@@ -687,3 +687,35 @@ export const WithTooltip: Story = {
         </Dialog.Root>
     ),
 };
+
+export const RTL: Story = {
+    args: {
+        children: 'I am a dialog content',
+        dir: 'rtl',
```
SamuelAlev, line 694: Instead of passing as props, maybe we can use a `<div dir="rtl"`?

PR #2467, opened 2025-10-07
feat(RadioList): support RTL

syeo66:
Screenshot 2025-10-07 at 15 40 10


packages/components/src/components/RadioList/RadioList.tsx
```diff
@@ -33,14 +34,28 @@ const RadioListRoot = ({
     value,
     'data-test-id': dataTestId,
 }: RadioListRootProps) => {
+    const [dir, setDir] = useState<'ltr' | 'rtl' | undefined>(undefined);
+    const rootRef = useRef<HTMLDivElement>(null);
```
SamuelAlev, line 38:
```suggestion
    const [dir, setDir] = useState<'ltr' | 'rtl' | undefined>(() => window.document.dir));
```
?

PR #2698, opened 2026-04-27
fix: handle meta keys in select key listener

noahwaldner:
- Ignore key commands on `Select` when a meta key is held down
- Use correct cursor for combobox element

### DIFF
--- a/packages/components/src/components/Select/components/ComboboxBase.tsx
+++ b/packages/components/src/components/Select/components/ComboboxBase.tsx
@@ -312,6 +312,11 @@ const ComboboxBaseInput = (
                                     // eslint-disable-next-line react-hooks/refs
                                     {...getInputProps({
                                         ref: inputCallbackRef,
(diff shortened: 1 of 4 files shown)


packages/components/src/components/Select/components/ComboboxBase.tsx
```diff
@@ -312,6 +312,11 @@ const ComboboxBaseInput = (
                                     // eslint-disable-next-line react-hooks/refs
                                     {...getInputProps({
                                         ref: inputCallbackRef,
+                                        onKeyDown: (event: React.KeyboardEvent) => {
```
SamuelAlev, line 315:
```suggestion
                                        onKeyDown: (event: KeyboardEvent) => {
```
same for the others + "import from react" up

SamuelAlev (approved):
Small comment left to fix before merging, but to not slow you down, already approving.

PR #2701, opened 2026-04-27
fix: bundling issues

jcosta33:
## What

Fixes `OrderableList` drag choreography (items shifting to make space) not working when `@frontify/fondue-components` is consumed from the published package, and resolves the regression introduced by the previous attempt at this fix.

## Why

The previous PR moved `vite.config.ts` from array-form externalization (exact-string match) to a function that matches declared deps and their subpaths. That correctly externalized `@dnd-kit/react/sortable` so consumers and the library share runtime instances (dnd-kit's registry uses `instanceof` identity, which silently fails when two copies exist).

That fix had a side effect: it also externalized `@frontify/fondue-tokens/themes`, which is a CSS Module subpath. Externalizing CSS Module subpaths leaks bundler-specific behavior into every downstream toolchain:

- Rspack/webpack `css-loader` v6+ doesn't emit a `default` export for CSS Modules → `ESModulesLinkingError: export 'default' (imported as 'p') was not found in '@frontify/fondue-tokens/themes'`.
- Vitest in Node can't load `.css` files natively → `TypeError: Unknown file extension ".css"`.

## Fix

Replace `rollupOptions.external` with a `resolveId`-based plugin that:

- externalizes JS subpaths of declared deps + peer deps (preserves the dnd-kit fix);
- inlines stylesheet subpaths even when their parent dep is external — they have no shared runtime to share, and externalizing them just leaks CSS Modules quirks into every consumer;
- only runs for the library build (`apply: build.lib`) so Storybook builds stay unaffected.

### DIFF
--- a/packages/components/vite.config.ts
+++ b/packages/components/vite.config.ts
@@ -1,17 +1,39 @@
 /* (c) Copyright Frontify Ltd., all rights reserved. */
 
 import react from '@vitejs/plugin-react';
+import { type Plugin } from 'vite';
 import dts from 'vite-plugin-dts';
 import tsConfigPaths from 'vite-tsconfig-paths';
 import { configDefaults, defineConfig } from 'vitest/config';
 
 import { dependencies as dependenciesMap, peerDependencies as peerDependenciesMap } from './package.json';
 
-const peerDependencies = Object.keys(peerDependenciesMap);
-const dependencies = Object.keys(dependenciesMap);
-const externalPackages = [...dependencies, ...peerDependencies, 'react-dom/client', 'react/jsx-runtime'];
+const externalCandidates = [...Object.keys(dependenciesMap), ...Object.keys(peerDependenciesMap)];
+const matchesExternalCandidate = (id: string) =>
+    externalCandidates.some((pkg) => id === pkg || id.startsWith(`${pkg}/`));
 
-const isExternal = (id: string) => externalPackages.some((pkg) => id === pkg || id.startsWith(`${pkg}/`));
+const STYLE_FILE = /\.(css|scss|sass|less|styl)(\?|$)/;
+
+// Externalize JS deps (incl. subpaths) so consumers share runtime instances; inline
+// stylesheet subpaths since they have no shared runtime to share.
+const externalizeJsDeps = (): Plugin => ({
+    name: 'externalize-js-deps',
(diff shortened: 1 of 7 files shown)


packages/components/vite.config.ts
```diff
@@ -1,17 +1,39 @@
...
-const isExternal = (id: string) => externalPackages.some((pkg) => id === pkg || id.startsWith(`${pkg}/`));
+const STYLE_FILE = /\.(css|scss|sass|less|styl)(\?|$)/;
+
+// Externalize JS deps (incl. subpaths) so consumers share runtime instances; inline
+// stylesheet subpaths since they have no shared runtime to share.
+const externalizeJsDeps = (): Plugin => ({
+    name: 'externalize-js-deps',
+    enforce: 'pre',
+    // Library build only. Storybook merges this config and would break on bare imports.
+    apply: (config, env) => env.command === 'build' && config.build?.lib !== undefined,
```
SamuelAlev, line 23: What do you think to flip the way we compute for storybook + make the plugin unaware of storybook, like this?
```ts
const isStorybook = process.argv[1]?.includes("storybook");

// ...

plugins: [!isStorybook && externalizeJsDeps()]
```

jcosta33: isn't it already unaware of storybook as is though?

SamuelAlev: You run it only in build (`env.command === 'build'`), what about other commands? Separation of concerns, the plugin just doesn't have to know about storybook + it simplifies the plugins logic by moving that out

jcosta33: it is not supposed to run for other commands though. honestly this feels like a legitimate use of `apply` here. Tying the execution of the plugin to storybook with the conditional append  rather than saying "you are only meant to EVER run for build (which is a pretty standard vite command which will never die)" feels like more of a violation of separation of concerns since at least IMO since what the plugin does and when it does it are extremely tied together since externalization only makes sense for build. The `vite-plugin-externalize-deps` also only runs on build though they use a different config to get there.

SamuelAlev (approved):

packages/components/vite.config.ts
SamuelAlev: /fine-by-me

PR #2702, opened 2026-04-28
fix: prevent input clicks and keydown events inside <Table.Row> from triggering row click [CU-869d2t9c6]

mbgspcii:
## What
This PR updates how keyboard and mouse events work in the `<Table.Row>` component.

If the `onClick` prop is defined for `<Table.Row>`:
- clicking inside an input cell will only focus the input, not trigger the row's click
- pressing `Space` or `Enter` inside an input cell will only trigger the input's native behavior and no more trigger the row's click

## Why
Previously, clicks and keyboard events from input cells would bubble up to the row and fire the row's click handler, making it difficult to interact with inputs inside a table row.

## Issue in video

https://github.com/user-attachments/assets/884c790d-a098-4820-a791-2e4b8a264cac

### DIFF
--- a/.changeset/itchy-tips-unite.md
+++ b/.changeset/itchy-tips-unite.md
@@ -1,0 +1,5 @@
+---
+"@frontify/fondue-components": patch
+---
+
+fix: prevent input clicks and keydown events inside <Table.Row> from triggering row click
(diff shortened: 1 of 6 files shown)


noahwaldner (approved):

mbgspcii:
@noahwaldner thanks++ for the quick review.

PR #2705, opened 2026-04-29
feat: allow configuring locale via `ThemeProvider`

noahwaldner:
Changes:
- Rename `tranlsations` to `locale` (Because it also contains the date locale) 
  - Add `locale` prop to `ThemeProvider`
  - Mark `translations`prop as deprecated
- Adg missing package export for fondue locale objects
- Pass locale to `ThemeProviders`inside portals
- Add storybook global to set locale in the ui

Additional fixes:
- added missing css rule in `DatePicker` to set color to token from theme - Fixes wrong text color in dark theme


storybook-docs/.storybook/preview.ts
```diff
@@ -79,8 +81,18 @@ export const globalTypes = {
...
+            icon: 'globe',
+            items: availableLocales.map((locale) => ({ value: locale, title: locale })),
+            dynamicTitle: true,
+        },
+    },
 };
 
 export const initialGlobals = {
     theme: 'light',
+    locale: 'en-US',
```
SamuelAlev, line 96: Could you add some guidance on how to lazy load the locale in the storybook?
We don't want to bundle all the locales by default otherwise code-splitting breaks

packages/fondue/package.json
```diff
@@ -30,6 +30,11 @@
             "import": "./dist/packages/components/fondue-components.js"
         },
         "./components/styles": "./dist/packages/components/style.css",
+        "./locales": {
```
SamuelAlev: Should be `/components/locales` (since `@frontify/fondue-components/locales`) if we follow the rest of the subexports

packages/components/src/components/Dropdown/Dropdown.tsx
```diff
@@ -279,11 +279,11 @@ export const DropdownSubContent = (
     { children, 'data-test-id': dataTestId = 'fondue-dropdown-subcontent' }: DropdownSubContentProps,
     ref: ForwardedRef<HTMLDivElement>,
 ) => {
-    const { theme, dir } = useFondueTheme();
+    const { theme, dir, locale } = useFondueTheme();
 
     return (
         <RadixDropdown.Portal>
-            <ThemeProvider theme={theme} dir={dir}>
+            <ThemeProvider theme={theme} dir={dir} locale={locale}>
```
SamuelAlev, line 295: Either here or in separate PR, you might want to look into the last few new comp (DatePicker and OrderableList, there might be few hardcoded string, like `aria-label` in `packages/components/src/components/DatePicker/DatePickerCalendar.tsx`)

storybook-docs/.storybook/preview.ts
noahwaldner: Added a Storybook story for the `ThemeProvider` component with all docs

noahwaldner: Modified the exports to also support importing from `@frontify/fondue/components/locales/de-DE` as only webpack/rspack supports splitting modules up into chunks by named exports

packages/components/src/components/Dropdown/Dropdown.tsx
noahwaldner: Will do in a follow up

SamuelAlev (approved):

PR #2720, opened 2026-05-11
feat(Accordion): add sticky flag to make headers sticky in ScrollArea

syeo66:
## Summary

Adds an opt-in `sticky` prop on `Accordion.Root`. When enabled, each `Accordion.Header` pins to the top of the nearest scrolling ancestor (e.g. a `ScrollArea`) while its item is in view, then is pushed out by the next item's header — the classic sticky-section-header pattern.

```tsx
<ScrollArea maxHeight={300}>
    <Accordion.Root sticky>
        <Accordion.Item value="…">
            <Accordion.Header>…</Accordion.Header>
            <Accordion.Content>…</Accordion.Content>
        </Accordion.Item>
        {/* … */}
    </Accordion.Root>
</ScrollArea>
```

Default is `false`, so existing usages are unchanged.

### DIFF
--- a/packages/components/src/components/Accordion/Accordion.tsx
+++ b/packages/components/src/components/Accordion/Accordion.tsx
@@ -46,6 +46,12 @@ export type AccordionRootProps = {
      */
     padding?: AccordionPadding;
     /**
+     * When `true`, each `Accordion.Header` becomes sticky and pins to the top of the nearest
+     * scrolling ancestor (e.g. a `ScrollArea`) while its item is in view.
+     * @default false
+     */
+    sticky?: boolean;
+    /**
      * Callback function that is called when the value of the accordion changes.
      */
     onValueChange?: (value: string[]) => void;
@@ -61,6 +67,7 @@ export const AccordionRoot = forwardRef<HTMLDivElement, AccordionRootProps>(
             disabled,
             value,
             padding = 'large',
+            sticky = false,
             onValueChange,
         }: AccordionRootProps,
         ref: ForwardedRef<HTMLDivElement>,
@@ -76,6 +83,7 @@ export const AccordionRoot = forwardRef<HTMLDivElement, AccordionRootProps>(
                 value={value}
                 data-border={border}
                 data-accordion-padding={padding}
+                data-sticky={sticky}
                 onValueChange={onValueChange}
             >
                 {children}
--- a/packages/components/src/components/Accordion/Accordion.tsx
+++ b/packages/components/src/components/Accordion/Accordion.tsx
@@ -9,7 +9,6 @@ import {
     isValidElement,
     useCallback,
     useContext,
-    useEffect,
     useMemo,
     useRef,
     type ForwardedRef,
@@ -20,7 +19,6 @@ import {
 import styles from './styles/accordion.module.scss';
 
 type AccordionRootContextValue = {
-    sticky: boolean;
     registerItem: (value: string, element: HTMLDivElement | null) => void;
 };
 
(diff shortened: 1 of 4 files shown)


noahwaldner:
We should probably add some functionality for scroll restoration. Because of the changing vertical size of each item, the active item vanishes when collapsing. In long sections (like theme settings) this would seem confusing to the user.

Measuring if the current scroll position is bigger than the Accordion.Headers distance from the top of the accordion would tell us if the item is in "sticky state". We could then manually set the scroll position so it's still top aligned after closing.

The difficulty with this would be to get the current scroll position if the scrollarea is added by the consumer

Video to demonstrate the issue.
https://github.com/user-attachments/assets/a8df4a49-b82a-4c33-bb31-f9c1aefe9546

syeo66:
> We should probably add some functionality for scroll restoration. Because of the changing vertical size of each item, the active item vanishes when collapsing. In long sections (like theme settings) this would seem confusing to the user.

@noahwaldner I think I got a reasonable solution now

packages/components/src/components/Accordion/Accordion.tsx
```diff
@@ -61,25 +109,73 @@ export const AccordionRoot = forwardRef<HTMLDivElement, AccordionRootProps>(
...
+        const registerItem = useCallback((itemValue: string, element: HTMLDivElement | null) => {
+            if (element) {
+                itemsRef.current.set(itemValue, element);
+            } else {
+                itemsRef.current.delete(itemValue);
+            }
+        }, []);
+
+        const contextValue = useMemo<AccordionRootContextValue>(
+            () => ({ sticky, registerItem }),
```
noahwaldner: `sticky` seems to be only be used in the root, so it doesn't need to be in the context

packages/components/src/components/Accordion/Accordion.tsx
```diff
@@ -61,25 +109,73 @@ export const AccordionRoot = forwardRef<HTMLDivElement, AccordionRootProps>(
             disabled,
             value,
             padding = 'large',
+            sticky = false,
             onValueChange,
         }: AccordionRootProps,
         ref: ForwardedRef<HTMLDivElement>,
     ) => {
+        const itemsRef = useRef<Map<string, HTMLDivElement>>(new Map());
+        const previousValueRef = useRef<string[]>(value ?? defaultValue ?? []);
```
noahwaldner: do we really need the `previousValueRef`? it always tracks the external value, could we just use that to compare the state?

packages/components/src/components/Accordion/Accordion.tsx
```diff
@@ -61,25 +109,73 @@ export const AccordionRoot = forwardRef<HTMLDivElement, AccordionRootProps>(
...
+                itemsRef.current.delete(itemValue);
+            }
+        }, []);
+
+        const contextValue = useMemo<AccordionRootContextValue>(
+            () => ({ sticky, registerItem }),
+            [sticky, registerItem],
+        );
+
+        const handleValueChange = (newValue: string[]) => {
```
noahwaldner: maybe wrap this function in `useCallback`?

packages/components/src/components/Accordion/Accordion.tsx
```diff
@@ -4,16 +4,58 @@ import { IconCaretDown } from '@frontify/fondue-icons';
...
+    if (!scrollContainer) {
+        return;
+    }
+
+    const itemRect = itemEl.getBoundingClientRect();
+    const containerRect = scrollContainer.getBoundingClientRect();
+    const delta = itemRect.top - containerRect.top;
+
+    // Only adjust if the item is scrolled above (or exactly at) the top of the
+    // scroll container — i.e. its header is currently pinned in sticky state.
```
noahwaldner: leftover ai comments

packages/components/src/components/Accordion/Accordion.tsx
```diff
@@ -4,16 +4,58 @@ import { IconCaretDown } from '@frontify/fondue-icons';
...
+    if (delta < 0) {
+        scrollContainer.scrollTo({ top: scrollContainer.scrollTop + delta, behavior: 'smooth' });
```
noahwaldner: this moves the scroll out of sync with the css animation, which can cause a "overshoot". See video below.

https://github.com/user-attachments/assets/c43ada51-0f63-40c6-af15-30dc0976fc92

noahwaldner: We need to track the animation progress to stay in sync. 

Something like:
```
const step = () => {
        const currentHeight = contentEl.offsetHeight;
        const progress = Math.min(Math.max(1 - currentHeight / startHeight, 0), 1);
        scrollContainer.scrollTop = startTop + delta * progress;
        if (currentHeight > 0 && progress < 1) {
            requestAnimationFrame(step);
        }
    };
requestAnimationFrame(step);
```

packages/components/src/components/Accordion/Accordion.tsx
```diff
@@ -108,9 +204,24 @@ export const AccordionItem = forwardRef<HTMLDivElement, AccordionItemProps>(
...
+                    ref.current = element;
+                }
+            },
+            [registerItem, ref, value],
+        );
+
         return (
             <RadixAccordion.Item
-                ref={ref}
+                ref={setRefs}
```
noahwaldner, line 233: nice one! 👏🏻

good idea to use a ref function

packages/components/src/components/Accordion/styles/accordion.module.scss
```diff
@@ -60,6 +64,14 @@
             grid-row: 1;
         }
     }
+
+    .root[data-sticky='true'] & {
+        position: sticky;
+        top: 0;
+        z-index: 1;
+        background-color: var(--color-surface-default);
+        box-shadow: 0 1px 0 var(--color-line-subtle);
```
noahwaldner, line 73: could we use `var(--shadow-default)` ?

syeo66: It's not actually a shadow. It's a line. Using box-shadow to not change the size of the box.

packages/components/src/components/Accordion/Accordion.tsx
syeo66: Well, yes, sort of. We still need to actively deal with the defaultValue. But then we don't have to track it in a useEffect.

noahwaldner (approved):

PR #2723, opened 2026-05-12
feat: add scopeClassName to ThemeProvider and merge context values

noahwaldner:
## Added `scopeClassName` prop

This prop can be used to set a classname for scoped styles, such as guideline blocks.
The value is passed to all portals through the theme context, so also apply in dialogs etc.

I avoided using the name `className` as it is misleading in what the prop does. It does not set the classname of the Provider component, but add one additional class to the existing set.

## Context merging

Previously, every nested `ThemeProvider` overrode the theme context with the default values.
We now merge the context and only override the values explicitly passed to the nested `ThemeProvider`

Screenshot 2026-05-12 at 11 13 39

### DIFF
--- a/packages/components/src/components/ThemeProvider/ThemeProvider.tsx
+++ b/packages/components/src/components/ThemeProvider/ThemeProvider.tsx
@@ -38,6 +38,12 @@ type ThemeProviderProps = {
      */
     locale?: LocaleConfig;
     /**
+     * Additional class name to apply to the theme provider, used to scope styles to a specific component or section of the application.
+     * The class will be propagated to components' portal targets.
+     * @default ""
+     */
+    scopeClassName?: string;
+    /**
      * Change the default rendered element for the one passed as a child, merging their props and behavior.
      * @default false
      */
@@ -48,12 +54,14 @@ type ThemeContextValue = {
     theme: AvailableTheme;
     dir: 'ltr' | 'rtl';
     locale: LocaleConfig;
+    scopeClassName: string;
 };
 
 export const ThemeContext = createContext<ThemeContextValue>({
     theme: 'light',
     dir: 'ltr',
     locale: enUS,
+    scopeClassName: '',
 });
 ThemeContext.displayName = 'ThemeContext';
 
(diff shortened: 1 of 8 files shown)


packages/components/src/components/ThemeProvider/ThemeProvider.tsx
```diff
@@ -37,6 +37,12 @@ type ThemeProviderProps = {
      * ```
      */
     locale?: LocaleConfig;
+    /**
+     * Additional class name to apply to the theme provider, used to scope styles to a specific component or section of the application.
+     * The class is propagated to portaled content (e.g. Dropdown, Tooltip, Dialog) so scoped styles are still applied.
+     * @default ""
+     */
+    scopeClassName?: string;
```
SamuelAlev: > but add one additional class to the existing set.

Thats exactly `className`? There is no need to explain + scope means nothing really, what do you think?

noahwaldner: Screenshot 2026-05-12 at 12 08 52

noahwaldner: I'm open to renaming, but just "className" might be misleading

packages/components/src/components/ThemeProvider/ThemeProvider.tsx
```diff
@@ -68,23 +76,32 @@ export const useFondueTheme = () => {
...
+            [dir, theme, locale, translations, scopeClassName, existingContext],
         );
 
         return (
             <ThemeContext.Provider value={contextValue}>
-                <Comp ref={forwardedRef} dir={dir} className={`${styles[theme]} fondue-theme-provider`}>
+                <Comp
+                    ref={forwardedRef}
+                    dir={contextValue.dir}
+                    className={['fondue-theme-provider', styles[contextValue.theme], contextValue.scopeClassName].join(
```
SamuelAlev: Since it does add a class name, there is nothing about scope here

SamuelAlev (approved):

PR #2726, opened 2026-05-13
feat(AssetInput): add AssetInput component

fulopdaniel:
This PR contributes a rewritten version of the legacy `AssetInput` to the `components` package.

Example usages:

### Placeholder

```
<AssetInput.Placeholder>
    <AssetInput.UploadInput onFileChange={fn()} />
    <AssetInput.BrowseInput onBrowse={fn()} />
</AssetInput.Placeholder>
```
image

### Preview

```
<AssetInput.Root orientation="horizontal" isOpen={false} onPress={fn()}>
    <AssetInput.Preview>
        <AssetInput.PreviewImage src="https://picsum.photos/100/150" />
    </AssetInput.Preview>
    <AssetInput.Title>foo1</AssetInput.Title>
    <AssetInput.Metadata>
        <Flex align="center" gap={0.5}>
            <IconArrowCircleUp size="16" />
            Uploaded
        </Flex>
    </AssetInput.Metadata>
    <AssetInput.Metadata>JPG</AssetInput.Metadata>
    <AssetInput.Metadata>2000 bytes</AssetInput.Metadata>
</AssetInput.Root>
```
image

See storybook for more examples.


packages/components/src/components/AssetInput/AssetInputPlaceholder.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+import { Flex } from '../Flex/Flex';
+
+import { AssetInputBrowseInput } from './AssetInputBrowseInput';
+import { AssetInputUploadInput } from './AssetInputUploadInput';
+import styles from './styles/asset-input.module.scss';
+
+export type AssetInputPlaceholderProps = { children: ReactNode };
+
+export const AssetInputPlaceholder = ({ children }: AssetInputPlaceholderProps) => {
```
noahwaldner: This works but seems overkill

We could ensure only the valid components are passed in with a type and avoid parsing all children

packages/components/src/components/AssetInput/AssetInputPlaceholder.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+    if (inputs.length === 0) {
+        return null;
+    }
+
+    const [firstInput, secondInput] = inputs;
+
+    return (
+        <div className={styles.placeholder}>
+            <Flex gap={2} height="100%" p={3}>
```
noahwaldner, line 20: We could use data attributes passed to the buttons to set the order with css. that way we ensure that upload is always on the left

packages/components/src/components/AssetInput/AssetInputPlaceholder.tsx
```diff
@@ -0,0 +1,42 @@
...
+                {firstInput}
+                {secondInput ? <div className={styles.separator} /> : null}
```
noahwaldner: we could add the separator with css to avoid iterating over the children

something like `.placeholder:has([data-asset-input-action='upload']) [data-asset-input-action='browse']` and add a `:before`

packages/components/src/components/AssetInput/AssetInputBrowseInput.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+type AssetInputBrowseInputProps = {
+    onBrowse: () => void;
+};
+
+export const AssetInputBrowseInput = ({ onBrowse }: AssetInputBrowseInputProps) => {
+    return (
+        <Button onPress={onBrowse} emphasis="weak" hugWidth={false}>
+            <IconImageStack size={20} />
+            Browse
```
noahwaldner: needs translation

you can check `Switch.tsx` for a example

packages/components/src/components/AssetInput/AssetInputUploadInput.tsx
```diff
@@ -0,0 +1,45 @@
...
+        if (fileInputRef.current) {
+            fileInputRef.current.click();
+        }
+    }, [fileInputRef]);
+
+    return (
+        <>
+            <Button onPress={openFileUploadDialog} emphasis="weak" hugWidth={false}>
+                <IconArrowCircleUp size={20} />
+                Upload
```
noahwaldner: needs translation

you can check Switch.tsx for a example

packages/components/src/components/AssetInput/AssetInputUploadInput.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+import { IconArrowCircleUp } from '@frontify/fondue-icons';
+import { useCallback, useId, useRef } from 'react';
+
+import { Button } from '../Button/Button';
+
+type AssetInputUploadInputProps = {
+    acceptFileType?: string;
+    allowMultiple?: boolean;
+    onFileChange: (event: React.ChangeEvent<HTMLInputElement>) => void;
```
noahwaldner: Do not use `React.`, import the ChangeEvent

packages/components/src/components/AssetInput/AssetInputUploadInput.tsx
```diff
@@ -0,0 +1,45 @@
...
+};
+
+export const AssetInputUploadInput = ({
+    acceptFileType,
+    allowMultiple = false,
+    onFileChange,
+}: AssetInputUploadInputProps) => {
+    const fileInputRef = useRef<HTMLInputElement | null>(null);
+    const id = useId();
```
noahwaldner: never used

packages/components/src/components/AssetInput/AssetInputUploadInput.tsx
```diff
@@ -0,0 +1,45 @@
...
+    const openFileUploadDialog = useCallback(() => {
```
noahwaldner: unnecessary callback

packages/components/src/components/AssetInput/AssetInputUploadInput.tsx
```diff
@@ -0,0 +1,45 @@
...
+type AssetInputUploadInputProps = {
+    acceptFileType?: string;
+    allowMultiple?: boolean;
+    onFileChange: (event: React.ChangeEvent<HTMLInputElement>) => void;
+};
+
+export const AssetInputUploadInput = ({
+    acceptFileType,
+    allowMultiple = false,
+    onFileChange,
```
noahwaldner: onSelect makes more sense `imo`

packages/components/src/components/AssetInput/AssetInputRoot.tsx
```diff
@@ -0,0 +1,81 @@
...
+import { AssetInputMetadata } from './AssetInputMetadata';
+import { AssetInputPreview } from './AssetInputPreview';
+import { AssetInputTitle } from './AssetInputTitle';
+import styles from './styles/asset-input.module.scss';
+
+export type AssetInputOrientation = 'horizontal' | 'vertical';
+
+export type AssetInputRootProps = {
+    children: ReactNode;
+    orientation: AssetInputOrientation;
```
noahwaldner: should be optional as there is a default

packages/components/src/components/AssetInput/AssetInputRoot.tsx
```diff
@@ -0,0 +1,81 @@
...
+    isOpen: boolean;
```
noahwaldner: has a default, should it be optional?

packages/components/src/components/AssetInput/AssetInputRoot.tsx
```diff
@@ -0,0 +1,81 @@
...
+    onPress: () => void;
```
noahwaldner: can be optional

packages/components/src/components/AssetInput/AssetInputRoot.tsx
```diff
@@ -0,0 +1,81 @@
...
+};
+
+export const AssetInputRoot = ({
+    children,
+    orientation = 'horizontal',
+    isOpen = false,
+    onPress,
+}: AssetInputRootProps) => {
+    const { parts, metadata } = useMemo(() => {
```
noahwaldner: This causes all children to be parsed everytime

Let's use a css grid with `grid-template-areas` instead
As all components are controlled by us, we can define their `grid-area` properties manually

The only exception is `AssetInput.Metadata`, as there can be multiple
If we use Multiple MetadataItem components and use AssetInput.metadata as a container, we can do it with pure css and avopid parsing children

packages/components/src/components/AssetInput/AssetInputPreview.tsx
```diff
@@ -0,0 +1,43 @@
...
+import { AssetInputPreviewImage } from './AssetInputPreviewImage';
+import { AssetInputPreviewLoading } from './AssetInputPreviewLoading';
+import styles from './styles/asset-input.module.scss';
+
+type AssetInputPreviewProps = {
+    children: ReactNode;
+};
+
+export const AssetInputPreview = ({ children }: AssetInputPreviewProps) => {
+    const previewParts = useMemo(() => {
```
noahwaldner: not necessary, we can use types to ensure only the correct children are passed

packages/components/src/components/AssetInput/AssetInputPreview.tsx
```diff
@@ -0,0 +1,43 @@
...
+            }
+            if (
+                child.type === AssetInputPreviewImage ||
+                child.type === AssetInputPreviewIcon ||
+                child.type === AssetInputPreviewLoading
+            ) {
+                previewParts.push(child);
+            }
+        });
+        if (previewParts.length > 1 && previewParts.length < 4) {
```
noahwaldner: use a constant to define the number of shown previews to avoid inconsistenies

packages/components/src/components/AssetInput/AssetInputPreview.tsx
```diff
@@ -0,0 +1,43 @@
...
+            const emptyParts = Array.from({ length: 4 - previewParts.length }, (_, i) => (
+                <div className={styles.previewSlot} key={`empty-part-${i}`} />
```
noahwaldner: not necessary, can be done by setting a bg on the preview container

packages/components/src/components/AssetInput/AssetInputPreview.tsx
```diff
@@ -0,0 +1,43 @@
...
+            ));
+            previewParts.push(...emptyParts);
+        }
+        return previewParts.slice(0, 4);
+    }, [children]);
+
+    return (
+        <div data-multiple-parts={previewParts.length > 1} className={styles.preview}>
```
noahwaldner: data-multiple-parts can be css driven

`.preview:has(> :nth-child(2))`

packages/components/src/components/AssetInput/AssetInputPreview.tsx
```diff
@@ -0,0 +1,43 @@
...
+import { AssetInputPreviewIcon } from './AssetInputPreviewIcon';
+import { AssetInputPreviewImage } from './AssetInputPreviewImage';
+import { AssetInputPreviewLoading } from './AssetInputPreviewLoading';
+import styles from './styles/asset-input.module.scss';
+
+type AssetInputPreviewProps = {
+    children: ReactNode;
+};
+
+export const AssetInputPreview = ({ children }: AssetInputPreviewProps) => {
```
noahwaldner: this component could be just a container with all the layouting in css

noahwaldner (changes requested):
Thank you for your contribution!

The component works and looks correct 💪🏻

I added comments regarding optimization. Many things could be done with css to avoid the expensive parsing of children. 

I did not yet review the scss file as it will change a lot once more layout logic is moved there.

packages/components/src/components/AssetInput/AssetInputPreview.tsx
fulopdaniel: The preview container background is white, and each item inside has this grayish background. There are thin white lines separating the items, that's why I opted for this "empty parts" approach.

Image

packages/components/src/components/AssetInput/AssetInputPreview.tsx
fulopdaniel: I adjusted all implementations to avoid the `Children.forEach` and rely on typing the `children` prop like this:

```
type AssetInputPreviewPart = ReactElement<
    AssetInputPreviewImageProps | AssetInputPreviewIconProps | AssetInputPreviewLoadingProps
>;

type AssetInputPreviewProps = {
    children: AssetInputPreviewPart | AssetInputPreviewPart[];
};
```

However, this does not seem to be working for me, type validation does not kick in. E.g. I can do this without any errors:

```
    <AssetInputPreview>
        <div>Invalid part</div>
        <AssetInputPreviewImage src="https://picsum.photos/101/150" />
    </AssetInputPreview>
```

What am I missing?

packages/components/src/components/AssetInput/AssetInput.tsx
```diff
@@ -0,0 +1,31 @@
...
+import { AssetInputPreviewIcon } from './AssetInputPreviewIcon';
+import { AssetInputPreviewImage } from './AssetInputPreviewImage';
+import { AssetInputPreviewLoading } from './AssetInputPreviewLoading';
+import { AssetInputRoot, type AssetInputRootProps } from './AssetInputRoot';
+import { AssetInputTitle } from './AssetInputTitle';
+import { AssetInputUploadInput } from './AssetInputUploadInput';
+
+export type { AssetInputRootProps };
+
+export const AssetInput = {
```
noahwaldner, line 19: let's add forward ref to all subcomponents

packages/components/src/components/AssetInput/AssetInputPreview.tsx
```diff
@@ -0,0 +1,32 @@
...
+    const parts = Children.toArray(children).slice(0, MAX_PREVIEW_PARTS);
+    const emptySlotCount = parts.length > 1 ? MAX_PREVIEW_PARTS - parts.length : 0;
+
+    return (
+        <div className={styles.preview}>
+            {parts}
+            {Array.from({ length: emptySlotCount }, (_, index) => (
+                <div className={styles.previewSlot} key={`empty-${index}`} />
```
noahwaldner: index as key is not stable

packages/components/src/components/AssetInput/__tests__/AssetInput.ct.tsx
```diff
@@ -0,0 +1,191 @@
...
+test('should render placeholder with upload and browse actions', async ({ mount }) => {
+    const onBrowse = sinon.spy();
+    const onFileChange = sinon.spy();
+    const wrapper = await mount(
+        <AssetInput.Placeholder>
+            <AssetInput.UploadInput acceptFileType="image/*" onSelect={onFileChange} />
+            <AssetInput.BrowseInput onBrowse={onBrowse} />
+        </AssetInput.Placeholder>,
+    );
+    await expect(wrapper.getByRole('button', { name: /Upload/i })).toBeVisible();
```
noahwaldner: lets use data test id and not depend on the translation labels

packages/components/src/components/AssetInput/AssetInputPreviewLoading.tsx
```diff
@@ -0,0 +1,17 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+import { LoadingCircle, type LoadingCircleProps } from '../LoadingCircle/LoadingCircle';
+
+import styles from './styles/asset-input.module.scss';
+
+export type AssetInputPreviewLoadingProps = {
+    size?: LoadingCircleProps['size'];
```
noahwaldner: why does this need to be customizable? this should be consistent

packages/components/src/components/AssetInput/__tests__/AssetInputRoot.ct.tsx
```diff
@@ -0,0 +1,191 @@
...
+            </AssetInput.Root>,
+        );
+        const root = wrapper.getByRole('button', { name: /3 assets/ });
+        await expect(root).toBeVisible();
+        await expect(root).toHaveAttribute('data-orientation', orientation);
+        await expect(root).toContainText('2 locations');
+        await expect(root.locator('img')).toHaveCount(2);
+        await expect(root.getByTestId('fondue-loading-circle-content')).toBeVisible();
+    });
+}
```
noahwaldner, line 196: This is a pretty rich component so it makes sense to add proper tests for all cases. Maybe it makes sense to split up the test file

Some of the gaps i see:

AssetInputPreview
- Logic for filling with place holders up to 4
- Truncation: pass5 children, only show 4
- Single child passed, no placeholders added

AssetInputUploadInput
- acceptFileType & allowMultiple propagation
- Default allowMultiple=false not tested
- clicking the button triggers the file input

AssetInputPlaceholder
- Order: no test if the visual order is consistent, even when the child order changes

AssetInputRoot
- Test for triggering with Enter key
- default orientation not tested
- caret rotation

packages/components/src/components/AssetInput/AssetInput.stories.tsx
```diff
@@ -0,0 +1,251 @@
...
+import { AssetInput, type AssetInputRootProps } from './AssetInput';
+
+type StoryArgs = AssetInputRootProps & { acceptFileType: string };
+
+const meta = {
+    title: 'Components/AssetInput',
+    tags: ['autodocs'],
+    component: AssetInput.Root,
+    subcomponents: {
+        Placeholder: AssetInput.Placeholder,
```
noahwaldner: Use the proper keys to display in the story (check Dropdown story for an example)

packages/components/src/components/AssetInput/AssetInput.stories.tsx
```diff
@@ -0,0 +1,251 @@
...
+                        <Dropdown.Item emphasis="danger" onSelect={action('remove')}>
+                            <IconTrashBin size={16} />
+                            Delete
+                        </Dropdown.Item>
+                    </Dropdown.Group>
+                </Dropdown.Content>
+            </Dropdown.Root>
+        );
+    },
+};
```
noahwaldner, line 405: Maybe add examples for the following

- Only Title
- Long title (overflow)
- Only one metadata item
- Many metadata items (to test overflow)
- Full example with switch from placeholder to AssetInput + Dropdown

packages/components/src/components/AssetInput/styles/asset-input.module.scss
```diff
@@ -0,0 +1,220 @@
...
+        order: 0;
+    }
+
+    [data-asset-input-action='browse'] {
+        order: 1;
+    }
+
+    &:has([data-asset-input-action='upload']) [data-asset-input-action='browse']::before {
+        content: '';
+        position: absolute;
```
noahwaldner, line 257: add a position: relative to the parent to make this safe

packages/components/src/components/AssetInput/styles/asset-input.module.scss
```diff
@@ -0,0 +1,220 @@
...
+            grid-template-columns: sizeToken.get(15.5) sizeToken.get(15.5);
+            grid-template-rows: sizeToken.get(15.5) sizeToken.get(15.5);
+            gap: sizeToken.get(1);
+            justify-content: center;
+        }
+    }
+}
+
+.preview {
+    grid-area: preview;
```
noahwaldner, line 133: add box-sizing

packages/components/src/components/AssetInput/styles/asset-input.module.scss
```diff
@@ -0,0 +1,220 @@
...
+        width: sizeToken.get(14);
+
+        .previewImage {
+            object-fit: contain;
+        }
+    }
+
+    &:has(> :nth-child(2)) {
+        display: grid;
+        grid-template-columns: sizeToken.get(11) sizeToken.get(11);
```
noahwaldner, line 154: The size is different when multiple assets are in the preview vs. when a single one is in there. Is thi sintentional?

noahwaldner (changes requested):

packages/components/src/components/AssetInput/AssetInputPreviewLoading.tsx
fulopdaniel: Not necessary, I just felt in vertical view the default size might not be big enough. And since it's a prop I could not style it dynamically with CSS.

image

packages/components/src/components/AssetInput/styles/asset-input.module.scss
fulopdaniel: It's already added at line98.

packages/components/src/components/AssetInput/styles/asset-input.module.scss
fulopdaniel: Yes, see legacy component:
image
image

noahwaldner: Not sure if we should keep that. aligning it probably makes more sense

packages/components/src/components/AssetInput/AssetInputPreviewLoading.tsx
noahwaldner: I would lock it to a size, else every usage will look different

packages/components/src/components/AssetInput/styles/asset-input.module.scss
fulopdaniel: I can't really make this decision, maybe design will have some issues with the removal of this view.
I played around and showing 4 images in the small view doesn't look so good.

I would love to keep this as is, and maybe reiterate on this later.

jcosta33 (approved):
everything looks good, I have nothing to add beyond what Noah already commented but I did notice a potential issue

Image

where the metadata seems to just get cut off. Not sure if this is something we can live with or not, just thought I should flag it.

jcosta33:
force merging as @noahwaldner is off and comments have been addressed

PR #2734, opened 2026-05-26
feat: migrate charts package to scss

noahwaldner:
Migrate fondue-charts to use scss instead of tailwind
- Move tailwind imports from index.css to tailwind.css file for storybook
- Migrate component styles to use scss classes
- add scss processing to vite bundle

Notice: This migration was performed with claude code. All changes have been checked manually.
Please review and test carefully


noahwaldner:
Preview for fondue charts

https://deploy-preview-2734.charts.fondue-components.frontify.com/?path=/story/charts-area-stack-chart--area-stacked-chart-storage

joseazevedo88:
This makes sense to me, even though I prefer tailwind over scss. One thing I noticed is a lot of repetition in fonts(should be extracted to shared fonts classes) and sizes should use an internal scale instead of rems probably. Other than that it's good to go for me. I would be interested in hearing the performance gains of this change

SamuelAlev (approved):

joseazevedo88 (approved):

PR #2738, opened 2026-05-27
feat: add sdk skill and cli

noahwaldner:
- Add SDK cli 
- Add claude skill

### DIFF
--- a/packages/sdk/adapters/claude-skill/skill/SKILL.md
+++ b/packages/sdk/adapters/claude-skill/skill/SKILL.md
@@ -1,0 +1,207 @@
+---
+name: fondue
+description: Queries the locally installed `@frontify/fondue/sdk` to surface accurate Fondue component names, props, design tokens, utility classes, and setup guides — pinned to the version the project depends on. Use when picking a Fondue component (button, dialog, dropdown, tabs, …), finding a design token or utility class, or setting up the `@frontify/fondue` package.
+allowed-tools: Bash(node *)
+---
+
+# Fondue
+
+Use `@frontify/fondue/sdk` as the source of truth for every Fondue question. Never invent component names, props, token ids, or imports from memory — query the SDK.
+
+**Runs against the project's locally installed Fondue.** Invoke `node` from the project root (or a subdirectory) so `@frontify/fondue` resolves from `node_modules` — results reflect the exact version this project depends on, not the version the skill was authored against.
+
+## Verify the SDK is available
+
+```bash
+node -e "import('@frontify/fondue/sdk').then((m) => console.log(JSON.stringify({ components: m.components.size, tokens: m.tokens.size, guides: m.guides.size })))"
+```
+
+If that errors with `Cannot find package`, the project hasn't installed Fondue yet:
+
+```bash
+pnpm add @frontify/fondue   # or npm i / yarn add
+```
+
+## How to query
+
+The SDK is a synchronous, zero-I/O ES module that resolves from the user's project. Three singletons share the same query surface:
+
+```ts
+import { components, tokens, guides } from '@frontify/fondue/sdk';
+
+// Same shape on each domain:
+components.list();
+components.get('Button');             // ComponentNode | undefined (never throws)
+components.has('Button');             // boolean
+components.where({ text: 'dropdown' });
+components.size;                      // number
+```
+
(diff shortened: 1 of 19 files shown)


packages/sdk/adapters/claude-skill/skill/reference.md
```diff
@@ -0,0 +1,238 @@
+# Fondue SDK — agent reference
+
+Formal contract for `@frontify/fondue/sdk`. Load this when the patterns in `SKILL.md` don't expose what you need — for example, when you're generating code that imports the SDK directly, or when you need to verify a filter clause name or method signature.
```
jcosta33: If the agent is reading this, it already loaded the reference. This is not indexed like the skills' descriptor

packages/sdk/adapters/claude-skill/skill/SKILL.md
jcosta33, line 1: see https://github.com/Frontify/skills/blob/main/docs

packages/sdk/adapters/claude-skill/skill/SKILL.md
```diff
@@ -0,0 +1,207 @@
+---
+name: fondue
+description: Queries the locally installed `@frontify/fondue/sdk` to surface accurate Fondue component names, props, design tokens, utility classes, and setup guides — pinned to the version the project depends on. Use when picking a Fondue component (button, dialog, dropdown, tabs, …), finding a design token or utility class, or setting up the `@frontify/fondue` package.
```
jcosta33:
```suggestion
description: Queries the locally installed `@frontify/fondue/sdk` to surface accurate Fondue component metadata. ALWAYS use when picking a Fondue component (button, dialog, dropdown, tabs, …), finding a design token or utility class, or setting up the `@frontify/fondue` package.
```

Claude is set up to only accept 1% of context to be filled up by skills descriptors. That sounds like a lot at first but if you have like 10 skills or so it is very quickly overloaded and then claude truncates things, leading to very poor pick up rates. So we have to be as economical as possible with the description

packages/sdk/adapters/claude-skill/skill/SKILL.md
```diff
@@ -0,0 +1,207 @@
...
+| Assuming a component exists ("there must be a `Combobox`") | `components.has('Combobox')` first, or `components.where({ text: 'combobox' })` |
+| Hardcoding a hex / px value in custom code | `tokens.where({ text: '<intent>' })` — most "obvious" values have a token |
+| Treating icons like normal components, reading `status` / `props` | Detect with `node.category().name === 'icon'`. Icons have empty `status`, `props`, `related`. |
+| Importing `Button` from the wrong path | `components.get('Button')?.importStatement` is authoritative |
+| Suggesting setup steps from memory | `guides.get('getting-started')?.content` — always the live text |
+| `console.log(node)` instead of `JSON.stringify` | Facet methods serialize as `[Function]`; always stringify before logging |
+
+## Going deeper
+
+[`reference.md`](./reference.md) sits next to this file and ships with it — the formal contract: every export, every filter clause per domain, every node and facet type, error semantics. Load it when you need to verify a filter clause is valid (`ComponentFilter`, `TokenFilter`, `GuideFilter`) or that a method exists. Prefer it over web lookups or upstream GitHub.
```
jcosta33:
```suggestion
[`reference.md`](./reference.md) is the formal contract: every export, every filter clause per domain, every node and facet type, error semantics. Load it when you need to verify a filter clause is valid (`ComponentFilter`, `TokenFilter`, `GuideFilter`) or that a method exists. Prefer it over web lookups or upstream GitHub.
```

packages/sdk/cli.ts
```diff
@@ -0,0 +1,114 @@
...
+const program = new Command().name('fondue').description('CLI for Frontify Fondue.').version(__FONDUE_VERSION__);
+
+// ─── `fondue adapter …` ────────────────────────────────────────────────────
+
+const adapter = program
+    .command('adapter')
+    .description('Install and manage Fondue SDK adapters (Claude Code skill, MCP, REST, …).');
+
+adapter
+    .command('install [name]')
```
jcosta33, line 67: being able to install globally is also important as well as having support not just for claude code but also cursor, codex and whatnot right? similar to https://github.com/vercel-labs/skills

noahwaldner: can be added in a iteration imo

jcosta33 (approved):

PR #2750, opened 2026-06-02
fix: align triggers with content and fix layout shift on active change

noahwaldner:
This PR fixes two issues:

- The Tab triggers for `variant=default` are now aligned with the left edge of the content

- The labels in the triggers caused a layout shift when switching active state, as the bold text has a different width.
  - It now two overlaying labels with different font weights and toggle between them. The same concept is already used in `SegmentedControls`
  
 Fixes FP-212

### DIFF
--- a/packages/components/src/components/Tabs/Tabs.tsx
+++ b/packages/components/src/components/Tabs/Tabs.tsx
@@ -14,8 +14,6 @@ import {
     type ReactNode,
 } from 'react';
 
-import { useControllableState } from '#/hooks/useControllableState';
-
 import { Button } from '../Button/Button';
 import { Dropdown } from '../Dropdown/Dropdown';
 import { useFondueTheme } from '../ThemeProvider/ThemeProvider';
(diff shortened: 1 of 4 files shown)


packages/components/src/components/Tabs/Tabs.tsx
```diff
@@ -142,7 +142,12 @@ export const TabsRoot = (
                                 className={styles.trigger}
                                 ref={trigger.ref}
                             >
-                                {trigger.element}
+                                <span className={styles.triggerLabel}>
+                                    <span className={styles.triggerLabelActive} aria-hidden="true">
```
syeo66, line 146: Not sure if 'active'/'inactive' are appropriate. It sound more like a state than describing the nature of the thing. I'd use `triggerLabel` for the actually visible element and something like `triggerLabelSpaceholder` for the invisible one.

syeo66 (approved):

packages/components/src/components/Tabs/Tabs.tsx
noahwaldner: Active inactive is kind of appropriate, as it's the label for when the tab related to this trigger is active or not

noahwaldner: `triggerLabelSpaceholder` could make it seem like it is never visible, but it switches between the two

syeo66: Yeah, did not realize they are switzing visibility.

PR #2758, opened 2026-06-05
feat: add tree component [BETA]

noahwaldner:
Add refactored tree component

Features:
- Hierarchical composition: arbitrarily nested folders via JSX (`Tree.Root`, `Tree.Item`, `Tree.Folder`).
- Keyboard navigation: focus, expand/collapse, and activate rows via arrow keys / Enter / Space.
- Multi-select (`multiSelect`): row checkboxes with cascading toggle and indeterminate state on folders whose descendants are partially selected.
- Drag-and-drop reordering (`reorderable`): reposition items across the tree with a drop indicator
- Drop gating (`accepts` on `Tree.Root` / `Tree.Folder`): drop rules to be combined with per-row tags for category-based matching.
- Async loading rows: `Tree.Loading` renders a translated "Loading…" placeholder inside a folder or at the root.
- Trailing actions: `Tree.ItemAction` as a child renders controls (overflow menu, edit/delete) at the end of a row.
- Active item state (`isActive`): declarative highlight for the current row, useful when the Tree drives navigation.
- Controlled state: render from the latest `TreeChangeState`; persist via global `onChange`. Per-row `onSelectChange`, `onExpandChange`, `onClick`, and `onMove` callbacks fan out from the global change for local
reactions.
- Accessibility; screen-reader hints for checkbox and reorder modes, ARIA-correct checked/indeterminate state.


packages/components/src/components/Tree/components/TreeRowChevron.tsx
```diff
@@ -0,0 +1,16 @@
...
+import styles from '../styles/tree.module.scss';
+
+type TreeRowChevronProps = {
+    isFolder: boolean;
+    isExpanded: boolean;
+};
+
+export const TreeRowChevron = ({ isFolder, isExpanded }: TreeRowChevronProps) => (
+    <span className={styles.chevron} aria-hidden>
+        {isFolder ? isExpanded ? <IconCaretDown size={12} /> : <IconCaretRight size={12} /> : null}
```
syeo66: The nested ternary feels a bit weird. Maybe wrap it into it's own local component with early return when it si not a folder?

syeo66 (approved):
Nice work! Could not find any strange behavior and the code looks clear and clean to me.

PR #2763, opened 2026-06-10
feat: improvements on tree component [BETA]

noahwaldner:
- Change api to use Slots/SubComponents instead of props
  - Allow for a icon to be passed in
- Remove default icon to match figma
- Add rename functionality
- When dropping into a folder, add a indent to the drop line indicator


syeo66 (approved):
Crazy focus handling. But works and looks only as bad as the thing it does. 😆 Nice one.

PR #2765, opened 2026-06-11
fix(Card): restore the border on the selected state

mnoleto:
### Selected card lost its border

When you select a card, it's supposed to get a border to show it's selected — and the same when you hover a card that's already selected. Right now that border isn't there at all: selecting a card (or hovering a selected one) leaves it with no border, so there's no clear indication that it's the selected one.

This brings the selected border back, using the color-line-strong value from the design.

### DIFF
--- a/.changeset/card-selected-border.md
+++ b/.changeset/card-selected-border.md
@@ -1,0 +1,7 @@
+---
+"@frontify/fondue-components": patch
+---
+
+fix(Card): restore the border on the selected state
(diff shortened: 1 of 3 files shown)


noahwaldner:
can you please add a test case for it?

mnoleto:
> can you please add a test case for it?

Done!

noahwaldner (approved):

PR #2775, opened 2026-06-30
feat(Accordion): add pill variant

syeo66:
Adds `pill` variant to the Accordion.

Screenshot 2026-06-30 at 12 54 25


jcosta33 (approved):
image

lgtm but are we missing soma space here?

PR #2781, opened 2026-07-05
fix(RichTextEditor, ScrollArea): wrap long words and URLs instead of scrolling horizontally

sgruetter:
## Intent

Long unbreakable strings — URLs, long words — in rich text no longer force a sideways scroll. They<br>overflowed horizontally instead of wrapping, most visibly in the web-app comment sidebar both while<br>typing and after posting (a WCAG 1.4.10 Reflow failure at narrow widths). This fixes the wrapping in<br>`RichTextEditor`.

The issue is shown here:

https://github.com/user-attachments/assets/a998647d-2dda-4534-9f69-c4b3e521f718

## Root cause

The editor and the serialized rendition already carried `overflow-wrap: break-word`, but `break-word`<br>only wraps *visually* — it does not lower the element's reported **min-content width**. Inside a<br>container that sizes to its content the width therefore still grows to the longest token, and the<br>text never wraps. The comment sidebar is exactly such a container: `ScrollArea`'s Radix viewport<br>wraps its children in an inline `display: table; min-width: 100%` element that shrink-wraps to<br>max-content. List items compound it by rendering as `inline-flex`, which also shrink-wraps.

`overflow-wrap: anywhere` wraps like `break-word` visually **and** lowers the reported min-content<br>width, so the content wraps to the available width even inside a table / inline-flex context.

## What changed

* `RichTextEditor` — paragraphs and list items, both while editing and in the serialized (posted)<br>rendition, now use `overflow-wrap: anywhere` instead of `break-word`.
* The earlier `ScrollArea` `display: block !important` override is **reverted**. The defect lives in<br>RTE, and forcing block on the shared Radix viewport (\~66 web-app consumers) was more blast radius<br>than the fix needs — `overflow-wrap: anywhere` fixes the content regardless of its container.
* The `tw-max-w-full` cap on list items is dropped — redundant once min-content is lowered.

Genuinely wide, unshrinkable content (fixed-width media, `nowrap`) still scrolls horizontally.

## Evidence

* New `RichTextEditor` component test (real browser, `LongContentWrapping.cy.tsx`) drives a long<br>unbreakable token through a faithful reproduction of Radix's `display: table; min-width: 100%`<br>viewport — for the editor paragraph, the editor list item, and the serialized rendition — and<br>asserts no horizontal overflow (`scrollWidth ≤ clientWidth`). It is **red on** `break-word`<br>**(1347/1564px overflow) and green on** `anywhere`.
* Full `@frontify/fondue-rte` Cypress suite: **468/468** (chromium); serializer specs updated to the<br>new class. No visual regression — normal text still wraps at word boundaries, only over-long<br>tokens break.
* `build`, `typecheck`, `lint` clean; `dist/style.css` emits `overflow-wrap: anywhere`.<br>`@frontify/fondue-components` is unchanged.

Tested locally and it looks like so:

https://github.com/user-attachments/assets/27636b10-63c0-43fd-a74d-b494377fe10a

## Notes for reviewers

* **No shared-primitive change** — `ScrollArea` is untouched; the fix is contained to<br>`RichTextEditor` (`overflow-wrap: anywhere` on paragraphs, list items, and the serializer).
* **In-repo behavioral coverage** — both the editor and the rendition paths now have an automated<br>wrap test here (`LongContentWrapping.cy.tsx`), plus a resizable `LongContentWrapping` Storybook<br>story for manual checking.
* **Final confirmation** — the fix is proven in Fondue's isolated environment; the web-app comment<br>sidebar (its real context) is best confirmed by a live render / e2e there before rollout.


sgruetter (commented):
Are the comments helpful? They are surprising (carry information) to some degree, but is it worth it?

packages/components/src/components/ScrollArea/styles/scrollArea.module.scss
```diff
@@ -16,6 +16,16 @@
...
 
+    /* Radix wraps the viewport children in an inline `display: table; min-width: 100%` element that
+       shrink-wraps to max-content. That widens the region to fit the longest unbreakable token, so
+       `overflow-wrap`/`word-break` on descendants never engages and long words/URLs scroll sideways
+       instead of wrapping. Forcing the wrapper to block caps it at the viewport width so text wraps;
+       genuinely-wide content (fixed-width media, nowrap) still overflows and scrolls. `!important`
+       is required to beat Radix's inline style. */
+    > div {
+        display: block !important;
+    }
```
jcosta33: we usually fix this at the web-app level when an issue manifests itself to avoid overriding radix globally (see `NavigationLayout.tsx:l42` for example). It's such a commonplace issue that this might be justified but we have to test very thoroughly to see if it doesn't create other side effects. @noahwaldner  wdyt?

sgruetter: I initially wanted to fix it directly in the comment area - and did so here https://github.com/Frontify/web-app/pull/15371 - but then closed it in favor of this change. If the Radix defaults lead to the problem, I'd fix this here (or make a component that wraps this - but then we'd have the same situation), as it otherwise just occurs again and again.

But of course you guys have to say, I'm merely testing the harness (and trying not to waste your time).

fulopdaniel (dismissed):
Stuff in rte package looks good, but I would remove the comments.

packages/components/src/components/ScrollArea/styles/scrollArea.module.scss
jcosta33: At the time we preferred the case by case basis because we were trying to avoid unintended consequences of doing it here, but after all this time we've seen nothing but issues so maybe it's not such a bad idea anymore. I would still loop in @noahwaldner  as I haven't worked on fondue in a long time, maybe he has a different take

sgruetter:
@fulopdaniel I treated this change with the latest "no comments by default - only allow then when they are the sole means to understand something" behavior, and it improved it a bit.

packages/components/src/components/ScrollArea/styles/scrollArea.module.scss
noahwaldner: This isn't a `ScrollArea` bug. Radix's display: table wrapper is working as intended for the vast majority of consumers.

The horizontal-scroll issue only surfaces because the legacy RTE emits content that shrink-wraps to a long token. Forcing `display: block !important` on the viewport wrapper changes behavior for every ScrollArea in the design system to solve a problem that only one type of content has. That's a lot of blast radius (and a Radix-internals coupling we'd have to babysit on upgrades) to fix a defect that lives in RTE.

packages/rte/src/components/RichTextEditor/Plugins/ListPlugin/ListItemContentMarkupElement.tsx
```diff
@@ -48,7 +49,7 @@ const ListBullet = () => {
 
 export const ListItemContentMarkupElementNode = ({ attributes, children, element }: PlateRenderElementProps) => {
     return (
-        <p className={getLicElementClassNames(element)} {...attributes}>
```
noahwaldner, line 51: `break-word` doesn't affect the reported min-content width of the text element, it only wraps the text visually, therefore causing the width to still grow.

`overflow-wrap: anywhere` also lowers the reported min width, but might have other implications that need to be checked

packages/rte/src/components/RichTextEditor/Plugins/ListPlugin/ListItemContentMarkupElement.tsx
```diff
@@ -48,7 +49,7 @@ const ListBullet = () => {
...
+        <p className={merge([getLicElementClassNames(element), 'tw-break-words'])} {...attributes}>
```
noahwaldner: Please add a behavioral test for this change

noahwaldner (changes requested):

packages/rte/src/components/RichTextEditor/Plugins/ListPlugin/ListItemContentMarkupElement.tsx
sgruetter: I thought so too, but I couldn't test it manually. How can we test this manually? Maybe it's right somehow as the tests seem to be correct – but I don't believe them in this case.

sgruetter: I could now test it manually and fixed it. You were right (unsurprisingly). The `overflow-wrap: anywhere` seems to fix it. See the updated description for the fix.

packages/rte/src/components/RichTextEditor/__tests__/LongContentWrapping.cy.tsx
```diff
@@ -0,0 +1,82 @@
...
+        const { scrollWidth, clientWidth } = $container[0];
+        expect(
+            scrollWidth,
+            `content width (${scrollWidth}px) stays within its container (${clientWidth}px) — the long token wrapped instead of forcing a horizontal scrollbar`,
+        ).to.be.at.most(clientWidth + 1);
+    });
+};
+
+describe('RichTextEditor wraps long unbreakable content in a shrink-wrapping container', () => {
+    it('wraps a long token while editing a paragraph', () => {
```
sgruetter, line 49: Do we also have the `should` house rule in Fondue? Then I'll update this and add it to the harness.

PR #2784, opened 2026-07-08
fix: type resolution during manifest generation

noahwaldner:
This fixes the collection of type declarations when generating manifest files to properly resolve nested types

### DIFF
--- a/packages/components/scripts/buildManifest/utils/collectTypeDefinitions.ts
+++ b/packages/components/scripts/buildManifest/utils/collectTypeDefinitions.ts
@@ -7,11 +7,17 @@ import {
     EmitHint,
     forEachChild,
     isInterfaceDeclaration,
+    isMappedTypeNode,
     isTypeAliasDeclaration,
+    isTypeQueryNode,
+    type InterfaceDeclaration,
     JsxEmit,
     NewLineKind,
     type Node,
     type Program,
+    type SourceFile,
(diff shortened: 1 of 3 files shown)


packages/components/scripts/buildManifest/utils/collectTypeDefinitions.ts
```diff
@@ -51,43 +57,94 @@ const extractTypeNamesFromString = (typeStr: string): string[] => {
...
+//   - mapped types → e.g. `[key in Breakpoint]?: TValue` instead of the expanded members
+// Detect them so we can fall back to the type checker, which resolves them to concrete types.
+const needsCheckerResolution = (node: Node): boolean => {
+    if (isTypeQueryNode(node) || isMappedTypeNode(node)) {
+        return true;
+    }
+    let found = false;
+    forEachChild(node, (child) => {
+        if (!found && needsCheckerResolution(child)) {
+            found = true;
```
syeo66: Couldn't that just return true? The loop could never set it to false, so as soon as it is true, it stays.

syeo66 (approved):
Looks good to me. Just a small comment/nit.

packages/components/scripts/buildManifest/utils/collectTypeDefinitions.ts
noahwaldner: you are right! wil adjust !

PR #2797, opened 2026-07-16
fix(Badge): font weight should not inherit from parent containers

SamuelAlev:
https://frontify.slack.com/archives/C06D0LAQFB4/p1784057776578259

### DIFF
--- a/.changeset/badge-font-weight-normal.md
+++ b/.changeset/badge-font-weight-normal.md
@@ -1,5 +1,6 @@
 ---
 '@frontify/fondue-components': patch
+'@frontify/fondue': patch
 ---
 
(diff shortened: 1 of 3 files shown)


noahwaldner (approved):

.changeset/badge-font-weight-normal.md
```diff
@@ -0,0 +1,5 @@
+---
+'@frontify/fondue-components': patch
```
noahwaldner, line 2: needs a bump of the main package, as a patch doesnt trigger it

noahwaldner (changes requested):

noahwaldner (approved):

PR #2801, opened 2026-07-16
feat: improve the sdk api

noahwaldner:
**Type safety**

- status is always set: story parser understands satisfies Meta<...>, six components got their stories wired up, icons ship as 'released'
- New exported literal unions derived from the bundled data: ComponentStatus, ComponentCategory, TokenCategory (filter typos now fail at compile time)
- null instead of '' for not-applicable fields: instructions, deprecationMessage, cssVariable

**API surface**

- New components.statuses() / components.status(name) facet accessors
- Every entity now carries a canonical id (equals name for components)
- Removed the build-system output field from tokens
- Removed the redundant ComponentDetails.subComponentNames
- Token values stay by-reference (value, cssVariable, tailwindClass): no resolved literals, rendered values are theme/brand-dependent at runtime

**Behavior**

- All data is deeply frozen: mutating nodes or list() results throws instead of corrupting shared state
- keyPathStartsWith is segment-aware: 'colors.chart' no longer matches colors.charts.*
- Facet ordering pinned to the 'en' locale for cross-machine determinism
- The SDK build now validates the data snapshot (missing status/category, dangling relatedComponents, duplicate ids) and fails instead of shipping inconsistencies

**Docs**

- Filter edge cases defined: empty-array clause matches nothing, text: '' is ignored
- tailwindClass * wildcard documented as a utility-prefix placeholder (bg-, text-, border-, ...)
- Token coverage documented as intentionally curated (colors + semantic)
- Icon quirks documented: empty description, empty props/related(), instructions: null
- Sub-components documented as not top-level addressable
- Fixed all stale claims: phantom status() method, icon-status contradictions, wrong counts, "fuzzy" vs substring match

**Versioning and packaging**

- Contract version = @frontify/fondue-sdk version; umbrella pins and re-exports it
- sideEffects: false, engines: node >=20


packages/fondue/scripts/detectDeprecatedExports/temp/active-exports.txt
syeo66, line 1: Do files in a 'temp' folder belong into the repository?

packages/sdk/src/utils/freeze.ts
```diff
@@ -0,0 +1,17 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+/**
+ * Recursively freeze a value in place. The SDK's data is a static snapshot,
+ * so freezing at build time makes the documented immutability real: mutating
+ * a shared collection (e.g. sorting `list()` in place) throws instead of
+ * silently corrupting every other consumer in the process.
+ */
+export const deepFreeze = <T>(value: T): T => {
```
syeo66, line 9: Image

syeo66 (approved):

PR #2809, opened 2026-07-23
feat(RadioList): render custom content as the radio [TEM-479]

mikeyrayvon:
## 📖 Background

Templates has two new cases where a `radiogroup` is the appropriate role for a custom form element. Instead of building a bespoke component behaving as an accessible radio group, we would like to benefit from the Fondue `RadioList`, but use our own custom card component as the `RadioButton`.

Today `RadioButton` hardcodes the `<Indicator>` dot as its only child, so `asChild` merges onto the dot rather than a card.

## 🎯 What does this PR do?

Lets a `RadioList.RadioButton` render your own element as the radio instead of the built-in dot. Pass `children` to replace the dot indicator, or `asChild` to make a consumer element (e.g. a large selectable card) *be* the radio — it still gets the radio role, checked state, roving tab order, and arrow-key navigation from the primitive. Existing radios (no `children`/`asChild`) are unchanged.

## 🧪 How to test

- Storybook → **Components/RadioList → AsSelectableCards**: click a card and arrow-key between them; only one is selected at a time and focus follows selection (single tab stop).
- Confirm the existing stories (Default, Vertical, Disabled, …) still render the standard radio dot.
- Style the selected state via the `data-state="checked"` attribute the primitive sets (shown in the story).

## 🖼️ Screenshots

Storybook:

image

Use cases:

image

image

## 📌 Related tickets & additional notes

- **TEM-479**
- In `asChild` mode the default `styles.item` dot styling is intentionally skipped so it doesn't fight the consumer's element; the consumer owns styling (and focus visuals) via `data-state`. Backwards-compatible — the change is purely additive.


syeo66 (approved):

noahwaldner:
I'm not sure if this is something that really needs to be in fondue.

@mikeyrayvon do you know if the designers had an alignemnt on this?

packages/components/src/components/RadioList/styles/radiolist.stories.module.scss
```diff
@@ -0,0 +1,33 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+.card {
```
noahwaldner, line 3: for story examples we use tailwind

noahwaldner (changes requested):

noahwaldner:
I had a call with @mikeyrayvon and we will pause this until johanna is back to discuss the addition of a "card" variant, to avoid accepting custom components

PR #2844, opened 2026-09-07
fix(Dialog): stack portaled overlays above host page content

peter-tudosa:
## Problem

`Dialog` renders its portal with no `z-index` of its own. The `.content` rule carries `z-index: 1`, but `.underlay` is `position: fixed` and therefore its own stacking context, so that `1` is scoped and invisible from the outside — the whole dialog competes at `z-index: auto`.

That means it loses to any positioned sibling carrying even `z-index: 1`. Confirmed on a production guideline: the RichTextEditor's floating link modal sits at `z-index: 1` / `position: absolute`, the LinkChooser dialog root at `auto` / `position: static`, so the chooser renders *behind* the modal that opened it. Notably the dialog was later in the DOM — DOM order favoured it, and the single `z-index: 1` overrode that.

## Fix

The value has to go on the positioned element *inside* each portal. The outermost node is a `position: static` `ThemeProvider` wrapper where `z-index` is inert; for the popper-based overlays that means the Radix `Content` class, whose computed `z-index` Radix copies onto `[data-radix-popper-content-wrapper]` — the mechanism `Select` was already relying on at 30.

| Layer | Value | Components |
| --- | --- | --- |
| `dialog` | 20 | `Dialog` underlay |
| `popover` | 30 | `Dropdown`, `Flyout`, `Select`, `Tooltip` |

All four popovers portal to `document.body` with no container, so today they clear an open `Dialog` only by DOM order. Once the dialog has a layer they must be lifted too, or a dropdown opened *inside* a dialog falls behind it. The scale and that constraint are documented in `utilities/zIndex.module.scss`.

Also drops the unused `floatingModal: 1000` from the RTE's `zIndexLayers` — nothing references it, and it would have re-broken this the moment it was wired up.

## Verification

- All five Sass modules compile; emitted output confirmed as `.underlay{z-index:20}` and `z-index:30` on the four popover classes.
- No remaining references to `floatingModal`.
- ⚠️ **Test suite not run.** `pnpm install` fails in my checkout for a pre-existing, unrelated reason (`EACCES` — root-owned `node_modules` from an earlier `sudo` install), which leaves `@vitejs/plugin-react` unresolvable and blocks both `pnpm build` and `vitest`. **The Playwright snapshot tests in particular need a run before merge.**

## Notes for review

- The value choice is deliberately conservative to limit blast radius against consumer chrome. The trade-off: the RTE's own `floatingToolbar` (50) and `combobox`/`emojiPopover` (100) still paint above an open dialog. That is pre-existing, not introduced here — but if we'd rather clear the whole RTE range it is a one-line change to `dialog: 200` / `popover: 300`.
- `Select`'s bare `30` became `zIndex.get(popover)` with no behaviour change.
- Needs a companion web-app change: its Themes RTE hardcodes `zIndex: 50` inline, which still beats `dialog: 20`. PR to follow.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.changeset/heavy-dialogs-stack.md
+++ b/.changeset/heavy-dialogs-stack.md
@@ -1,0 +1,18 @@
+---
+"@frontify/fondue-components": patch
+"@frontify/fondue-rte": patch
+"@frontify/fondue": patch
+---
+
+fix(Dialog): stack portaled overlays above host page content
+
+`Dialog` rendered its portal with no `z-index` of its own, so it landed at `auto` and
+lost to any positioned sibling carrying even `z-index: 1` -- for example the
+RichTextEditor's floating link modal, which left the internal LinkChooser rendering
+behind the modal that opened it.
+
+`Dialog` now sits at 20, and `Dropdown`, `Flyout`, `Select` and `Tooltip` at 30 so they
+stay above a `Dialog` they are opened from. The scale lives in
+`utilities/zIndex.module.scss`. Also drops the unused `floatingModal: 1000` entry from
+the RichTextEditor's `zIndexLayers`, which would have re-broken this the moment it was
+wired up.
--- a/packages/components/src/components/Dialog/styles/dialog.module.scss
+++ b/packages/components/src/components/Dialog/styles/dialog.module.scss
@@ -4,6 +4,7 @@
 @use '../../../utilities/focusStyle.module.scss';
 @use '../../../utilities/transitions.module.scss';
 @use '../../../utilities/mediaQuery.module.scss';
+@use '../../../utilities/zIndex.module.scss';
 
 .content {
     display: block;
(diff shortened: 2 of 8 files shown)


fulopdaniel:
As far as I know the goal is to have as little z-index as possible on the portaled elements and rely on the order. Is it possible to remove the RichTextEditor's floating link modal `z-index: 1` instead?

The code looks good btw, so if the solution is accepted by FPT I'm happy to approve.

peter-tudosa:
> As far as I know the goal is to have as little z-index as possible on the portaled elements and rely on the order. Is it possible to remove the RichTextEditor's floating link modal `z-index: 1` instead?
> 
> The code looks good btw, so if the solution is accepted by FPT I'm happy to approve.

I think it could work, might be that we need to take a look at the RTE toolbar as well as right now it has z-index: 50 and sits on top of the Link/Button plugin modal.

PR #2847, opened 2026-09-10
feat: add indeterminate option for Select.Multiple and Combobox.Multiple

noahwaldner:
`Select.Multi` and `Select.Combobox.Multiple` accept a new `indeterminateValues` prop for values that apply to only some of the items.

### Behaviour
- Partially applied options render a dash instead of a checkmark and expose `aria-checked="mixed"`.
- The field shows a "2 mixed" badge after the selection badges. The count is also part of the accessible description.
- Indeterminate values are purely visual. They are never reported by `onSelect`. 

**Toggling a indeterminate item**
- Clicking a indeterminate item sets it to "selected" and it is reported back in `onSelect`
- Clicking it again, removes the "selected" state again and it is omitted from the array passed in `onSelect`
important: The consumer needs to manually remove it from the passed array of `indeterminateValues`, else it will show as indeterminate instead of "unselected" 

### Badge collapsing in narrow fields
- Selection badges collapse into `+N` as before.
- Once no selection badge fits, a "3 selected" badge replaces `+N`
- With mixed values, a single "Mixed" badge takes over when the two counts no longer fit side by side.


packages/components/src/components/Select/components/SelectMenu.tsx
```diff
@@ -169,6 +188,7 @@ export const SelectMenu = ({
                                                 }}
                                                 {...itemProps}
                                                 aria-selected={isSelected}
+                                                aria-checked={ariaChecked}
```
jcosta33, line 191: these 2 can coexist?

noahwaldner: yes technically this is possible as they server different purposes

jcosta33 (approved):
Diff a bit hard to parse but didn't find anything strange and it seems to work perfectly

